PR #21, opened 2022-01-25
Barrel threading, Memory Model and Instruction Decode modifications

donofrio:
Adds barrel threading to the Rev model by duplicating the register file. Current thread priority selection is  RR.  Adds functionality to the RevMem memory model to allow for large jumps. Memory is sectioned into "pages" and a basic logical to physical translation is performed. Several bugs fixed in instruction decode and execution, primarily in the floating point (double and single) instructions


donofrio:
Many changes / diffs are the result of RegFile going from a single struct to an array of structs (one per thread)

jleidel:
there may be an issue with enabling statistics from the `RevProc`.  The PR doesn't compile against SST 11.1.0.  The failure is due to the addition of the `SST_ELI_DOCUMENT_STATISTICS` macro in the header file.  Technically speaking, the `RevProc` class is not a component or subcomponent.  Its just a normal C++ class that lives inside the `SST::RevCPU` namespace.  As a result, the macro the the statistics definition won't hook up correctly in SST 11.1.0.  Rather than defining this inside the RevCPU class, we can hoist the definition of  values up to the `RevProc` level per CPU.  We track the actual values at the `RevProc` level as each HART executes.  Once the HART has completed, the top-level `RevCPU` component asks each `RevProc` for the appropriate statistics values.

jleidel (approved):
built + tested against SST 11.0.0 & SST 11.1.0; passed all tests

PR #22, opened 2022-02-11
syscall wrappers

ct-clmsn:
sys call wrappers - offers a way to pack parameters and make system calls


jleidel:
closing this PR.  This is targeting `master` and should be targeting `devel`.  Needs to be rebased from `devel` as the branch is currently stale.

PR #29, opened 2023-02-21
Adding SoftFP

Artlands:
This update implements the floating operations based on the SoftFP library (https://bellard.org/softfp/) and can support the C math library. 

It also fixes several bugs in DecodeInst, where some instructions (slli, srli, srai, R4-type instructions, etc.) were not decoded correctly. The trick for handling rounding mode is avoided by setting the `rm` field in advance. 

Some bugs in the instruction simulation are also fixed. Tested against `big_loop`, `math_log`, and `minfft` in the test folder.


jleidel:
@Artlands, we're still reviewing the implications of your PR.  There is a lot included here in a single commit.  We'll come back with more comments and questions as soon as we can.

Artlands:
> we're still reviewing the implications of your PR. There is a lot included here in a single commit. We'll come back with more comments and questions as soon as we can.

Thank you for the update! Please let me know if there is anything specific that I can provide or clarify to aid in your review process.

jleidel:
this needs to be split into multiple PR's

PR #31, opened 2023-03-24
Merging functional memHierarchy infrastructure to devel

jleidel:
includes basic memHierarchy support; note that caching is currently disabled as it causes a bus error

### DIFF
--- a/src/RevPrefetcher.cc
+++ b/src/RevPrefetcher.cc
@@ -8,7 +8,7 @@
 // See LICENSE in the top level directory for licensing details
 //
 
-#include "RevPrefetcher.h"
+#include "../include/RevPrefetcher.h"
 
 bool RevPrefetcher::IsAvail(uint64_t Addr){
 
--- a/src/RevPrefetcher.cc
+++ b/src/RevPrefetcher.cc
@@ -112,7 +112,7 @@ void RevPrefetcher::Fill(uint64_t Addr){
   // determine if the address is 32bit aligned
   if((Addr & 0x3ull)){
     // not 32bit aligned, adjust the base address by 2 bytes
-    Fill(Addr-2);
+    Fill(Addr&0xFFFFFFFFFFFFFFFC);
   }
 
   // allocate a new stream buffer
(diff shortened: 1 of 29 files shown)


src/RevPrefetcher.cc
```diff
@@ -102,6 +108,13 @@ bool RevPrefetcher::InstFetch(uint64_t Addr, bool &Fetched, uint32_t &Inst){
 }
 
 void RevPrefetcher::Fill(uint64_t Addr){
+
+  // determine if the address is 32bit aligned
+  if((Addr & 0x3ull)){
+    // not 32bit aligned, adjust the base address by 2 bytes
+    Fill(Addr-2);
```
donofrio, line 115: do we want to subtract or mask to do the alignment?

jleidel: subtracting aligns it to the next lowest aligned word.  is that sufficient?

src/RevProc.h
```diff
@@ -100,6 +101,7 @@ namespace SST{
           uint64_t totalCycles;
           uint64_t cyclesBusy;
           uint64_t cyclesIdle;
+          uint64_t cyclesStalled;
```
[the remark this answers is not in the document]
jleidel: this is deprecated... not the file location

src/RevPrefetcher.cc
donofrio: So if Addr == 0x1234567F, subtracting 2 gives 0x12234567D which is still not aligned, right?

jleidel: that's assuming that the incoming address request is not aligned.  in theory, it should be, but I see your point.  If we masked off the lower two bits, it would be aligned.  0x123456F would be 0x1234567C

jleidel: see the new code in [81d19f9](https://github.com/tactcomplabs/rev/pull/31/commits/81d19f92ed0f6342fc43d3a70cf6538c3a5bbccb)

donofrio: Looks good!

PR #32, opened 2023-03-28
Pipeline hazard

donofrio:
Adding Dependency Check logic so pipeline stalls on register hazards. New test added to check functionality.  Also reduced the loop size of big_loop test from two nested 1024 loops to nested 512 loops to reduce test runtime under 30 sec


jleidel:
Looks entirely reasonable.  Do we need to do any direct interfacing with the instruction prefetcher (other than checking the status of `Stalled`)?

jleidel:
review complete.  good to merge

PR #34, opened 2023-04-03
Isa fixes

donofrio:
Beginning focused ISA tests. Adding initial RV ASM infrastructure to allow the testing of individual instructions, this test requires write() to work properly, so id not included in the standard test suite. 

In addition, in creating / debugging this test some functional ISA bugs were uncovered, fixes for those are included

### DIFF
--- a/include/insns/RV32I.h
+++ b/include/insns/RV32I.h
@@ -552,7 +552,7 @@ namespace SST{
           R->RV32[Inst.rd] = dt_u32((int32_t)(td_u32(R->RV32[Inst.rs1],32)) + (int32_t)(td_u32(Inst.imm,12)),32);
           R->RV32_PC += Inst.instSize;
         }else{
-          R->RV64[Inst.rd] = dt_u32((int32_t)(td_u32(R->RV64[Inst.rs1],32)) + (int32_t)(td_u32(Inst.imm,12)),32);
+          R->RV64[Inst.rd] = dt_u64(td_u64(R->RV64[Inst.rs1],64) + td_u64(Inst.imm,12),64);
           R->RV64_PC += Inst.instSize;
         }
         return true;
--- a/include/insns/RV32I.h
+++ b/include/insns/RV32I.h
@@ -257,7 +257,7 @@ namespace SST{
           R->RV32_PC += Inst.instSize;
         }else{
           R->RV64[Inst.rd] = 0x00;
-          R->RV64[Inst.rd] = (Inst.imm << 12);
+          SEXT(R->RV64[Inst.rd], Inst.imm << 12, 32); 
           R->RV64_PC += Inst.instSize;
         }
         return true;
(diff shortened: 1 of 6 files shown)


include/insns/RV32I.h
```diff
@@ -257,7 +257,7 @@ namespace SST{
           R->RV32_PC += Inst.instSize;
         }else{
           R->RV64[Inst.rd] = 0x00;
-          R->RV64[Inst.rd] = (Inst.imm << 12);
+          SEXT(R->RV64[Inst.rd], Inst.imm << 12, 32);
```
donofrio, line 260: When running RV64, instruction lui	sp,0xffff8
		* Results in :0x00000000ffff8000
		* Correct result: 0xffffffffffff8000
		* Solution: needs SEXT on immd after shift

src/RevProc.cc
```diff
@@ -550,6 +550,7 @@ RevInst RevProc::DecodeCIInst(uint16_t Inst, unsigned Entry){
       // sign extend
       CompInst.imm |= 0b11111111111111000000000000000000;
     }
+    CompInst.imm >>= 12;  //immd value will be re-aligned on execution
```
donofrio, line 553: There is a difference in how c.lui and lui handle immd values - lui assumes they are aligned to the LSB position and because the implementation of c.lui calls lui this results in the immd in c.lui being left-shited twice. So compensating here in the decode by aligning the c.lui immd to the LSB position

include/insns/RV64I.h
```diff
@@ -32,7 +32,7 @@ namespace SST{
                         RevMem *M, RevInst Inst) {
         // c.swsp rs2, $imm = sw rs2, x2, $imm
         Inst.rs1  = 2;
-        Inst.imm = ((Inst.imm&0b11111)*8);
+        ZEXT(Inst.imm, ((Inst.imm&0b11111))*8, 32);
```
donofrio, line 36: The RV spec indicates that the result must be zero-extended.

include/insns/RV32I.h
```diff
@@ -552,7 +552,7 @@ namespace SST{
           R->RV32[Inst.rd] = dt_u32((int32_t)(td_u32(R->RV32[Inst.rs1],32)) + (int32_t)(td_u32(Inst.imm,12)),32);
           R->RV32_PC += Inst.instSize;
         }else{
-          R->RV64[Inst.rd] = dt_u32((int32_t)(td_u32(R->RV64[Inst.rs1],32)) + (int32_t)(td_u32(Inst.imm,12)),32);
+          R->RV64[Inst.rd] = dt_u64(td_u64(R->RV64[Inst.rs1],64) + td_u64(Inst.imm,12),64);
           R->RV64_PC += Inst.instSize;
```
donofrio, line 556: copy paste error  - the RV64 implementation was a copy of the RV32 and was using 32-bit operands rather than 64bit

test/isa/add.c
donofrio, line 1: This is the test that caught the ISA errors... it is NOT complete and depends on the write() function  still in development. Adding here only to demonstrate how the ISA bugs were found, will continue to expand this (and other) asm tests

jleidel:
All changes look great.  merging to devel.

PR #57, opened 2023-05-22
System Call Support & Software Threading

rkabrick:
### Changes
0. Renamed all `thread` variables to Hart due to them being representative of Hardware Threads and would be easily confused with Software threads. 
1. Introduction of `RevThreadCtx` (ie. Software Threads)
2. Rudimentary support for exception handling via `RV(32|64)_SCAUSE` Registers (Only exception is for SystemCalls)
- When a call to `ecall` is made, the instruction simply sets a flag in the relevant `SCAUSE` register, saves the current PC in the `SEPC` register, and returns
- Upon returning (inside `RevProc::ClockTick`), we check if the `_SCAUSE` register has been set (indicating an ecall instruction happened) and then perform `RevProc::ExecEcall` which will find the corresponding function pointer inside the `RevProc::Ecalls` table based on the value inside the `a7` register per the spec.
3. New Interface to Register Files
- Instead of RegFile being a vector of length `numThreads` where each thread (Hart) has a single static register file, the register files of each Hart are dynamic and simply point to a RevThreadCtx's Register File. 
- `RevProc::RegFile()` and `RevProc::RegFile(HartID)` functions index into `RevProc::ThreadTable` to find a given Hart's active RegFile
4. `common/syscalls.h` file has been added which contains all of the c functions with corresponding inline assembly needed to circumvent need for `riscv*-unknown-linux` compilers to compile code with system calls in them (libc implementations introduced a tremendous amount of headaches and `riscv64-unknown-elf`'s only flaw was a lack of system call support. All system calls (with some exceptions) can be compiled by using them as one normally would and simply prefixing the call with `rev_`. (ie. `rev_chdir`, `rev_fork`, etc.)
---

### Notes:
- Ecalls currently only support RV64 but will support RV32 with relatively minor, straightforward changes
- Must include `common/syscalls.h` in any file that uses system calls and you must only use the `rev_` prefixed names
- If a systemcall that is not in `RevProc::Ecalls` is used SST will crash but the last line of output before the error message should tell you what the ecall number that was reached for was 
- Not all ecalls that are in `RevProc::Ecalls` actually do something other than return... these will continued to be implemented as time goes on
- The `Mem*` variables in a `RevThreadCtx` serve no purpose right now but will in the future
- Verbosity levels have been chosen somewhat arbitrarily but for the most part you should see relevant info with `verbose` set to 2 and even more so at 6
- In RISC-V, `fork` uses `clone` under the hood simply without any customization over the resources shared between parent and child. Until demand paging is integrated it's impractical to copy the parent's entire memory space so only `rev_fork` is supported and `rev_clone3` will give the same behavior as `rev_fork`. 
-  Several system calls will pass their arguments to the host (ie. most file related calls such as `write`, `read`, `chdir`, etc. as Rev has no filesystem of its own and it would be extremely impractical to have to keep files inside of simulated memory)
- In general, everything should be sufficiently commented with its use or lack thereof in the current state of this code


include/insns/RV64P.h
```diff
@@ -11,6 +11,7 @@
 #ifndef _SST_REVCPU_RV64P_H_
 #define _SST_REVCPU_RV64P_H_
 
+#include "../RevExt.h"
```
jleidel, line 14: this is broken.  never use relative header dirs

PR #69, opened 2023-06-22
Adding TLB & Corresponding Statistics

rkabrick:
This PR adds TLB functionality by creating two data structures (one used as the actual TLB) and the other `LRUQueue` used for tracking the Least Recently Used elements inside the TLB. Once the TLB is over the configured `tlbSize` added as a parameter to the RevCPU params in the python file, the least recently used element is removed. 

All TLB computation occurs inside `RevMem::CalcPhysAddr` via the `SearchTLB` and `AddToTLB` methods. 

Additionally, the TLBHits & TLBMiss statistics were added.


src/RevMem.cc
```diff
@@ -149,24 +153,68 @@ unsigned RevMem::RandCost( unsigned Min, unsigned Max ){
...
-  if(pageMap.count(pageNum) == 0){
-    // First touch of this page, mark it as in use
-    pageMap[pageNum] = std::pair<uint32_t, bool>(nextPage, true);
-    physAddr = (nextPage << addrShift) + ((pageSize - 1) & Addr);
+uint64_t RevMem::SearchTLB(uint64_t vAddr){
+  auto it = TLB.find(vAddr);
+  if (it == TLB.end()) {
+      // TLB Miss :(
+      memStats.TLBMisses++;
+      return -1;
```
donofrio, line 161: Returning signed value on a function with unsigned return type, would prefer a constant like INVALID_ADDRESS or something similar

src/RevMem.cc
```diff
@@ -149,24 +153,68 @@ unsigned RevMem::RandCost( unsigned Min, unsigned Max ){
...
+    }
+    // Insert the vAddr and physAddr into the TLB and LRU list
+    LRUQueue.push_front(vAddr);
+    TLB.insert({vAddr, {physAddr, LRUQueue.begin()}});
+  }
+}
+
+uint64_t RevMem::CalcPhysAddr(uint64_t pageNum, uint64_t vAddr){
+  uint64_t physAddr = SearchTLB(vAddr);
+  if( physAddr == -1 ){
```
donofrio, line 199: same comment as above - signed / unsigned match, can re-use the constant val

include/RevMem.h
donofrio, line 1: May want to add the ability to flush the TLB

donofrio:
Updates look good, I'm good to merge once John has finished his review

jleidel (approved):
approving the changes

PR #70, opened 2023-06-24
Fptest

donofrio:
Updates to floating point instruction operation, and most importantly, updates to how instructions are encoded in the instruction table.  Also now includes RV64F extension in instruction table.


include/insns/RV32F.h
```diff
@@ -176,16 +177,16 @@ namespace SST{
 
       static bool fnmadds(RevFeature *F, RevRegFile *R,RevMem *M,RevInst Inst) {
         if( F->IsRV32D() ){
-          R->DPF[Inst.rd] = (float)((-(float)(R->DPF[Inst.rs1])) *
-                                    (float)(R->DPF[Inst.rs2]) +
+          R->DPF[Inst.rd] = (float)(-(((float)(R->DPF[Inst.rs1])) *
```
jleidel, line 180: looks like `fnmadds` is doing subtraction and `fnmsubs` is doing addition.  Is this correct?

donofrio: The "N" in that opcode name is the clue... it is non-obvious, but FNMADD.S computes -(rs1×rs2)-rs3 and FNMSUB.S computes -(rs1×rs2)+rs3.

PR #71, opened 2023-06-26
Merging major functional update to devel

jleidel:
merging major functional update to devel.  Includes:
- major fixes to crack & decode
- numerical stability updates to integer and sp/dp float
- addition of memHierarchy infrastructure
- improved statistics
- CMake build infrastructure
- CTest test harness


rkabrick:
I'm going to go ahead and try to clean up existing comments and add ones to anything that may be ambiguous. Not changing any actual code as I believe `devel` is in an okay state.

donofrio:
I'm going to search for any spurious std::cout messages and remove those, but agree the code in devel is in good shape

donofrio:
OK - squashed a few couts, I'm good to merge whenever Ryan is ready

src/RevProc.cc
```diff
@@ -1581,45 +1801,89 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
+      RegFile->cost = 0; // We failed dependency check, so set cost to 0 - this will
+      Stats.cyclesIdle_Pipeline++;        // prevent the instruction from advancing to the next stage
+      HART_CTE[HartToDecode] = false;
+      HartToExec = _REV_INVALID_HART_ID_;
+    }else {                 
+      Stats.cyclesBusy++;
+      HART_CTE[HartToDecode] = true;
+      HartToExec = HartToDecode;
+    };
+    Inst.cost = RegFile->cost;
```
donofrio, line 1863: I believe this should be InstTable[Inst.entry].cost

jleidel: RHS of the operation?  The cost is set during decode in order to track the lifetime of the pipeline cost. Maybe I'm missing something though.

rkabrick:
> I'm going to go ahead and try to clean up existing comments and add ones to anything that may be ambiguous. Not changing any actual code as I believe `devel` is in an okay state.

So upon looking through everything (specifically my ThreadCtx stuff, I realized that with the pending MemSegment changes, all of the `MemSize` and `MemStartingAddr` along with their corresponding getters and setters are not only unused, but are also going to be completely changed. For that reason, I think I will change some code but logically it won't change anything. 

Basically a prune of the current Memory logic inside the RevThreadCtx object which is *unused*

src/RevProc.cc
donofrio: Initially I was worried that the RegFile.cost may not be updated correctly, but I ran a few tests and it appears to work correctly, so I'm good with leaving as-is.

jleidel:
are we clear to merge?

donofrio:
Clear to merge here

rkabrick:
Looks good to meOn Jun 28, 2023, at 16:51, donofrio ***@***.***> wrote:﻿
Clear to merge here

PR #74, opened 2023-06-29
X0fix

donofrio:
Replacing the array based register file with a template class that forces reg0 to always be 0. Adding qsort and memcpy benchmark tests. Adding test for use of x0.


jleidel:
Have you tested the performance of big_loop before/after the change?

donofrio:
Impact seems to be minimal - with and without the change performance varies between 3.9 and 4.2 sec on my laptop

jleidel (approved):
we can fix the one comment once this lands on `devel`

PR #89, opened 2023-08-02
Dynamic Memory Allocation (Minor Loader Changes)

rkabrick:
This PR enables the following:
- `rev_mmap` for allocating dynamic memory 
- `MemSegs` vector and `MemSegment` class which is used for managing 'valid' memory
- The loader now adds valid memory segments based on the ELF Program Headers & Section Headers via the `InitStaticMem` function & modifications to the `LoadElf(32|64)` functions
- `maxHeapSize` configuration option to the python interface

### DIFF
--- a/include/RevMem.h
+++ b/include/RevMem.h
@@ -94,7 +94,7 @@ namespace SST {
           State = "| Allocated | "; 
         }
         std::cout << "---------------------------------------------------------------" << std::endl;
-        return os << State << " | 0x" << obj.getBaseAddr() << " | 0x" << obj.getTopAddr() << " | Size = " << obj.getSize();
+        return os << State << " | 0x" << std::hex << obj.getBaseAddr() << " | 0x" << std::hex << obj.getTopAddr() << " | Size = " << std::dec << obj.getSize();
       }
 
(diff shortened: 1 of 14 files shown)


include/RevMem.h
```diff
@@ -54,6 +58,61 @@ namespace SST {
...
+          void setIsFree(bool freeState){ IsFree = freeState; }
+
+
+      // Custom Print Function
+      friend std::ostream& operator<<(std::ostream& os, MemSegment& obj) {
+        std::string State = "| Free |";
+        if( !obj.IsFree ){
+          State = "| Allocated | "; 
+        }
+        std::cout << "---------------------------------------------------------------" << std::endl;
```
jleidel, line 102: remove the `std::cout` or block it with `ifdef`

test/tls/basic-tls.c
jleidel, line 1: we should probably make these header locations relative at some point.

jleidel (approved):
We can clean up a few things once merged, but i approve for now.

donofrio:
Looks good and a major functionality upgrade. Approved

PR #90, opened 2023-08-03
Remove make-based builds

leekillough:
Remove `make`-based builds as obsolete; only support CMake builds now

Reword and reformat `README.md`
- remove `make` instructions and only mention `cmake`
- replace ` `` ` with `` ` `` for inline code
- wrap paragraphs at 80 characters
- remove `$` from the beginning of command-line examples to make it easier to copy-and-paste
- add accepting of `LICENSE` to the contributor requirements
- remove comment about not allowing C++11 or later

Remove `Makefile` and `src/Makefile`

Make `build/` directory a regular part of the tree with a single `.gitignore` file, removing `build` from the top-level `.gitignore` file

Tweak `.gitignore` to ignore editor temporary files

Set `CXX_STANDARD` CMake variable instead of using explicit `-std=c++17` flag


donofrio:
Thanks for the cleanup! Will merge to devel

Issue #84, opened 2023-07-20
[P2] Broken division of negative numbers


ukasz:
We observe calculation errors on negative number division, see example below.
```
#define assert(x) if (!(x)) { asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); }
int main(){
        int d = -20;
        int x = -3;
        int div = d/x;
        assert(div == 6);
        return 0;
}
```

leekillough:
I cannot reproduce this. I [added a test](https://github.com/tactcomplabs/rev/pull/93).

leekillough:
Fixed by #115

PR #93, opened 2023-08-08
Add Divw2 test

leekillough:
Adds https://github.com/tactcomplabs/rev/issues/84


jleidel:
Lee, all PR's should target `devel`

PR #94, opened 2023-08-08
updating LRSC functionality to support memHierarchy

jleidel:
updating LRSC functionality to support memHierarchy

- Updates the memory interfaces and outstanding tag tracking for LRSC
- Adds AQ/RL support to request interfaces

### DIFF
--- a/src/RevMem.cc
+++ b/src/RevMem.cc
@@ -122,23 +122,110 @@ bool RevMem::StatusFuture(uint64_t Addr){
   return false;
 }
 
-bool RevMem::LR(unsigned Hart, uint64_t Addr){
-  std::pair<unsigned,uint64_t> Entry = std::make_pair(Hart,Addr);
-  LRSC.push_back(Entry);
+bool RevMem::LRBase(unsigned Hart, uint64_t Addr, size_t Len,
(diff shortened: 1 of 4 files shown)


src/RevMem.cc
```diff
@@ -122,24 +122,111 @@ bool RevMem::StatusFuture(uint64_t Addr){
...
+bool RevMem::LRBase(unsigned Hart, uint64_t Addr, size_t Len,
+                    void *Target, uint8_t aq, uint8_t rl,
+                    StandardMem::Request::flags_t flags){
+  std::vector<std::tuple<unsigned,uint64_t,unsigned,uint64_t*>>::iterator it;
+
+  for( it = LRSC.begin(); it != LRSC.end(); ++it ){
+    if( (Hart == std::get<LRSC_HART>(*it)) &&
+        (Addr == std::get<LRSC_ADDR>(*it)) ){
+      // existing reservation; return w/ error
+      uint32_t *Tmp = (uint32_t *)(Target);
```
donofrio, line 134: given Target is a void* we may want to do c++ casting here and check for null to make sure it succeeds

jleidel:
kiling this PR as all upstream changes will induce lots of updates outside of LR/SC

PR #99, opened 2023-08-14
Loader Changes & Misc Cleanup

rkabrick:
- Making the logic behind the loader assigning HeapStart / HeapEnd more robust so it bases this address off of section headers. It looks for `.bss`, then `.data`, then `.text` as a last resort
- Cleaned up some other code in RevLoader (No logical changes)
- Removed some unused code from RevMem
- Added an `InitHeap` function that acts solely as a cleaner interface for the loader to assign the HeapStart & HeapEnd addresses

### DIFF
--- a/src/RevLoader.cc
+++ b/src/RevLoader.cc
@@ -159,6 +159,9 @@ bool RevLoader::LoadElf32(char *membuf, size_t sz){
     // Print information about the program header
     // std::cout << "Section Header " << i << ": " << sh[i].sh_type << ", Address: " << sh[i].sh_addr << ", Size: "  << std::endl;
     mem->AddMemSeg(sh[i].sh_addr, sh[i].sh_size, true);
+//    if( sh[i].sh_name ){
+//      std::cout << "Section Header " << i << ": " << sh[i].sh_type << ", Address: " << sh[i].sh_addr << ", Size: " << sh[i].sh_size << ", Name: " << shstrtab + sh[i].sh_name << std::endl;
+//    }
   }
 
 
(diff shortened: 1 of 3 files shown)


src/RevLoader.cc
```diff
@@ -143,24 +143,46 @@ bool RevLoader::LoadElf32(char *membuf, size_t sz){
...
+    StaticDataEnd = BSSEnd;
+  } else if ( DataEnd > 0 ){
+    // BSS Doesn't exist, but data does
+    StaticDataEnd = DataEnd;
+  } else if ( TextEnd > 0 ){
+    // Text is last resort 
+    StaticDataEnd = TextEnd;
+  } else {
+    // Can't find any (Text, BSS, or Data) sections
+    output->fatal(CALL_INFO, -1, "Error: No text, data, or bss sections --- RV64 Elf is unrecognizable\n" );
```
donofrio, line 184: Noob question - are these sections always exclusive?

src/RevMem.cc
```diff
@@ -877,6 +832,17 @@ uint64_t RevMem::ShrinkMemSeg(std::shared_ptr<MemSegment> Seg, const uint64_t Ne
   }
 }
 
+/// @brief This function is called from the loader to initialize the heap
+/// @param EndOfStaticData: The address of the end of the static data section (ie. end of .bss section)
+void RevMem::InitHeap(const uint64_t& EndOfStaticData){
+  if( EndOfStaticData <= 0 ){
+    output->fatal(CALL_INFO, 7, "EndOfStaticData = 0x%lx which is less than or equal to 0. This is a bug.", EndOfStaticData);
```
donofrio, line 839: Is this reachable since EndOfStaticData is unsigned? Or is there a risk the compiler may cast it as signed (given "0" is ambiguous) which would give strange results

jleidel: Theoretically reachable if `EndofStaticData == 0`; but not otherwise..... I would suggest revising to: 
```
if( EndOfStaticData <= 0x00ull )
```

jleidel (approved):

src/RevLoader.cc
rkabrick: Yes. The same sections may be present within the same segment however as I understand it:
- `.text`: Contains the actual executable code (ie. Always present)
- `.data`: Contains initialized global and static variables 
- `.bss`: Contains uninitialized global and static variables

The latter two are absent from the `isa` tests, however, should be present in most other programs.

src/RevMem.cc
rkabrick: Fixed in most recent push

PR #100, opened 2023-08-16
Adding explicit AMO support

jleidel:
Supports AMOs through to RevMem and memHierarchy.


rkabrick:
As a general question, these operations are based on the HartID... Should they eventually take in the ThreadID instead? If not, I imagine the upcoming SMT logic will just prevent context switching while a given Thread has an AMO instruction in flight. Otherwise, you'd be returning values to a potentially incorrect register file.

jleidel:
We have identified the bug in using memHierarchy with AMOs.  Its a pipeline hazarding bug.  We're working on a fix.  Please do not merge this until we push the fix and test fully

PR #104, opened 2023-08-18
Overhaul of Rev, especially to fix handling of floating-point and integer conversions

leekillough:
This Rev change gets rid of many unnecessary casts and complicated sign-extension code, and turns lots of functionality into C++ functions instead of long `if-else` statements with duplicated code. It fixes a lot of bugs.

# Compiler warnings and errors

Add nearly all GCC warnings (`-Wall -Wextra -Werror`), including `printf`/`scanf` format mismatches, and make warnings errors. Suppress some warnings in SST headers by putting all SST `#includes` in a new `SST.h` Rev header.

Fix over a dozen `printf` format mismatches. Make the banner string a `const char[]` array instead of a `const char*` pointer, to avoid the GCC warning about insecure string pointers being passed as `printf` formats.

Enforce `-Wdouble-promotion`, `-Wnarrowing` as errors, to prevent unexpected conversions or truncations. `-Wconversion` was tried, and many of its warnings were removed, but there are too many places in Rev where this warning occurs, so that will have to be fixed in another PR.

# Use C++ Best Practices

## Class / enum names instead of `typedef`

Instead of:
```
typedef struct {
} Name;
```
Or:
```
typedef enum {
} Name;
```
Use:
```
struct Name {
};

enum Name {
};
```
In C++, `struct` and `enum` tags create a type of the same name.

## Use the [Rule of 3/5/0](https://en.cppreference.com/w/cpp/language/rule_of_three) (C++11):

(3) If the copy constructor, copy assignment operator, or destructor are user-declared, all should be user-declared, even if only with `= default` or `= delete`
(5) If the Rvalue move constructor or move assignment operator are declared, also declare the copy constructor, copy assignment operator, and destructor, even if only with `= default` or `= delete`
(0) For aggregates lke C `structs` (Plain Old Datatypes), do not specify user-declared constructors/destructors/assignment operators

Use `= default` declarations instead of empty destructors or handwritten constructors / assignment operators with default semantics.

Use `= delete` declarations instead of `private` or `protected` to remove default constructors or assignment operators (or, more rarely, destructors).

## Exploit move semantics

When using heavy data structures like `std::string` or `std::vector` whose values will be moved, pass them by value instead of by reference, and then use `std::move` to move them to their final destination. This moves deep copying up the call stack, doing only shallow moves at the lower levels, and allows shallow moves at all levels if the value truely is an Rvalue.

| Passing by reference                           | Passing by value                         |
|------------------------------------------------------|-------------------------------------------------|
| `void func(const std::vector& vector_ref){` <br />`    vector_var = vector_ref;` <br />`}` | `void func(std::vector vector_val){` <br />   `    vector_var = std::move(vector_val);` <br />`}` |
| `vector_var` is always deep-copied, even if `vector_ref` was passed an rvalue and can be moved instead of copied. | `vector_val` can be deeply or shallowly constructed depending on whether the argument is an rvalue. `vector_var` is always initialized by shallow move. |

## Explicitly Declare Enums

Use explicit declarations of underlying integer types, for both scoped (`enum class`) and unscoped (`enum`) enums (affects `RevFeatureType`, `EXCEPTION_CAUSE`, `RevInstF`, `RevRegClass`, `RevImmFunc`, etc.).

Use enum types instead of underlying integer types (or just `int`) when declaring variables and parameters of enum type. Unscoped enums can decay into their underlying type, while scoped enums cannot. Neither scoped nor unscoped enums can be set to integer types without a cast.

Use a trailing comma on the last enum so that tables can be extended easily without having to insert a comma and lining up the text formatting.

## Misc

Add default non-static class member definitions (`Type mem = val;`) if it is clearer or protects against accidental uninitialization in constructors. In the case of `RevInstDefaults`, the whole class was cut in half in size because initializers replaced the need for a constructor.

For very small functions such as constructors, make them inline by defining them in the header file instead of declaring them in the header file and defining them in a `.cpp` file, making optimization (e.g. inlining) easier.

Make constructors with single parameters `explicit` unless implicit conversion is desired.

When passing a C++ reference as a function parameter, use `const` when the function does not need to modify it.

Do not use `using namespace XYZ::ABC;` in a header file. Use it in source files (but never `using namespace std;`). Headers can use `using mytype = XYZ::ABC:type;` or `using MyBase::Member;`.

# `RevFeature`

For `RevFeatureType` enum type, use bit masks to allow for any set of features.

Inline most query functions (`IsModeEnabled()`, `GetMachineModel()`, `IsRV32()`, `IsRV64F()`, etc.). Add new query functions `HasF()`, `HasD()`.

Rewrite the parser, allowing multi-character option strings as long as they follow the RISC-V canonical ordering (Section 27.11 of the Unpriviledged ISA). Future-proof it with non-ratified extensions as long as they are listed in 27.11.

An out-of-order misspelling of the architecture string was found and fixed in the `strlen_c` and `strlen_cxx` tests.

The requirement for spelling in canonical order is specified in 27.11 of the spec, and GCC and Clang enforce it if you try an invalid `-march=` string. Furthermore, having a canonical ordering allows the parser to work in linear instead of quadratic time of the size of the table.

# `RevRegFile` Overhaul

The `RevRegFile` struct has been overhauled.

Default constructors, assignment operators, and destructors are now explicitly declared `= default`.

The `RevRF` proxy class, which overloaded `operator[]` to provide access to registers by subscript while protecting `x0` from being overwritten, has been replaced with a `GetX()/SetX()` templated mechanism.

`GetX<T>(F, rs)` returns register `rs` cast to type `T`, or `0` if `rs == 0`. `XLEN == 32` and `XLEN == 64` are handled automatically, greatly reducing the size of code.

`SetX(F, rd, val)` sets register `rd` to `val` if `rd != 0`, automatically casting `val` to `int32_t` or `int64_t` depending on `XLEN`.

`GetFP32(F, rs)` returns the `float` value of register `rs`, automatically NaN-unboxing it on machines with `FLEN > 32`. This means making sure the upper bits are all 1's, and returning `NAN` (quiet NaN `float`) if they are not.

`SetFP32(F, rd, value)` sets register `rd` to the `float value`, automatically NaN-boxing it on machines with `FLEN > 32`.

`AdvancePC(F, bytes)` advances the program counter `bytes` bytes, regardless of whether `XLEN == 32` or `XLEN == 64`. This gets rid of hundreds cases like this:

```
if( F->IsRV32() ){
  R->RV32_PC += Inst.instSize;
} else {
  R->RV64_PC += Inst.instSize;
}
```

## Helper functions

`ImmSignExt(n)` is a helper member of `RevInst` which returns a sign-extended value formed by the bottom `n` bits of `imm`. It is only guaranteed to return up to `int32_t`, but if it is converted to `int64_t` it will continue the sign-extension. This is used to extract sign-extended immediate operands of various lengths from the instruction encoding.

`CvtFpToInt()` is a C++ template function which implements 8 instructions: `{ float, double } -> { int32_t, uint32_t, int64_t, uint64_t }`. It uses the out-of-range handling described in Table 11.4 in the RISC-V F specification (no similar table is shown in the D or Q specifications, which seems like a simple oversight).

`fclass()` is a template function which supplements `std::fpclassify()` to make it implement the RISC-V `fclass` instruction's semantics, such as by distinguishing between quiet and signaling NaN.

Throughout the code, there are many instances where long functions are templatized into one, and then references to instantiations of that template are used to replace the long functions. The way to do this is to use a `constexpr` reference to the function, as in:

```
      static constexpr auto& beq  = bcond<std::equal_to>;
      static constexpr auto& bne  = bcond<std::not_equal_to>;
      static constexpr auto& blt  = bcond<std::less,          std::make_signed_t>;
      static constexpr auto& bltu = bcond<std::less,          std::make_unsigned_t>;
      static constexpr auto& bge  = bcond<std::greater_equal, std::make_signed_t>;
      static constexpr auto& bgeu = bcond<std::greater_equal, std::make_unsigned_t>;
```

This implements 6 branch instructions from one template. The `static` makes it a static member not related to a particular class instance; the `constexpr` makes it a constant expression initialized at compile-time (no out-of-class definitions are required in C++17 or later, since it's also considered `inline`); the `auto` makes it unnecessary to declare the full function prototype here; the `&` makes this a reference, so its name is an alias to a function (or object) as if were instantiated with those template arguments.

## Atomic Memory Operations (`AMO`)

The very long, hard-to-read code for executing atomic operations based on flags, has been greatly simplified, using tables and C++ lambdas. What used to require complex `if`-`else` sequences, is now table-driven, and the `XLEN==32` and `XLEN==64` distinctions, are handled by two separate calls to the same generic lamba.

9 atomic instructions in both `RV32I` and `RV64I`, have been condensed into a single template.

## Miscellaneous

Rename `RevInstTables.h` (plural) to `AllRevInstTables.h`, so that keyboard autocompletion can find `RevInstTable.h` (singular) without ambiguity.

Add `.gdb_history` to `.gitignore` since GDB sessions often create it in working tree.

Do not overwrite `RVCC` environment variable with blank CMake `RVCC` variable, allowing `RVCC` to be set either on the CMake command-line or in the environment.

Add C++ standard header `#include`s where they are missing, even if they get indirectly `#include`ed already.

Add spaces after commas in function calls and declarations/definitions. i.e. instead of `F(A, B,C,D, E)`, normalize this `F(A, B, C, D, E)`.

Casts to `(void *)` are almost never needed, since they are cast implicitly from other pointers. There are also a lot of other useless casts to the same type removed, and `-Wuseless-cast` has been added to the compiler flags, `-Wold-style-cast` was tested, and all old-style casts eliminated from Rev, but `-Wold-style-cast` could not seem to be ignored a with `#pragma` during the SST `#include`s, perhaps because there were old-style casts in macros defined outside of the confines of the `#pragma`, so `-Wold-style-cast` is turned off for now.

There were some types like `tlbSize` and `maxHeapSize` which were entended to 64-bit variables because `memSize` and other were already 64-bit, and I was getting conversion warnings of going back and forth between 32- and 64-bits without explicit casts.

`RevThreadCtx::ChildrenPIDs` and `RevThreadCtx::fildes` were changed from `std::vector<>` to `std::set<>`, since the only thing that was really needed was membership information querying and setting, not a contiguous array of values.

In `RevCPU.cc` there were several cases of `new` being checked for `nullptr` even though `std::nothrow` was not used, and hence `new` should never return `nullptr`. I added `TODO:` comments warning about them.

In `RevExt.cc` there were a lot of debug output statements commented out by `#if 0`. I changed them all to `#if TRACE_EXECUTE` so that the definition of `TRACE_EXECUTE` was in one place.

Explicitly-sized types like `int32_t`, `int64_t`, `uint32_t` and `uint64_t` are preferred to `int`, `long`, `unsigned long`, etc. For some domains, `size_t` or `ptrdiff_t` are preferred.

In `RevMem.cc` and perhaps others, there were fatal error messages which were not terminated by newlines consistently.

In `RevProc.cc` a variable-length-array (VLA) was used which is a non-standard extension, and which puts the allocation on the stack instead of the heap, increasing the risk of stack overflows. I replaced it with `std::vector<char>`.

The tests in `test/isa/fadd.c` had numerical rounding issues which made the tests fail even on `pinhead`. The number of nonzero bits in a `float` needs to be set low enough that the exact result can be obtained.


jleidel:
Closing this PR, retarget to the `Lee` branch

PR #106, opened 2023-08-22
Co-processor functionality

donofrio:
Adding basic co-processor functionality to Rev


test/coproc_ex/rev-test-coproc_ex.py
jleidel, line 1: we should also add a test in the near future that includes memHierarchy

include/RevCoProc.h
```diff
@@ -0,0 +1,154 @@
...
+    protected:
+      SST::Output *output;                                ///< RevCoProc: sst output object
+    }; // class RevCoProc
+
+    // ----------------------------------------
+    // RevSimpleCoProc
+    // ----------------------------------------
+    class RevSimpleCoProc : public RevCoProc{
+    public:
+      SST_ELI_REGISTER_SUBCOMPONENT_DERIVED(RevSimpleCoProc, "revcpu",
```
jleidel, line 74: `SST_ELI_REGISTER_SUBCOMPONENT_DERIVED` is deprecated in SST 13.0.0.  This should be `SST_ELI_REGISTER_SUBCOMPONENT`

jleidel: Fixed in fc635e3

PR #107, opened 2023-08-23
Memseg cleanup

rkabrick:
### FreeMemSegs
- Added a FreeMemSegs vector which in turn simplified quite a bit of the AddMemSeg code
The FreeMemSegs vector occupies all the space after StaticDataEnd (ie. HeapSpace) up to `StaticDataEnd + MaxHeapSize`.
- Removed the `isFree` boolean from `MemSegment` class
- When checking if a virtual address is valid via the `isValidVirtAddr` this cuts down significantly on overhead from checking every allocated segment and then checking if it's free... Now, once a segment is free, we don't have to bother looking through it

### Internal functions (User cannot trigger)
**AddMemSegAt(BaseAddr, Size)**
- Add mem seg starting at BaseAddr of size 

**AddRoundedMemSeg(BaseAddr, Size, RoundUpSize)**
- Add mem seg at BaseAddr of Size rounded up to the nearest multiple of RoundUpSize (this is used mainly in the loader where the headers get added and rounded up to the nearest multiple of PAGE_SIZE

### System Call Triggerable Functions (Alloc/Dealloc)

**AllocMem(SegSize)** 
- Attempts to find a space >= SegSize inside of the FreeMemSegs vector 
- If a FreeSeg is found, we shrink FreeSeg by moving its baseAddr up to `FreeSeg->getBaseAddr()+NewSegSize` and then add the NewSeg (w/ `BaseAddr = OldFreeSeg->BaseAddr` ) to the MemSegs (ie. allocated segment vector)
- If `FreeSeg->getSize == SegSize`, we simply move that segment from the `FreeMemSegs` vector to the `MemSegs` vector 
- If there is no `FreeMemSeg` big enough, we add it to the `HeapEnd` and expand the heap

**AllocMemAt(BaseAddr, SegSize)**
- Checks `FreeMemSegs` to see if we can fit the segment where the user requested
- If it finds a segment big enough, handle the four cases:
	1. New Segment is the same size as FreeSeg found
	2. New Segment fits in the middle of the FreeSeg
	3. NewSegment is allocated at the beginning of the FreeSeg (ie. `FreeSeg->getBaseAddr == BaseAddr`) 
        4. NewSegment is allocated at the end of the `FreeSeg` (ie. `FreeSeg->getTopAddr == BaseAddr + SegSize`)
       
*NOTE: In an actual OS, I believe Linux will try to fit the segment at the nearest address to the requested one if it can’t fit it exactly where the user wants it. We will not be supporting this as there are much higher priority items in flight. Consequently, if we can’t allocate a segment at the requested location we will end the simulation.*

**DeallocMem**:
- Logic in this one pretty much stayed the same
- The only difference is moving the deallocated memory to the FreeMemSegs vector instead of setting the (now removed) `isFree` flag for that segment

### DIFF
--- a/include/RevMem.h
+++ b/include/RevMem.h
@@ -87,7 +87,7 @@ namespace SST {
           // Check if a given range is inside a segment
           bool contains(const uint64_t& vBaseAddr, const uint64_t& Size){
             // exclusive top address
-            uint64_t vTopAddr = vBaseAddr + Size - 1; 
+            uint64_t vTopAddr = vBaseAddr + Size - 1;
             return (this->contains(vBaseAddr) && this->contains(vTopAddr));
           };
 
@@ -95,7 +95,7 @@ namespace SST {
       // Override for easy std::cout << *Seg << std::endl;
       friend std::ostream& operator<<(std::ostream& os, MemSegment& obj) {
         std::cout << "---------------------------------------------------------------" << std::endl;
-        return os << State << " | 0x" << std::hex << obj.getBaseAddr()
+        return os << " | 0x" << std::hex << obj.getBaseAddr()
                   << " | 0x" << std::hex << obj.getTopAddr()
                   << " | Size = " << std::dec << obj.getSize();
       }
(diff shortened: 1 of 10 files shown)


jleidel:
@rkabrick, need a quick merge cleanup on this.  Do you want me to tackle this?

Issue #56, opened 2023-05-19
[P0] Atomic increment doesn't work


ukasz:
We observe cpp atomic operations to fail on data sizes wider than char (tested sha: cd753f7 compiled with LLVM)
Simplest example of such behavior is depicted below.
```
#include <atomic>
#define assert(x) if (!(x)) { asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); }

int main() {
        std::atomic<int> a;
        a = 0;
        a++;
        assert(a == 1);
}
```

ukasz:
Even if atomic operations cannot be "atomic" yet, we would like to see their behavior emulated, so they are functionally working on a single thread.

jleidel:
the single threaded versions are functionally emulated using only the embedded memory infrastructure; currently adding support for these on the `amo` branch

jleidel:
fixed

PR #108, opened 2023-08-25
LoadHazards for memHierarchy

jleidel:
Creating load hazard infrastructure for all AMOs and LOAD operations.  Correctly ordering pipeline hazards.  Fix for Issue #56


rkabrick:
No issue for me... good to merge

src/RevMemCtrl.cc
```diff
@@ -1294,6 +1291,8 @@ void RevBasicMemCtrl::handleReadResp(StandardMem::ReadResp* ev){
         if( isAMO ){
           handleAMO(op);
         }
+        bool *Hazard = op->getHazard();
+        *Hazard = false;
         delete op;
```
donofrio, line 1296: So we allocated this in RevProc, and we will also be calling `destroyHazard` from there - do we want to delete this value here or just set to nullptr and have it cleaned up in RevProc?

donofrio: Agreed - revProc allocates so RevProc has to destroy

PR #109, opened 2023-08-26
Add initialization of uninitialized variables

leekillough:
This doesn't fix any current failures, but I noticed it when going through the code.


jleidel (approved):
reviewed and good to merge

PR #111, opened 2023-08-31
Update test/ makefiles to rely on RVCC

vcave:


jleidel (approved):
removing hard coded toolchain

PR #113, opened 2023-09-05
Adding Mem H support to syscalls 

donofrio:
Updated the ECALLs to support mem hierarchy as well as RevMem. Summary of changes:

 * ECALL now returns a status (SUCCESS, CONTINUE or ERROR) - on returning CONTINUE the PC is NOT incremented and the pipeline will call into the function again on the next clock. This allows us to assemble long read buffers over multiple cycles
 * For ECALLs that require multiple cycles a dependency is set on the return register
 * New Dependency Set / Clear functions added to allow setting / clearing with only the knowledge of HART, RegNum and if it is a float
 *  ECALLs track state between calls by looking at a few new bits of persistent state: 
-  ECALL_buf - stores data returned by a single call to ReadMem
- ECALL_string - buffers data over multiple calls to ReadMem
- ECALL_bytesRead - running total of number of bytes read by an ECALL function

* Added tests to use MemH for each of the syscall tests updated


test/syscalls/write/write.c
```diff
@@ -4,11 +4,24 @@
 int main() {
 
   const char msg[10] = "Greetings\n"; 
-  ssize_t bytes_written = rev_write(STDOUT_FILENO, &msg, sizeof(msg));
+  ssize_t bytes_written = rev_write(STDOUT_FILENO, msg, sizeof(msg));
 
   if( bytes_written < 0 ){
     rev_exit(1);
   }
+  const char msg2[67] = "Greetings Dave Greetings Dave Greetings Dave 123456789012345678901\n";
```
jleidel, line 12: Is this a leftover?

donofrio: yes and no, I did want to add successively longer strings to the write test to improve the test itself, I can change the string to be a bit less weird though

jleidel (approved):
good to merge

rkabrick:
What is the reason for intermixing `ReadVal` and `ReadMem`?

jleidel:
> What is the reason for intermixing `ReadVal` and `ReadMem`?

Can you reference a file + line number?

src/RevProc.cc
rkabrick, line 1: `ReadMem` used here

src/RevProc.cc
```diff
@@ -2520,175 +2564,201 @@ void RevProc::ECALL_brk(){
...
-  mem->ReadMem(CloneArgsAddr, sizeof(uint64_t), &args);
+ if(0 == ECALL_bytesRead){
+    // First time through the function... 
+    /* Fetch the clone_args */
+    // struct clone_args args;  // So while clone_args is a whole struct, we appear to be only
+                                // using the 1st uint64, so that's all we're going to fetch
+    uint64_t* args = reinterpret_cast<uint64_t*>(ECALL_buf);
+    mem->ReadVal<uint64_t>(HartToExec, CloneArgsAddr, args, inst.hazard, REVMEM_FLAGS(0x00));
+    ECALL_bytesRead = 8;
+    rtval = RevProc::ECALL_status_t::CONTINUE;
```
rkabrick, line 2586: `ReadVal` used here

src/RevProc.cc
```diff
@@ -2746,54 +2816,89 @@ void RevProc::ECALL_getcwd(){
...
+    rtv = RevProc::ECALL_status_t::SUCCESS;
+    DependencyClear(HartToExec, 10, false);
+
+  }else {
+    if(ECALL_bytesRead > 0){
+      //Not our first time through... so capture previous read data
+      ECALL_string = ECALL_string.append(ECALL_buf);
+    }
+    if(1 == (nbytes - ECALL_bytesRead)){
+      mem->ReadMem(HartToExec, RegFile->RV64[11] + ECALL_bytesRead, sizeof(char), &ECALL_buf[0], inst.hazard, REVMEM_FLAGS(0x0));
```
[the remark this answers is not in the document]
jleidel: Good catch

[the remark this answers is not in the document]
donofrio: ReadMem replaced with ReadVal. also tightened up the math for bytesRead to use sizeof().  Write test string replaced with something more sane

rkabrick (approved):

PR #115, opened 2023-09-07
Overhaul of Rev

leekillough:
This Rev change gets rid of many unnecessary casts and complicated sign-extension code, and turns lots of functionality into C++ functions instead of long `if-else` statements with duplicated code. It templatizes many functions to create instructions, leading to single-line implementations of instructions. It fixes a lot of bugs.

# Remaining concerns

The asynchronous load / hazard mechanism may still need to be cleaned up. 
- The sign/zero extension in `handleFlagResp()` did not seem right, because it did not consider the number of bytes in the request, and because the 64-bit sign/zero extensions were essentially no-ops. I have changed it to use `8 * op->getSize()` as the number of bits to extend to the input word size (`*target` in `handleFlagResp()`), but I am concerned about whether the reading code already pads to 4 or 8 bytes and if there is any dependence on endianness (of either the host machine or the simulator model). 
- The sign/zero extension, and the `NaN`-boxing which I have added to `flags`, are repeated twice, once on garbage (or a sentinel value if we choose to initialize `rd`) after the request is made, and once again when an asynchronous load finishes. I would like the same routine to be called in the synchronous case (`ctrl == nullptr`) so that we do not need to do it in the load instruction after the request is made, and so that the logic will be the same based on flags for both non-hierarchical and hierarchical cases (more test coverage).
- The atomics need more test coverage. There was a lot of copy-and-paste code which was very hard to read, which I distilled down to compact tables. I tried modifying the `AMOADD_CXX_MEMH` test to change the values of the variables, but that sometimes made the test fail.
- The documentation needs to be updated to mention the new functionality, which reduces the size of instruction simulation code by over 75%.

# Random Number Generator

The code before this PR was making frequent calls to `srand()` and `rand()`. The calls to `srand()` were taking 28% of the total time in `big_loop`!!!  See [this video](https://www.youtube.com/watch?v=LDPMpc-ENqY) for why using `rand()` is considered harmful.

A new `RevRand(min, max)` function has been created, which returns uniform random numbers between `[min,max]` for integers, and `[min,max)` for floating-point. The `std::common_type` of `min` and `max` is used to determine the result type, so `RevRand(0,sizeof(T)-1)` would, for example, return a `size_t` between `0` and `sizeof(T)-1`, inclusive. The `0` is automatically promoted to `size_t`, since it's the common type.

`RevRand()` uses the C++11 concept of separating "engines" from "distributions". An engine generates random numbers of arbitrary bits, while a distribution distributes them across a range for a given statistical distribution.

There is one `thread_local` engine per thread, which is `mt19937_64` (Mersenne twister). There is only one engine per thread even if there are multiple distributions or random number data types `(min, max)` used. The engine is initialized the first time a thread needs it, not at thread initialization or program startup. This reduces the overhead of creating new threads which do not use random number generation.

`RevRand()` is thread-safe, and initializes its `thread_local` engine's seed with hardware random number generators (`std::random_device`) and a hash function of the thread's id (`std::this_thread::get_id()`).

A bug in the random code was found where `2 ^ bits` was used instead of `1 << bits` for computing powers of 2. [See this history](https://softwareengineering.stackexchange.com/a/331392).

# Compiler warnings and errors

Add nearly all GCC warnings (`-Wall -Wextra -Werror`), including `printf`/`scanf` format mismatches, and make warnings errors. Suppress some warnings in SST headers by putting all SST `#includes` in a new `SST.h` Rev header.

Fix over a dozen `printf` format mismatches. Make the banner string a `const char[]` array instead of a `const char*` pointer, to avoid the GCC warning about insecure string pointers being passed as `printf` formats.

Enforce `-Wdouble-promotion`, `-Wnarrowing` as errors, to prevent unexpected conversions or truncations. `-Wconversion` was tried, and many of its warnings were removed, but there are too many places in Rev where this warning occurs, so that will have to be fixed in another PR.

# Use C++ Best Practices

## Class / enum names instead of `typedef`

Instead of:
```
typedef struct {
} Name;
```
Or:
```
typedef enum {
} Name;
```
Use:
```
struct Name {
};

enum Name {
};
```
In C++, `struct` and `enum` tags create a type of the same name.

## Use the [Rule of 3/5/0](https://en.cppreference.com/w/cpp/language/rule_of_three) (C++11):

(3) If the copy constructor, copy assignment operator, or destructor are user-declared, all should be user-declared, even if only with `= default` or `= delete`
(5) If the Rvalue move constructor or move assignment operator are declared, also declare the copy constructor, copy assignment operator, and destructor, even if only with `= default` or `= delete`
(0) For aggregates lke C `structs` (Plain Old Datatypes), do not specify user-declared constructors/destructors/assignment operators

Use `= default` declarations instead of empty destructors or handwritten constructors / assignment operators with default semantics.

Use `= delete` declarations instead of `private` or `protected` to remove default constructors or assignment operators (or, more rarely, destructors).

## Exploit move semantics

When using heavy data structures like `std::string` or `std::vector` whose values will be moved, pass them by value instead of by reference, and then use `std::move` to move them to their final destination. This moves deep copying up the call stack, doing only shallow moves at the lower levels, and allows shallow moves at all levels if the value truely is an Rvalue.

| Passing by reference                           | Passing by value                         |
|------------------------------------------------------|-------------------------------------------------|
| `void func(const std::vector& vector_ref){` <br />`    vector_var = vector_ref;` <br />`}` | `void func(std::vector vector_val){` <br />   `    vector_var = std::move(vector_val);` <br />`}` |
| `vector_var` is always deep-copied, even if `vector_ref` was passed an rvalue and can be moved instead of copied. | `vector_val` can be deeply or shallowly constructed depending on whether the argument is an rvalue. `vector_var` is always initialized by shallow move. |

## Explicitly Declare Enums

Use explicit declarations of underlying integer types, for both scoped (`enum class`) and unscoped (`enum`) enums (affects `RevFeatureType`, `EXCEPTION_CAUSE`, `RevInstF`, `RevRegClass`, `RevImmFunc`, etc.).

Use enum types instead of underlying integer types (or just `int`) when declaring variables and parameters of enum type. Unscoped enums can decay into their underlying type, while scoped enums cannot. Neither scoped nor unscoped enums can be set to integer types without a cast.

Use a trailing comma on the last enum so that tables can be extended easily without having to insert a comma and lining up the text formatting.

## Misc

Add default non-static class member definitions (`Type mem = val;`) if it is clearer or protects against accidental uninitialization in constructors. In the case of `RevInstDefaults`, the whole class was cut in half in size because initializers replaced the need for a constructor.

For very small functions such as constructors, make them inline by defining them in the header file instead of declaring them in the header file and defining them in a `.cpp` file, making optimization (e.g. inlining) easier.

Make constructors with single parameters `explicit` unless implicit conversion is desired.

When passing a C++ reference as a function parameter, use `const` when the function does not need to modify it.

Do not use `using namespace XYZ::ABC;` in a header file. Use it in source files (but never `using namespace std;`). Headers can use `using mytype = XYZ::ABC:type;` or `using MyBase::Member;`.

# `RevFeature`

For `RevFeatureType` enum type, use bit masks to allow for any set of features.

Inline most query functions (`IsModeEnabled()`, `GetMachineModel()`, `IsRV32()`, `IsRV64F()`, etc.). Add new query functions `HasF()`, `HasD()`.

Rewrite the parser, allowing multi-character option strings as long as they follow the RISC-V canonical ordering (Section 27.11 of the Unpriviledged ISA). Future-proof it with non-ratified extensions as long as they are listed in 27.11.

An out-of-order misspelling of the architecture string was found and fixed in the `strlen_c` and `strlen_cxx` tests.

The requirement for spelling in canonical order is specified in 27.11 of the spec, and GCC and Clang enforce it if you try an invalid `-march=` string. Furthermore, having a canonical ordering allows the parser to work in linear instead of quadratic time of the size of the table.

# `RevRegFile` Overhaul

The `RevRegFile` struct has been overhauled.

Default constructors, assignment operators, and destructors are now explicitly declared `= default`.

The `RevRF` proxy class, which overloaded `operator[]` to provide access to registers by subscript while protecting `x0` from being overwritten, has been replaced with a `GetX()/SetX()` templated mechanism.

`GetX<T>(F, rs)` returns register `rs` cast to type `T`, or `0` if `rs == 0`. `XLEN == 32` and `XLEN == 64` are handled automatically, greatly reducing the size of code.

`SetX(F, rd, val)` sets register `rd` to `val` if `rd != 0`, automatically casting `val` to `int32_t` or `int64_t` depending on `XLEN`.

`GetFP32(F, rs)` returns the `float` value of register `rs`, automatically NaN-unboxing it on machines with `FLEN > 32`. This means making sure the upper bits are all 1's, and returning `NAN` (quiet NaN `float`) if they are not.

`SetFP32(F, rd, value)` sets register `rd` to the `float value`, automatically NaN-boxing it on machines with `FLEN > 32`.

`AdvancePC(F, bytes)` advances the program counter `bytes` bytes, regardless of whether `XLEN == 32` or `XLEN == 64`. This gets rid of hundreds cases like this:

```
if( F->IsRV32() ){
  R->RV32_PC += Inst.instSize;
} else {
  R->RV64_PC += Inst.instSize;
}
```

## Helper functions
`ZeroExt(T val, size_t bits)` zero-extends `val`, keeping the bottom `bits` bits.

`SignExt(T val, size_t bits)` sign-extends `val`, keeping the bottom `bits` bits and copying bit `bits-1` to the higher positions.

`BoxNaN(double* dest, void* value)` takes a `float` at `value` and `NaN`-boxes it into a `double` stored in `dest`. This API was used to make it easy to call from many places in the code.

**Note:** Portable C++ is used for these functions, by using bitwise operations which are well-defined and work for any integral type `T`. For example, signed values cannot be portably shifted left. `memcpy()` is used to observe strict aliasing, and I verified that the compiler generates the same assembly code whether I use `memcpy()` or illegal pointer type punning.

`ImmSignExt(n)` is a helper member of `RevInst` which returns a sign-extended value formed by the bottom `n` bits of `imm`. It is only guaranteed to return up to `int32_t`, but if it is converted implicitly or explicitly to `int64_t`, it will continue the sign-extension. This is used to extract sign-extended immediate operands of various lengths from the instruction encoding.

`CvtFpToInt()` is a C++ template function which implements 8 instructions: `{ float, double } -> { int32_t, uint32_t, int64_t, uint64_t }`. It uses the out-of-range handling described in Table 11.4 in the RISC-V F specification (no similar table is shown in the D or Q specifications, which seems like a simple oversight).

`fclass()` is a template function which supplements `std::fpclassify()` to make it implement the RISC-V `fclass` instruction's semantics, such as by distinguishing between quiet and signaling NaN.

`load()` and `store()` are templates which implement all possible load/store combinations, based on a single template parameter indicating the type to be loaded/stored, such as `uint8_t`, `int8_t`, `uint16_t`, `int16_t`, `uint32_t`, `int32_t`, and `int64_t`.

`fload()` and `fstore()` are templates which implement all possible FP load/stores, and automatically handle `SPF` vs `DPF` and `NaN`-boxing.

`foper()` is a template for handling FP binary operations (both `float` and `double`).

`FMin` and `FMax` are functor classes to pass to `foper()` when the operation is a FP max or FP min (the `<algorithm>` `std::max` and `std::min` do not handle `NaN` correctly, so `std::fmax` and `std::fmin` are used, but they don't have a C++ standard functor, so I wrote trivial classes to be used as functors.

`fcondop()` is a template for handling all FP conditional instructions where a boolean result is stored in `rd`.

`oper()` is a template for implementing most binary integer operations. It takes an Operator, the Operand Kind (`OpKind::Imm` or `OpKind::Reg`), an optional `std::make_unsigned_t` to make the operation unsigned instead of signed, and an optional `bool` which, if set to `true`, makes the operation `W`-type (32-bit on RV64).

`ShiftLeft` and `ShiftRight` are functors which can be passed to `oper()` to implement all shift instructions.

Throughout the code, there are many instances where long functions are templatized into one, and then references to instantiations of that template are used to replace the long functions. The way to do this is to use a `constexpr` reference to the function, as in:

```
      static constexpr auto& beq  = bcond<std::equal_to>;
      static constexpr auto& bne  = bcond<std::not_equal_to>;
      static constexpr auto& blt  = bcond<std::less,          std::make_signed_t>;
      static constexpr auto& bltu = bcond<std::less,          std::make_unsigned_t>;
      static constexpr auto& bge  = bcond<std::greater_equal, std::make_signed_t>;
      static constexpr auto& bgeu = bcond<std::greater_equal, std::make_unsigned_t>;
```

This implements 6 branch instructions from one template. The `static` makes it a static member not related to a particular class instance; the `constexpr` makes it a constant expression initialized at compile-time (no out-of-class definitions are required in C++17 or later, since it's also considered `inline`); the `auto` makes it unnecessary to declare the full function prototype here; the `&` makes this a reference, so its name is an alias to a function (or object) as if were instantiated with those template arguments.

## Atomic Memory Operations (`AMO`)

The very long, hard-to-read code for executing atomic operations based on flags, has been greatly simplified, using tables and C++ lambdas. What used to require complex `if`-`else` sequences, is now table-driven, and the `XLEN==32` and `XLEN==64` distinctions, are handled by two separate calls to the same template.

9 atomic instructions in both `RV32I` and `RV64I`, have been condensed into a single template.

```
  // Table mapping atomic operations to executable code
  static const std::pair<RevCPU::RevFlag, std::function<void()>> table[] = {
    { RevCPU::RevFlag::F_AMOADD,  [&]{ *TmpTarget += TmpBuf; } },
    { RevCPU::RevFlag::F_AMOXOR,  [&]{ *TmpTarget ^= TmpBuf; } },
    { RevCPU::RevFlag::F_AMOAND,  [&]{ *TmpTarget &= TmpBuf; } },
    { RevCPU::RevFlag::F_AMOOR,   [&]{ *TmpTarget |= TmpBuf; } },
    { RevCPU::RevFlag::F_AMOSWAP, [&]{ *TmpTarget  = TmpBuf; } },
    { RevCPU::RevFlag::F_AMOMIN,  [&]{ *TmpTarget  = std::min(*TmpTarget,  TmpBuf);  } },
    { RevCPU::RevFlag::F_AMOMAX,  [&]{ *TmpTarget  = std::max(*TmpTarget,  TmpBuf);  } },
    { RevCPU::RevFlag::F_AMOMINU, [&]{ *TmpTargetU = std::min(*TmpTargetU, TmpBufU); } },
    { RevCPU::RevFlag::F_AMOMAXU, [&]{ *TmpTargetU = std::max(*TmpTargetU, TmpBufU); } },
  };
```

## Miscellaneous

My initial `JALR` implementation had a race condition where `PC` or the input register could be overwritten before they were used. I wrote a `JALR` testcase which uses both `rd == rs1` and `rd != rs1`.

Rename `RevInstTables.h` (plural) to `AllRevInstTables.h`, so that keyboard auto-completion (`TAB`) can find `RevInstTable.h` (singular) without ambiguity.

Add `.gdb_history` to `.gitignore` since GDB sessions often create it in working tree.

Do not overwrite `RVCC` environment variable with blank CMake `RVCC` variable, allowing `RVCC` to be set either on the CMake command-line or in the environment.

Add C++ standard header `#include`s where they are missing, even if they get indirectly `#include`ed already.

Add spaces after commas in function calls and declarations/definitions. i.e. instead of `F(A, B,C,D, E)`, normalize this to `F(A, B, C, D, E)`.

Casts to `(void *)` are almost never needed, since they are converted to implicitly from other pointers. There are also a lot of other useless casts to the same type removed, and `-Wuseless-cast` was temporarily added to the compiler flags, `-Wold-style-cast` was tested, and all useless casts and most old-style casts eliminated from Rev, but `-Wuseless-cast` is not compatible with Clang, and `-Wold-style-cast` could not seem to be ignored a with `#pragma` during the SST `#include`s, so `-Wuseless-cast` and `-Wold-style-cast` are turned off for now.

There were some types like `tlbSize` and `maxHeapSize` which were entended to 64-bit variables because `memSize` and other were already 64-bit, and I was getting conversion warnings of going back and forth between 32- and 64-bits without explicit casts.

`RevThreadCtx::ChildrenPIDs` and `RevThreadCtx::fildes` were changed from `std::vector<>` to `std::set<>`, since the only thing that was really needed was membership information querying and add/remove, not a contiguous array of values.

In `RevCPU.cc` there were several cases of `new` being checked for `nullptr` even though `std::nothrow` was not used, and hence `new` should never return `nullptr`. I added `TODO:` comments warning about them.

In `RevExt.cc` there were a lot of debug output statements commented out by `#if 0`. I changed them all to `#if TRACE_EXECUTE` so that the definition of `TRACE_EXECUTE` was in one place. (When I set `TRACE_EXECUTE` to `1`, I got compilation errors about undefined names, so I wonder if the code should just be removed.)

Explicitly-sized types like `int32_t`, `int64_t`, `uint32_t` and `uint64_t` are preferred to `int`, `long`, `unsigned long`, etc. For some domains, `size_t` or `ptrdiff_t` are preferred. `size_t` is the type returned by `vector<>::size()`, for example.

In `RevMem.cc` and perhaps others, there were fatal SST error message calls which were not terminated by newlines consistently.

Variable-length-arrays (VLA) were used in a few places, which is a non-standard extension, and which puts the allocation on the stack instead of the heap, increasing the risk of stack overflows. I replaced them with `std::vector<>`.

The tests in `test/isa/fadd.c` had numerical rounding issues which made the tests fail even on `pinhead`. The number of nonzero bits in a `float` needs to be set low enough that the exact result can be obtained.

I removed `using namespace SST::RevCPU;` from several _header files_, since it sort of defeats the purpose of namespaces to use them in header files. Those `using namespace SST::RevCPU;` have been moved to the `.cc` files.

The shorter C++17 syntax `namespace SST::RevCPU { ... }` was used instead of `namespace SST { namespace RevCPU { ... } }`, and the [code inside the namespace was un-indented](https://google.github.io/styleguide/cppguide.html#Namespace_Formatting).


CMakeLists.txt
```diff
@@ -87,7 +85,7 @@ message (LDFLAGS="${LDFLAGS}")
 #------------------------------------------------------------------------
 #-- COMPILER OPTIONS
 #------------------------------------------------------------------------
-set(CMAKE_CXX_FLAGS "-O2 -Wall ${CMAKE_CXX_FLAGS} -std=c++17 -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-O2 -Wall -Wextra -Werror -Wuninitialized -Wmaybe-uninitialized -Wformat=2 -Wformat-overflow=2 -Wformat-signedness -Wformat-truncation=2 -Wuseless-cast -Wdouble-promotion -Wno-error=double-promotion -Wno-variadic-macros -Wno-error=variadic-macros -Wno-deprecated-declarations -Wvolatile -Wno-unused-variable -Wno-unused-parameter -Wno-shift-count-overflow ${CMAKE_CXX_FLAGS} -std=c++17 -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
```
donofrio, line 88: clang on macOS does not like some of these options:
`clang: error: -Wl,-undefined: 'linker' input unused [-Werror,-Wunused-command-line-argument]
clang: error: -Wl,dynamic_lookup: 'linker' input unused [-Werror,-Wunused-command-line-argument]`

leekillough: @donofrio: it should be fixed now.  I changed the options to only include ones which are compatible with both GCC and Clang.

include/PanExec.h
jleidel, line 1: please revert the changes to the PAN files

CMakeLists.txt
donofrio: Looks like clang still does not like `-Werror` I'm running:

Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin22.6.0
Thread model: posix

common/include/PanAddr.h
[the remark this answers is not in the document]
leekillough: > 

An exported C header which must work with both C and C++ clients?

include/PanExec.h
leekillough: I'm curious what broke them?
- C++ `struct name{ ... };` declarations instead of C `typedef struct { ... } name;` In C++ `struct`, `enum` and `class` tags define a new type, while in C the namespaces for `struct` / `enum` tags, and types, are distinct
- Minor implementation changes in the C++ classes (like moving an empty-body constructor or destructor to the header file, to make it inlined)
- `#include "SST.h"`, which replaced `#include <sst/...>` so that `#pragmas` could disable compiler warnings when building external SST code, and that it was all in one place instead of scattered multiple times
- Adding `const` to accessor methods which do not change `*this` (should not break anything unless the exact method signature is repeated elsewhere and must exactly match)
- Removing useless casts like `int x, y; x = (int) y;` or `if((bool) cond)`, which shouldn't affect behavior, or even be visible externally
- Changing an expensive `std::string` to a `const char*`. If the class may outlive the data that the `const char*` points to, then a `std::string` copy may be necessary, but `const char*` and `std::string_view` can still be lightweight ways of returning a string in a query function
- Adding `explicit` to a PAN constructor with one argument where implicit conversion is not expected
- Changing an `enum` to be a C++11 `enum` with a declared underlying type, rather than a C `typedef enum` whose underlying integer type is up to the compiler
- Adding trailing commas at the end of `enum` or initializer lists, so that each line ends the same and can be reordered or appended to, without having to make a special case for dropping the comma from the last line

CMakeLists.txt
leekillough: `-Werror` has been supported by Clang for many years. [Here's where the 14.0.0 docs mention it](https://releases.llvm.org/14.0.0/tools/clang/docs/UsersManual.html#options-to-control-error-and-warning-messages).

The `-W` options I give in CMake are unrelated to `-Wl,`. Your errors seem to be implying that you're specifying `-Wl,` options which are not being used if there is no linking (only compiling) going on. I don't know why Clang is taking `-W` warning options and turning them into `-Wl,` options.

**I would like a `-v` option added to `CMAKE_CXX_FLAGS`, so we can see how the compiler and linker are actually being invoked when you get your error (it makes the driver echo its subcommands).**

I suspect CMake more than Clang is the source of the problem, which is also why setting `CMAKE_CXX_STANDARD` to `17` didn't work for you. 

There's no reason my changing a CMake variable to add more `-W` warnings should cause that error, which seems to be interpreting `-W...` options as `-Wl,...` options instead. I wonder if our CMake code, or another script, has a regexp searching for `-W` options, and can't tell `-Werror` from `-Wl,-s`.

jleidel: This definitely fails on MacOS with Clang 13.0.0
```
[  6%] Building CXX object src/CMakeFiles/revcpu.dir/PanExec.cc.o
clang: error: -Wl,-undefined: 'linker' input unused [-Werror,-Wunused-command-line-argument]
clang: error: -Wl,dynamic_lookup: 'linker' input unused [-Werror,-Wunused-command-line-argument]
make[2]: *** [src/CMakeFiles/revcpu.dir/PanExec.cc.o] Error 1
make[1]: *** [src/CMakeFiles/revcpu.dir/all] Error 2
make: *** [all] Error 2
```

jleidel: This patch fixes the issue w/ a CMake guard: 
```
#------------------------------------------------------------------------
 #-- COMPILER OPTIONS
 #------------------------------------------------------------------------
-set(CMAKE_CXX_FLAGS "-std=c++17 -O2 -Wall -Wextra -Werror -Wvla -Wuninitialized -Wfloat-conversion -Wdouble-promotion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS "-std=c++17 -O2 -Wall -Wextra -Wvla -Wuninitialized -Wfloat-conversion -Wdouble-promotion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+else()
+  set(CMAKE_CXX_FLAGS "-std=c++17 -O2 -Wall -Wextra -Werror -Wvla -Wuninitialized -Wfloat-conversion -Wdouble-promotion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+endif()
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -Wall ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall ${REVCPU_COMPILER_MACROS}")
```

leekillough: No `-Werror` for Clang? It seems strange that that would resolve "linker errors".

I'll merge it and the rest of the `devel` changes into my PR.

jleidel:
we need to freeze these changes such that we can review and merge.

src/RevProc.cc
```diff
@@ -1908,21 +1897,14 @@ void RevProc::DependencyClear(uint16_t HartID, uint16_t RegNum, bool isFloat){
...
-      nextID++;
-      if(nextID >= _REV_HART_COUNT_){
+  if(HART_CTS.any()){
+    while(!HART_CTS[nextID]){
+      if(++nextID >= _REV_HART_COUNT_)
         nextID = 0;
-      }
-      if(HART_CTS[nextID]){ break; };
     }
     output->verbose(CALL_INFO, 6, 0,
```
donofrio, line 1906: This now prints every cycle even when not switching. Recommend reverting back to previous logic

jleidel:
The vector test was removed.  Please revert and put it back in the tree.  There are specific corner cases with respect to memory handling

leekillough:
`cbeqz` had a minor typo of `9` instead of a `6` bits. This caused the `STRLEN_C` failures. Thanks to  @donofrio for figuring this out.

This PR is ready. `vector` tests are back, and the constant printing of Hart-changes, has been fixed.

```
100% tests passed, 0 tests failed out of 94
```

rkabrick (approved):
Given tests passing and our internal review today I am good to merge

donofrio (approved):
Agreed - passing all tests and we had detailed code walk-through. Good to merge.

jleidel (approved):

PR #118, opened 2023-09-14
Unify tests' makefiles & run scripts

vcave:
- Unified environment variables across makefiles and run scripts
- Updated ctest setup to forward environment variables to run scripts
- This patch conserves the pre-existing default operation mode:
  - 'cd' in any test folder and invoke the local run script which relies on default settings in the makefile and local SST config file
  - ctest goes over all the run scripts
- 'test/README.md' shows how developers can now override certain variables such as RVCC, ARCH, ABI, REV_SST_CONFIG from the ctest command line


leekillough:
There are a lot of conflicts with https://github.com/tactcomplabs/rev/pull/115 .

PR #123, opened 2023-09-19
Use smart pointers

leekillough:
This turns the hazard pointers into `std::shared_ptr<bool>`, and uses reference counting to determine when to delete the underlying allocation.

This uses `std::unique_ptr` to hold objects which are allocated on the heap for the lifetime of the class, preventing the need to `delete` them in the destructor. The destruction of the `std::unique_ptr` object will automatically `delete` them.

There is no need to explicitly delete these smart pointers, since removing them as elements from a `std::vector`, `std::list` or other container, will automatically call their destructor, and in the case of `std::shared_ptr`, delete the memory when the reference count goes to zero.

### DIFF
--- a/documentation/ExtensionDevelopment.md
+++ b/documentation/ExtensionDevelopment.md
@@ -7,10 +7,9 @@ There are also several limitations of the current extension functionality.  Thes
 
 1. The Rev crack/decode functions currently only support the standard set of RISC-V instruction formats.  Additional formats can be supported, but will require additional source code modifications to the crack/decode engine.
 2. The Rev model supports the ability for users to override default extensions (for example, the D-extension) with new functionality.  However, you cannot load two extensions with conflicting encodings.  This will break the Rev crack/decode engine.  
-3. The Rev model does not currently support the compressed encodings.
-4. The naming convention for the new extension cannot conflict with the standard set of the extensions if they are utilized in the same core.  Eg, an RV64IMAFD device cannot have a second "F" extension.  You must utilize a different naming convention.
-5. The Rev model utilizes a standard ELF loader.  If the extension breaks the base RISC-V (RV32, RV64) relocation, then the device may not function as expected.
-6. The Rev model supports the standard set of RV32/RV64-G registers for integer and floating point.  Any extension-specific register state will require additional source code modifications.
+3. The naming convention for the new extension cannot conflict with the standard set of the extensions if they are utilized in the same core.  Eg, an RV64IMAFD device cannot have a second "F" extension.  You must utilize a different naming convention.
+4. The Rev model utilizes a standard ELF loader.  If the extension breaks the base RISC-V (RV32, RV64) relocation, then the device may not function as expected.
+5. The Rev model supports the standard set of RV32/RV64-G registers for integer and floating point.  Any extension-specific register state will require additional source code modifications.
 
 ## Source Code Organization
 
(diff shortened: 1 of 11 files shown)


jleidel (approved):

jleidel:
There appears to be a performance regression with this PR.  While I think is a critical set of changes, we should eventually evaluate the impact of certain other code regions.  
Before changes (big_loop.py)
```
real	1m15.933s
```

After changes (same test)
```
1m46.031s
```

leekillough:
This performance regression is not unexpected, but whether smart pointers are the best way to go is not clear. Why, for example, were allocatable bytes used for single Boolean hazard conditions?

Is there a resource (instruction, etc.) to which a hazard bit always belongs, and can it be used to provide the storage for it? Or can a mapping like a `std::set` or `std::unordered_set` be used with other resources (maybe even pointers) as keys, to indicate whether they belong to a hazard or not?

It was not clear to me that @kpgriesser's original change to `delete` the hazard after it was cleared, was necessary, or that it was addressing a memory leak.

I converted everything to `shared_ptr`, passing references to `shared_ptr` rather than copies when possible (such as passing a `shared_ptr` by `const std::shared_ptr<>&` to a function which doesn't actually need to take ownership of it until its value is stored in a data structure which lasts longer than the function itself).

I'd consider if some kind of associative container like sets could be used to query and manipulate hazard information. The question is, which Rev data structure would own this container, such that its lifetime lasts long enough to satisfy any queries or operations to set it, and so that it is automatically destroyed when the owner is destroyed. That can all be done without the expensive `new`/`delete`/`shared_ptr` operations, which happen at the hazard bit level now.

leekillough:
Seems like it's spending 6.5% of its time calling `remove(HB)` on `std::list<std::shared_ptr<bool>>`. It has to search a linked list, remove an entry if it's found, and then destroy it, which involves an atomic decrement of reference count, and a possible `delete`.  The old method just found it and removed it, and, after @kpgriesser's change, deleted it too, but there were still other references to it.

I think some kind of `std::unordered_set<hazard_key>` might be better. Instead of passing smart pointers or pointers to dynamically-allocated `bool` around, you pass keys, perhaps pointers to other objects themselves, which are then used to look up membership in `std::unordered_set` or add/remove them from it.  The question is where should this set live, so that it has right lifetime, and so that can be made, e.g., Hart-private (perhaps Hart should just be a part of the key).

jleidel:
> Seems like it's spending 6.5% of its time calling `remove(HB)` on `std::list<std::shared_ptr<bool>>`. It has to search a linked list, remove an entry if it's found, and then destroy it, which involves an atomic decrement of reference count, and a possible `delete`. The old method just found it and removed it, and, after @kpgriesser's change, deleted it too, but there were still other references to it.
> 
> I think some kind of `std::unordered_set<hazard_key>` might be better. Instead of passing smart pointers or pointers to dynamically-allocated `bool` around, you pass keys, perhaps pointers to other objects themselves, which are then used to look up membership in `std::unordered_set` or add/remove them from it. The question is where should this set live, so that it has right lifetime, and so that can be made, e.g., Hart-private (perhaps Hart should just be a part of the key).

I like your suggestion of an unordered map with a hazard key.  That might provide better scalability as we move forward to larger HART counts.  I'm running a test through valgrind w/ callgrind/cachegrind analysis.  I'll post the results once it finishes

leekillough:
Did you use Release build?  My times for `big_loop` are much shorter with Release:

Old:
```
Simulation is complete, simulated time: 10.2298 ms

real    0m5.442s
user    0m4.621s
sys     0m0.719s
```
New:
```
Simulation is complete, simulated time: 10.2298 ms

real    0m5.735s
user    0m5.024s
sys     0m0.611s
```
So the new code is only about 5% slower than the old code.

I have removed:
```c++
  std::list<std::shared_ptr<bool>> LoadHazards;        ///< RevProc: list of allocated load hazards

  /// RevProc: creates a new pipeline load hazard and returns a pointer to it
  std::shared_ptr<bool> createLoadHazard();

  /// RevProc: destroys the target load hazard object and removes it from the list
  void destroyLoadHazard(const std::shared_ptr<bool>& hazard);
```
As being unnecessary, now that smart pointers are used. Apparently `LoadHazards` was a management class to take care of allocating and freeing the hazards, even though they were used elsewhere.

Now that they are smart pointers, we don't need to manage them with `LoadHazards`, and the large time spent in `remove`ing and `deleting` `LH` is no longer necessary, since `std::shared_ptr<bool>` will automatically handle the deletions.

VTune hotspots:
```
Top Hotspots
Function                                                                                                                                                                                                                             Module        CPU Time  % of CPU Time(%)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------  ------------  --------  ----------------
__memset_avx2_unaligned_erms_rtm                                                                                                                                                                                                     libc.so.6       0.580s              9.0%
std::shared_ptr<SST::RevCPU::RevThreadCtx>::~shared_ptr                                                                                                                                                                              librevcpu.so    0.478s              7.4%
__gnu_cxx::__exchange_and_add_dispatch                                                                                                                                                                                               sstsim.x        0.364s              5.6%
std::unordered_map<unsigned int, std::shared_ptr<SST::RevCPU::RevThreadCtx>, std::hash<unsigned int>, std::equal_to<unsigned int>, std::allocator<std::pair<unsigned int const, std::shared_ptr<SST::RevCPU::RevThreadCtx>>>>::find  librevcpu.so    0.344s              5.3%
SST::RevCPU::RevProc::DecodeInst                                                                                                                                                                                                     librevcpu.so    0.248s              3.8%
[Others]                                                                                                                                                                                                                             N/A             4.446s             68.8%
```

jleidel:
> Did you use Release build? My times for `big_loop` are much shorter with Release:
> 
> Old:
> 
> ```
> Simulation is complete, simulated time: 10.2298 ms
> 
> real    0m5.442s
> user    0m4.621s
> sys     0m0.719s
> ```
> 
> New:
> 
> ```
> Simulation is complete, simulated time: 10.2298 ms
> 
> real    0m5.735s
> user    0m5.024s
> sys     0m0.611s
> ```
> 
> So the new code is only about 5% slower than the old code.
> 
> I have removed:
> 
> ```c++
>   std::list<std::shared_ptr<bool>> LoadHazards;        ///< RevProc: list of allocated load hazards
> 
>   /// RevProc: creates a new pipeline load hazard and returns a pointer to it
>   std::shared_ptr<bool> createLoadHazard();
> 
>   /// RevProc: destroys the target load hazard object and removes it from the list
>   void destroyLoadHazard(const std::shared_ptr<bool>& hazard);
> ```
> 
> As being unnecessary, now that smart pointers are used. Apparently `LoadHazards` was a management class to take care of allocating and freeing the hazards, even though they were used elsewhere.
> 
> Now that they are smart pointers, we don't need to manage them with `LoadHazards`, and the large time spent in `remove`ing and `deleting` `LH` is no longer necessary, since `std::shared_ptr<bool>` will automatically handle the deletions.
> 
> VTune hotspots:
> 
> ```
> Top Hotspots
> Function                                                                                                                                                                                                                             Module        CPU Time  % of CPU Time(%)
> -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------  ------------  --------  ----------------
> __memset_avx2_unaligned_erms_rtm                                                                                                                                                                                                     libc.so.6       0.580s              9.0%
> std::shared_ptr<SST::RevCPU::RevThreadCtx>::~shared_ptr                                                                                                                                                                              librevcpu.so    0.478s              7.4%
> __gnu_cxx::__exchange_and_add_dispatch                                                                                                                                                                                               sstsim.x        0.364s              5.6%
> std::unordered_map<unsigned int, std::shared_ptr<SST::RevCPU::RevThreadCtx>, std::hash<unsigned int>, std::equal_to<unsigned int>, std::allocator<std::pair<unsigned int const, std::shared_ptr<SST::RevCPU::RevThreadCtx>>>>::find  librevcpu.so    0.344s              5.3%
> SST::RevCPU::RevProc::DecodeInst                                                                                                                                                                                                     librevcpu.so    0.248s              3.8%
> [Others]                                                                                                                                                                                                                             N/A             4.446s             68.8%
> ```
> 
> Looks like `std::shared_ptr<SST::RevCPU::RevThreadCtx>` is where the time is being spent now w.r.t. smart pointers. The Hazards don't show up in the top hotspots.

I was comparing a `Debug` build of the upstream devel against a `Debug` build of your code.  The local system may be incurring some degree of congestion, which would explain higher that normal performance deltas

I'm good to merge this!

PR #126, opened 2023-09-20
Improved hazarding logic + Hart terminology cleanup

donofrio:
Adding a Load / Store queue to Rev (although, truthfully we only track loads, but that is another matter) to replace the bool* Hazard logic.  There now is a single LSQueue per RevProc implemented as an unordered_map. A shared_ptr to LSQueue is provided to each RegFile. 

Pipeline stalls are now handled through the existing DependencyCheck() method - the LS Queue is checked if any instruction source registers are currently the target of an in-flight read. Pipeline execution will continue until a dependent register is found.

Replacing the bool* that was once passed with every call to ReadVal() there is now a new POD class, MemReq. This tracks the destination register, address to be read, as well as the Hart the request originated from. 

To use this new bookkeeping: *Prior* to calling ReadVal(), create a MemReq and add to the LSQueue. The hash is computed through a lambda make_lsq_hash().  Note that an unordered map will overwrite an existing hash table entry if you give it two identical keys.  Currently we do not check on insert if a value was overwritten, but can add that if we want more defensive coding. It is a matter of checking the return value.

Clearing the outstanding load is accomplished by the RevProc::MarkLoadAsComplete() function. All MemReq objects have a pointer to this member function, so as the memory system completes the writeback to the RegFile it calls this function directly to mark the Load as complete and clear the Dependency Scoreboard.  If you attempt to clear a load that is not in the LS Queue an error is thrown.

Finally, the RevFeature::GetHart was not being used consistently.  There is now a GetProcID and a GetHartToExec - these were previously collapsed.

### DIFF
--- a/src/RevMemCtrl.cc
+++ b/src/RevMemCtrl.cc
@@ -24,7 +24,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
                    RevMemOp::MemOp Op, StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size), Inv(false),
     Op(Op), CustomOpc(0),
-    SplitRqst(1), flags(flags), target(nullptr), hazard(nullptr){
+    SplitRqst(1), flags(flags), target(nullptr), procReq(){
 }
 
 RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr,
@@ -32,7 +32,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                    RevMemOp::MemOp Op, StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size), Inv(false),
     Op(Op), CustomOpc(0),
-    SplitRqst(1), flags(flags), target(target), hazard(nullptr){
+    SplitRqst(1), flags(flags), target(target), procReq(){
 }
 
 RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
@@ -40,7 +40,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
                    StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size),
     Inv(false), Op(Op), CustomOpc(0),
-    SplitRqst(1), flags(flags), target(nullptr), hazard(nullptr){
+    SplitRqst(1), flags(flags), target(nullptr), procReq(){
   for(uint32_t i = 0; i < Size; i++){
     membuf.push_back(buffer[i]);
   }
@@ -51,7 +51,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
                    StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size),
     Inv(false), Op(Op), CustomOpc(0),
-    SplitRqst(1), flags(flags), target(target), hazard(nullptr){
+    SplitRqst(1), flags(flags), target(target), procReq(){
   for(uint32_t i = 0; i < Size; i++){
     membuf.push_back(buffer[i]);
   }
@@ -62,7 +62,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
                    StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size),
     Inv(false), Op(Op), CustomOpc(0),
-    SplitRqst(1), membuf(buffer), flags(flags), target(nullptr), hazard(nullptr){
+    SplitRqst(1), membuf(buffer), flags(flags), target(nullptr), procReq(){
 }
 
 RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
@@ -70,7 +70,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr, uint32_t Size,
                    StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size), Inv(false), Op(Op),
     CustomOpc(CustomOpc), SplitRqst(1), flags(flags),
-    target(target), hazard(nullptr){
+    target(target), procReq(){
 }
 
 RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr,
@@ -78,7 +78,7 @@ RevMemOp::RevMemOp(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                    unsigned CustomOpc, RevMemOp::MemOp Op,
                    StandardMem::Request::flags_t flags )
   : Hart(Hart), Addr(Addr), PAddr(PAddr), Size(Size), Inv(false), Op(Op),
-    CustomOpc(CustomOpc), SplitRqst(1), flags(flags), target(nullptr), hazard(nullptr){
+    CustomOpc(CustomOpc), SplitRqst(1), flags(flags), target(nullptr), procReq(){
   for(uint32_t i = 0; i< Size; i++){
     membuf.push_back((uint8_t)(buffer[i]));
   }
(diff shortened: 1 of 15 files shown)


src/RevMemCtrl.cc
```diff
@@ -1296,10 +1292,8 @@ void RevBasicMemCtrl::handleReadResp(StandardMem::ReadResp* ev){
         if( isAMO ){
           handleAMO(op);
         }
-        auto Hazard = op->getHazard();
-        if( Hazard != nullptr ){
-          *Hazard = false;
-        }
+        MemReq r = op->getMemReq();
+        r.MarkLoadComplete(r);
```
jleidel, line 1296: There may be a bug here, but I can't quite tell.  The idea behind this logic is to clear a load hazard in the case where the load spans a cache line (eg, the "split" request).  If we see an AMO occurring, should we clear the hazard?

donofrio: That is a bug - good catch. The handleAMO function will call performAMO() which will clear the hazard. Clearing again here will cause an error. I can add a check similar to line 1316 that only clears if the request is NOT an amo.

leekillough:
`common/include/RevCommon.h` is missing a newline at the end of file.

`RevInstTable.h`:
```
/*static_assert(std::is_trivially_copyable_v<RevRegFile>,
              "RevRegFile must be trivially copyable to be able to use memcpy() and memset()");*/
```
In the past we used `memcpy()` to copy and `memset()` to zero-out registers. I need to make sure we're not still doing this. The `static_assert` was to prevent surprising bugs caused by using `memcpy()` or `memset()` on classes which are not trivially copyable. The comment can be totally removed if we no longer do that, but rather than putting repetitious code to do copies and/or clearing of registers element by element, we should provide such methods in the `RevRegFile` class itself.

In `RevProc.cc`:
```
     // Zero all register data
    memset(regFile->RV32, 0, sizeof(*(regFile->RV32)));
    memset(regFile->RV64, 0, sizeof(*(regFile->RV64)));
    memset(regFile->SPF, 0, sizeof(*(regFile->SPF)));
    memset(regFile->DPF, 0, sizeof(*(regFile->DPF)));
    memset(regFile->RV32_Scoreboard, 0, sizeof(*(regFile->RV32_Scoreboard)));
    memset(regFile->RV64_Scoreboard, 0, sizeof(*(regFile->RV64_Scoreboard)));
    memset(regFile->SPF_Scoreboard, 0, sizeof(*(regFile->SPF_Scoreboard)));
    memset(regFile->DPF_Scoreboard, 0, sizeof(*(regFile->DPF_Scoreboard)));
```
That should be moved into the `RevRegFile` class/struct, possibly as a default constructor `RevRegFile()` (which can be declared `= default` or omitted entirely if "member initialization" syntax is used, as shown below). Rather than using `memset`, you can use `{}` member "value" initialization notation in the `RevRegFile` class to make it simpler and faster:
```
  uint32_t RV32[_REV_NUM_REGS_]{};    ///< RevRegFile: RV32I register file
  uint64_t RV64[_REV_NUM_REGS_]{};    ///< RevRegFile: RV64I register file
  float SPF[_REV_NUM_REGS_]{};        ///< RevRegFile: RVxxF register file
  double DPF[_REV_NUM_REGS_]{};       ///< RevRegFile: RVxxD register file

  /* Supervisor Mode CSRs */
  uint64_t RV64_SSTATUS{}; // During ecall, previous priviledge mode is saved in this register (Incomplete)
  uint64_t RV64_SEPC{};    // Holds address of instruction that caused the exception (ie. ECALL)
  uint64_t RV64_SCAUSE{};  // Used to store cause of exception (ie. ECALL_USER_EXCEPTION)
  uint64_t RV64_STVAL{};   // Used to store additional info about exception (ECALL does not use this and sets value to 0)
  uint64_t RV64_STVEC{};   // Holds the base address of the exception handling routine (trap handler) that the processor jumps to when and exception occurs

  uint32_t RV32_SSTATUS{};
  uint32_t RV32_SEPC{};
  uint32_t RV32_SCAUSE{};
  uint32_t RV32_STVAL{};
  uint32_t RV32_STVEC{};

  bool RV32_Scoreboard[_REV_NUM_REGS_]{}; ///< RevRegFile: Scoreboard for RV32I RF to manage pipeline hazard
  bool RV64_Scoreboard[_REV_NUM_REGS_]{}; ///< RevRegFile: Scoreboard for RV64I RF to manage pipeline hazard
  bool SPF_Scoreboard[_REV_NUM_REGS_]{};  ///< RevRegFile: Scoreboard for SPF RF to manage pipeline hazard
  bool DPF_Scoreboard[_REV_NUM_REGS_]{};  ///< RevRegFile: Scoreboard for DPF RF to manage pipeline hazard

  std::shared_ptr<std::unordered_map<uint64_t, MemReq>> LSQueue;
  std::function<void(MemReq)> MarkLoadComplete;

  uint32_t RV32_PC{};                 ///< RevRegFile: RV32 PC
  uint64_t RV64_PC{};                 ///< RevRegFile: RV64 PC
  uint64_t FCSR{};                    ///< RevRegFile: FCSR

  uint32_t cost{};                    ///< RevRegFile: Cost of the instruction
  bool trigger{};                     ///< RevRegFile: Has the instruction been triggered?
  unsigned Entry{};                   ///< RevRegFile: Instruction entry

```
Basically, putting `{}` at the end of a variable declaration, with no `=` sign, "value-initializes" (zeros) it, even if it's an array.

(`= {}` syntax can be used, but then it requires an accessible copy constructor, even if C++17 elides the copy, and it gets into the complicated rabbit hole of [aggregate initialization](https://open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1008r1.pdf). Using `{}` directly without `=` does not require an accessible copy constructor, so it's better practice.)

If you need to _assign_ a zero'ed value to an already-existing `RevRegFile`, you can do it like:
```
reg = RevRegFile{};
```

There are several instances of:
```
MemReq req = MemReq();
```
Although C++17's mandatory copy elision doesn't change the meaning between the above and below, it is usually written more succinctly as (with braces, because parentheses make it look like a function declaration):
```
MemReq req{};
```
In `RevCommon.h`:
```
     MemReq():
        Addr(_INVALID_ADDR_), DestReg(0), RegType(RegUNKNOWN), Hart(_REV_INVALID_HART_ID_),
        ReqType(MemOpCUSTOM), isOutstanding(false), MarkLoadComplete(nullptr)
      {};
```
You don't need `MemReq req{}` or `MemReq req = MemReq()` right now, with the way the class's default constructor is user-supplied above; you can write simply `MemReq req;`. But if you feel better with an explicit initialization, you should use `MemReq req{};`.
```
 std::function<void(MemReq)> MarkLoadComplete;
```
If there are any captures in `MarkLoadComplete` which require dynamic allocation, then copying `MemReq` will be expensive. Ways to mitigate this: 1. Wrap `MarkLoadComplete` into a `std::shared_ptr<...>` so that only the `shared_ptr` is copied but is freed automatically when the last used is finished; 2. Pass `MemReq` by reference instead of by value, making copies or assignments only when necessary.

`RevMem.h`:
```
   bool ReadMem( unsigned Hart, uint64_t Addr, size_t Len, void *Target,
                MemReq req,
                 StandardMem::Request::flags_t flags);
```
```
   template <typename T>
   bool ReadVal( unsigned Hart, uint64_t Addr, T *Target,
                MemReq req,
                 StandardMem::Request::flags_t flags){
    return ReadMem(Hart, Addr, sizeof(T), Target, Hazard, flags);
    return ReadMem(Hart, Addr, sizeof(T), Target, req, flags);
   }
```
```
   template <typename T>
   bool AMOVal( unsigned Hart, uint64_t Addr, T *Data, T *Target,
               MemReq req,
                StandardMem::Request::flags_t flags){
    return AMOMem(Hart, Addr, sizeof(T), Data, Target, req, flags);
   }
```
```
   bool LRBase(unsigned Hart, uint64_t Addr, size_t Len,
              void *Data, uint8_t aq, uint8_t rl, MemReq req,
               StandardMem::Request::flags_t flags);
```
```
   bool AMOMem(unsigned Hart, uint64_t Addr, size_t Len,
              void *Data, void *Target, MemReq req,
               StandardMem::Request::flags_t flags);
```
`RevMemCtrl.h`:
```
   virtual bool sendREADRequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                               uint32_t Size, void *target, MemReq req,
                                StandardMem::Request::flags_t flags) = 0;
```
```
   virtual bool sendAMORequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                               uint32_t Size, char *buffer, void *target,
                              MemReq req,
                               StandardMem::Request::flags_t flags) = 0;
```
```
   virtual bool sendREADLOCKRequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                                   uint32_t Size, void *target, MemReq req,
                                    StandardMem::Request::flags_t flags) = 0;
```
```
   virtual bool sendREADRequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                               uint32_t Size, void *target, MemReq req,
                                StandardMem::Request::flags_t flags) override;
```
```
   virtual bool sendAMORequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                               uint32_t Size, char *buffer, void *target,
                              MemReq req,
                               StandardMem::Request::flags_t flags) override;
```
```
   // RevBasicMemCtrl: send a readlock request
   virtual bool sendREADLOCKRequest(unsigned Hart, uint64_t Addr, uint64_t PAddr,
                                   uint32_t Size, void *target, MemReq req,
                                    StandardMem::Request::flags_t flags) override;
```

See above. All of those `MemReq` pass-by-value arguments will require copy time. If  you pass by `MemReq& req` (`const MemReq& req`, is better, if possible), it will be faster.

`RevMemCtrl.h`:
```
 MemReq getMemReq() const { return procReq; }
```
should be:
```
 const MemReq& getMemReq() const { return procReq; }
```
or:
```
 MemReq getMemReq() const { return procReq; }
```
to prevent expensive copies of requests. Then in `RevMemCtrl.cc`:

```
        MemReq r = op->getMemReq();
    MemReq r = op->getMemReq();
  MemReq r = Tmp->getMemReq();
        MemReq r = op->getMemReq();
    MemReq r = op->getMemReq();
```
should similarly be:
```
        const MemReq& r = op->getMemReq();
    const MemReq& r = op->getMemReq();
  const MemReq& r = Tmp->getMemReq();
        const MemReq& r = op->getMemReq();
    const MemReq& r = op->getMemReq();
```
without the `const` if it can't work -- but in `RevProc.cc` you can probably use:
```
void RevProc::MarkLoadComplete(const MemReq& req){
```
In general, copying `std::function` is an expensive operation, so references should be used if possible, or it should be wrapped in `std::shared_ptr` if not.

PR #128, opened 2023-09-21
Fix the RevReq PR changes

leekillough:
Too long to discuss right now ( I have an appointment in 30 minutes) -- there were several bugs, such as enums defined in multiple places, and a multiple-defined symbol.  See this:

https://github.com/tactcomplabs/rev/pull/126#issuecomment-1729975404


jleidel (commented):
testing now

jleidel:
build error on `pennywise`

```
/home/jleidel/dev/working/SST-DEV/GHRev-PRFIX/src/RevProc.cc:1836:21: error: comparison of promoted bitwise complement of an unsigned value with unsigned [-Werror=sign-compare]
 1836 |   if( ( (HartToExec != _REV_INVALID_HART_ID_) && !RegFile->trigger) && !Halted && HART_CTE[HartToExec]){
/home/jleidel/dev/working/SST-DEV/GHRev-PRFIX/src/RevProc.cc:2082:20: error: comparison of promoted bitwise complement of an unsigned value with unsigned [-Werror=sign-compare]
 2082 |     if( HartToExec != _REV_INVALID_HART_ID_ ){
```

compiler = gcc-11 (Ubuntu 11.1.0-1ubuntu1~18.04.1) 11.1.0

leekillough:
That particular error is a compiler bug:

https://gcc.gnu.org/legacy-ml/gcc-bugs/2008-12/msg00018.html
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=38341
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107465
https://stackoverflow.com/questions/18447210/understanding-the-warning-comparison-of-promoted-unsigned-with-unsigned

However, during the `~` operation, `uint16_t` _does_ get promoted to `int`, so comparing a `uint16_t` against `(~uint16_t{0})` does produce a different-signed conversion, which is another warning we've enabled. It seems from the discussion list above that GCC optimizes away the promotion, so that's why in my build, I did not see the signed-compared-with-unsigned warning/error.

I've tweaked the cast to fix this.

(I have no idea why the C/C++ committees chose to make the promotion of `unsigned char` or `unsigned short` to `signed int` the default if a `signed int` can represent all values of the original type, and `unsigned int` otherwise -- they should only promote signed -> signed and unsigned -> unsigned. The default promotion to `int` turns an unsigned value into a signed value.)

jleidel:
tested good... merging

PR #129, opened 2023-09-22
Threading Rework (Pthread-like functionality)

rkabrick:
# Overview 
## User Functionality Added
`int rev_pthread_create( rev_pthread_t* thread, void* fn, void* arg )`
- Arg support will be added soon (right now it doesn't matter what gets put there)
- Eventually, we will add a `rev_pthread_attr` object that will be passed this function

`int rev_pthread_join( rev_pthread_t thread )`
- The calling thread waits for the passed-in thread to complete its execution

*NOTE*: All thread 'scheduling' is a simple FIFO. This is easily extended and will likely be extended in the near future 

## RevCPU 
**Threads Table**:
```cpp
// Global table of threads
// - The actual RevThread objects live in this table 
// - Only place where the presence of a RevThread on this CPU is guaranteed throughout execution
std::map<uint32_t, std::shared_ptr<RevThread>> Threads;
```

**AssignedThreads Vector of Vectors**:
```cpp
  // 2D vector to hold threads assigned to each processor and its corresponding harts.
  // AssignedThreads[i][j] holds the RevThread executing on processor 'i' and hart 'j'.
  // Oversubscription is not supported in hardware 
  // (ie. AssignedThreads.at(i).size() will never exceed the number of HARTS)
  std::vector<std::vector<std::shared_ptr<RevThread>>> AssignedThreads;
```

**UpdateThreadAssignments(uint32_t ProcID)**:
```cpp
// Checks core 'ProcID' to see if it has any available harts to assign work to
// if it does and there is work to assign (ie. ThreadQueue is not empty)
// assign it and enable the processor if not already enabled.
void UpdateThreadAssignments(uint32_t ProcID);
```

**ThreadQueue**: Thread IDs to be scheduled (ie. assigned to a Proc)
**BlockedThreads**: Set of Thread IDs that are currently blocked (ie. Waiting to join another thread)
**CompletedThreads**: Set of Thread IDs that have completed execution

**ThreadID Assignment**:
- `__INVALID_TID__ = 0`: No thread should ever have this ID 
- The main thread for each RevCPU will be equal to the ComponentID + 1 (+1 is necessary to prevent conflicts when ComponentID is 0)
- Every thread after that is generated via an `SST::RNG::MarsagliaRNG*` inside of RevCPU

```cpp
  // Random Number Generator for generating new Thread IDs.
  SST::RNG::MarsagliaRNG* RNG;

  // Generates a new Thread ID using the RNG.
  uint32_t GetNewThreadID() { return RNG->generateNextUInt32(); }
```

### Thread Management in RevCPU::clockTick:
- Each `RevProc` has a corresponding bitset to signal to the RevCPU that one of their `AssignedThreads` has experienced a change of state
  ```cpp
  std::bitset<_REV_HART_COUNT_> ThreadStateChanges; ///< RevProc: used to signal to RevCPU that the thread assigned to HART has changed state
  ```

- Each `clockTick` in `RevCPU` checks if a Proc's `ThreadStateChanges.any()` have been set

### Overview of New RevCPU::clockTick 
    1) `UpdateThreadAssignments`:
        1a. Check a Proc's utilization (Number of threads assigned to this Proc / Number of HARTs this Proc has)
        1b. If a proc's utilization is < 100 we have room to place a thread (if we have one to schedule in ThreadQueue)
        1c. Call `AssignThread(ThreadQueue.pop())`
        1d. If utilization is 0 and we have no work, disable this Proc
    2) Execute `RevProc::ClockTick`
        2a. If RevCPU assigned a thread to this proc at index 2 of this Proc's AssignedThreads reference vector, Hart 2 will execute it (in barrel fashion)
        2b. If during this thread's execution, it experiences a state transition (ie. call to join [transitions to BLOCKED], completes execution [transitions to DONE], etc)
            set the corresponding bit in ThreadStateChanges which signals to RevCPU that it needs to handle the AssignedThread.at(ThisProcID).at(ThreadIndexThatChangedState);
    3) Back to RevCPU: If Procs[i]->GetThreadStateChanges().any():
        3a. Figure out what the new state is for the thread that changed state (Let's assume the thread has transitioned to BLOCKED)
        3b. Figure out what thread this one is waiting to join (`WaitingOnTID = Thread->GetWaitingToJoinTID()`)
        3c. Check if that thread has already completed (`CompletedThreads.find(WaitingOnTID)`)
        3d. If it has completed, change this thread's state back to `RUNNING` and it will resume execution where it was 
        3e. If it hasn't completed, remove this thread from this Proc's AssignedThreads vector & add it's Thread ID to `BlockedThreads`
            NOTE: if a thread completes, it does the same thing but moves the ThreadID to `CompletedThreads`
        3f. Zero the ThreadStateChanges.at(i) as the CPU has handled the state change
    4) If all Procs are disabled: 
        - We check to make sure that there are no remaining threads to schedule (ie. `ThreadQueue.empty()` )
        - There are no BlockedThreads (`BlockedThreads.size() == 0` )
        - There has been at least 1 thread that has completed (`CompletedThreads.size()`)
        - RevCPU will call the new `PrintStatSummary` function which prints the same statistics that the `RevProc::ClockTick` function used to, indicating the end of execution

## RevProc
**AssignedThreads**:
- Vector which is a reference to the AssignedThreads vector that makes up the 2nd dimension of the RevCPU's `AssignedThreads` vector
  which ensures all changes to this vector maintain parity (`RevCPU::AssignedThreads.at(i)` is the same as `RevProc::AssignedThreads`)
- If a Proc's `AssignedThreads` vector ever has more entries than it does HARTs, it will call `output->fatal` complaining that it has become oversubscribed
- All previous `*Ctx*`, `ThreadTable`, etc functions have been removed 
- `clone` functionality was removed for now 
- All `pid` related functions & objects removed. Rev supports multithreading, not multiple processes
- `GetNewThreadID` function pointer to `RevCPU`'s RNG to generate a new thread ID whenever a thread on this Proc encounters a `pthread_create` function
- `Reset` function was gutted because of the above changes (ie. Don't zero register file anymore because it doesn't live in RevProc anymore)

## RevMem:
**ThreadMemSegs**:
- Vector of `MemSegments`that represent the stack memory & TLS memory (TLS not implemented yet) for a thread (Each `RevThread` has a pointer to its corresponding stack & memory)
 
## Moved Functionality
- `SetupArgs` function now in `RevCPU` because it only has to run for MainThread
- System Call Implementations are now in their own file (`RevSysCalls.cc`)
- `RevProc` no longer prints its stats in ClockTick... instead, it has a `PrintStatSummary` function that is called by RevCPU at the end of execution
- ThreadID generation now happens in RevCPU instead of RevMem
- `RevThreadCtx` -> `RevThread`

--- 

## New Printing Method for RevThread
- Added a `Thread.print()` method that you can call using SSTOutput like so: 
`output->verbose(CALL_INFO, int, int, "Thread Info: %s", MyThread->print())`

```
|============================================================|
| Thread 1                                                   |
|------------------------------------------------------------|
 ==> State: START
 ==> ParentTID: 0
 ==> Blocked by TID: N/A
|------------------------------------------------------------|
|  Reg | Alias |               Value |                  Info |
|------|-------|---------------------|-----------------------|
|   x0 |  zero |                 0x0 |         Zero Register |
|   x1 |    ra |                 0x0 |        Return Address |
|   x2 |    sp |        0x1073740800 |         Stack Pointer |
|   x3 |    gp |             0x72888 |        Global Pointer |
|   x4 |    tp |                 0x0 |        Thread Pointer |
|   x5 |    t0 |                 0x0 |    Temporary Register |
|   x6 |    t1 |                 0x0 |    Temporary Register |
|   x7 |    t2 |                 0x0 |    Temporary Register |
|   x8 |    s0 |             0x72888 | Callee Saved Register |
|   x9 |    s1 |                 0x0 | Callee Saved Register |
|  x10 |    a0 |                 0x0 |   Arg/Return Register |
|  x11 |    a1 |                 0x0 |   Arg/Return Register |
|  x12 |    a2 |                 0x0 |     Argument Register |
|  x13 |    a3 |                 0x0 |     Argument Register |
|  x14 |    a4 |                 0x0 |     Argument Register |
|  x15 |    a5 |                 0x0 |     Argument Register |
|  x16 |    a6 |                 0x0 |     Argument Register |
|  x17 |    a7 |                 0x0 |     Argument Register |
|  x18 |    s2 |                 0x0 | Callee Saved Register |
|  x19 |    s3 |                 0x0 | Callee Saved Register |
|  x20 |    s4 |                 0x0 | Callee Saved Register |
|  x21 |    s5 |                 0x0 | Callee Saved Register |
|  x22 |    s6 |                 0x0 | Callee Saved Register |
|  x23 |    s7 |                 0x0 | Callee Saved Register |
|  x24 |    s8 |                 0x0 | Callee Saved Register |
|  x25 |    s9 |                 0x0 | Callee Saved Register |
|  x26 |   s10 |                 0x0 | Callee Saved Register |
|  x27 |   s11 |                 0x0 | Callee Saved Register |
|  x28 |    t3 |                 0x0 |    Temporary Register |
|  x29 |    t4 |                 0x0 |    Temporary Register |
|  x30 |    t5 |                 0x0 |    Temporary Register |
|  x31 |    t6 |                 0x0 |    Temporary Register |
|------------------------------------------------------------|
```


include/RevProc.h
```diff
@@ -122,61 +124,38 @@ class RevProc{
...
+
+  ///< RevProc: Holds the info for all new threads to be spawned
+  ///           Right now this is through the following functions:
+  ///           - rev_pthread_create
+  std::queue<std::shared_ptr<RevThread>> NewThreadInfo;
 
-  bool UpdateRegFile();
-  // bool UpdateRegFile(uint16_t HartID);
+  ///< RevProc: Used for scheduling in RevCPU (if Utilization < 1, there is at least 1 unoccupied HART )
+  double GetUtilization(){ return ((double)AssignedThreads.size() / _REV_HART_COUNT_) * 100; }
```
leekillough, line 150: More concise:
```c++
return AssignedThreads.size() * 100.0 / _REV_HART_COUNT_;
```
By putting `* 100.0` in the middle, you force all operands to be converted to `double` without an explicit cast.

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+using namespace SST::RevCPU;
+using MemSegment = RevMem::MemSegment;
+
+// Enum for tracking the state of a RevThread.
+enum class ThreadState {
+    START,   // Initial state; resources are allocated.
+    RUNNING, // Thread is assigned and executing on a Proc/HART.
+    BLOCKED, // Waiting for I/O or thread synchronization.
+    READY,   // In the ready list; waiting for CPU (not implemented yet).
+    DONE     // Thread has finished; deallocate resources.
```
leekillough, line 35: Please use trailing commas on all items, so that adding, removing, or permuting entries does not require adjusting commas and lining up spacing.

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+
+class RevThread {
+public:
+    RevThread(uint32_t inputThreadID, uint32_t inputParentThreadID,
+              uint64_t inputStackPtr, uint64_t inputFirstPC,
+              std::shared_ptr<MemSegment>& inputThreadMem,
+              RevFeature* inputFeature);
+
+    // ThreadID operations
+    uint32_t GetThreadID(){ return ThreadID; }
```
leekillough, line 46: Getter functions should be `const` if possible (if not, consider a `mutable` member).

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+
+    // State checks
+    bool isRunning();
+    bool isBlocked();
+    bool isDone();
+
+    // Printing functions
+    std::string GetStateString();
+    friend std::ostream& operator<<(std::ostream& os, RevThread& Thread);
+
```
leekillough, line 83: `Thread` should be `const RevThread&`.

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+    const char* print() {
+      std::ostringstream oss;
+      oss << *this;  // Assuming operator<< is already overloaded for RevThread
+      SSTOutputBuffer = oss.str();  // Store the string
+      return SSTOutputBuffer.c_str();  // Return C-string
+    }
```
leekillough, line 89: `to_string()` is the C++ idiom.  Callee should decide whether to store the returned `std::string` in a variable for use of `c_str()` or other methods, rather than filling up the class with `SSTOutputBuffer`.

Also should be `const` (once `operator<<` is fixed to use `const`).
```
std::string to_string() const {
      std::ostringstream oss;
      oss << *this;  // Assuming operator<< is already overloaded for RevThread
      return oss.str();  // Return the string
    }
```

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+    RevFeature* Feature;                    // Feature set for this thread
+    uint64_t ThreadPtr;                     // Thread pointer
+    ThreadState State = ThreadState::START; // Thread state (Initializes as )
+    RevRegFile RegFile;  // Register file
+    std::vector<uint32_t> ChildrenThreadIDs = {};
+    std::vector<int> fildes = {0, 1, 2};  // Default file descriptors
+
+    uint32_t WaitingToJoinTID = __INVALID_TID__;
+    
+   std::string SSTOutputBuffer;  // Buffer to hold the printed string
```
leekillough, line 106: See above.

include/RevCPU.h
```diff
@@ -212,6 +221,63 @@ class RevCPU : public SST::Component{
...
+  // 2D vector to hold threads assigned to each processor and its corresponding harts.
+  // AssignedThreads[i][j] holds the RevThread executing on processor 'i' and hart 'j'.
+  // Oversubscription is not supported in hardware 
+  // (ie. AssignedThreads.at(i).size() will never exceed the number of HARTS)
+  std::vector<std::vector<std::shared_ptr<RevThread>>> AssignedThreads;
+
+  // Initializes a RevThread object.
+  // - Moves it to the 'Threads' map 
+  // - Adds it's ThreadID to the ThreadQueue to be scheduled
+  void InitThread(std::shared_ptr<RevThread> ThreadToInit);
```
leekillough, line 239: Should be
```c++
void InitThread(const std::shared_ptr<RevThread>& ThreadToInit);
```

Only pass `std::shared_ptr` by value when assigning ownership to a new owner, or when multiple threads (on the host) are using the same `std::shared_ptr` and you could have race conditions. Otherwise each call and return incurs an atomic increment/decrement.

include/RevCPU.h
```diff
@@ -212,6 +221,63 @@ class RevCPU : public SST::Component{
...
+  std::set<uint32_t> BlockedThreads = {};
+
+  // Set of Thread IDs that have completed their execution.
+  std::set<uint32_t> CompletedThreads = {};
+
+  // Random Number Generator for generating new Thread IDs.
+  SST::RNG::MarsagliaRNG* RNG;
+
+  // Generates a new Thread ID using the RNG.
+  uint32_t GetNewThreadID() { return RNG->generateNextUInt32(); }
```
leekillough, line 280: See `include/RevRand.h`. Please do not use another RNG.

include/RevLoader.h
```diff
@@ -298,6 +298,14 @@ class RevLoader {
   /// RevLoader: retrives the elf info structure
   ElfInfo GetInfo() { return elfinfo; }
 
+      /// RevLoader: Gets TLS base address
+      uint64_t GetTLSBaseAddr() { return TLSBaseAddr; }
```
leekillough, line 302: `const`

include/RevLoader.h
```diff
@@ -298,6 +298,14 @@ class RevLoader {
...
+  
+      /// RevLoader: Gets TLS size
+      uint64_t GetTLSSize() { return TLSSize; }
```
leekillough, line 305: `const`

include/RevMem.h
```diff
@@ -99,10 +99,7 @@ class RevMem{
 
     // Override for easy std::cout << *Seg << std::endl;
     friend std::ostream& operator<<(std::ostream& os, MemSegment& obj) {
```
leekillough, line 101: `MemSegment& obj` should be `const MemSegment& obj`

include/RevMem.h
```diff
@@ -260,17 +257,26 @@ class RevMem{
   void SetMaxHeapSize(const unsigned MaxHeapSize){ maxHeapSize = MaxHeapSize; }
 
   /// RevMem: Get memSize value set in .py file
-  uint64_t GetMemSize() const { return memSize; }
+  uint64_t GetMemSize(){ return memSize; }
```
leekillough, line 260: `const`

include/RevMem.h
```diff
@@ -313,25 +325,30 @@ class RevMem{
...
+  uint64_t NextThreadMemAddr = memSize-1024;                  ///< RevMem: Next top address for a new thread's memory (starts at the point the 1024 bytes for argc/argv ends)
+
   uint64_t SearchTLB(uint64_t vAddr);                       ///< RevMem: Used to check the TLB for an entry
   void AddToTLB(uint64_t vAddr, uint64_t physAddr);         ///< RevMem: Used to add a new entry to TLB & LRUQueue
   void FlushTLB();                                          ///< RevMem: Used to flush the TLB & LRUQueue
   uint64_t CalcPhysAddr(uint64_t pageNum, uint64_t vAddr);  ///< RevMem: Used to calculate the physical address based on virtual address
-  bool isValidVirtAddr(const uint64_t vAddr);               ///< RevMem: Used to check if a virtual address exists in MemSegs
-
-  uint32_t PIDCount = 1023;   ///< RevMem: Monotonically increasing PID counter for assigning new PIDs without conflicts
+  bool isValidVirtAddr(const uint64_t& vAddr);               ///< RevMem: Used to check if a virtual address exists in MemSegs
```
leekillough, line 351: Passing `vAddr` by value should be faster than passing by reference, because it is a trivial scalar type `uint64_t`, and references always create aliasing opportunities the compiler might not optimize.

include/RevProc.h
```diff
@@ -122,61 +124,38 @@ class RevProc{
...
 
-  ///< RevProc: PIDs & corresponding RevThreadCtx objects (Software Threads)
-  std::unordered_map<uint32_t, std::shared_ptr<RevThreadCtx>> ThreadTable;
+  ///< RevProc: Get this Proc's feature
+  RevFeature* GetRevFeature(){ return feature; }
```
leekillough, line 153: `const`

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+  return RevProc::ECALL_status_t::SUCCESS;
+}
+
+// 1000, int pthread_create(pthread_t *restrict thread,
+//                          const pthread_attr_t *restrict attr,
+//                          void *(*start_routine)(void *),
+//                          void *restrict arg);
+RevProc::ECALL_status_t RevProc::ECALL_pthread_create(RevInst& inst){
+  output->verbose(CALL_INFO, 2, 0, "ECALL: pthread_create called by thread %" PRIu32 "\n", GetActiveThreadID());
+  uint64_t tidAddr = RegFile->RV64[10]; // Used to hold the new ThreadID
```
leekillough, line 2532: `uint64_t tidAddr = RegFile->GetX<uint64_t>(feature, 10);`

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+  // const pthread_attr_t *restrict attr = (const pthread_attr_t *)RegFile->RV64[11];
+  uint64_t NewThreadPC = RegFile->RV64[11];
```
leekillough, line 2534: `uint64_t NewThreadPC = RegFile->GetX<uint64_t>(feature, 11);`

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+  uint64_t ArgPtr = RegFile->RV64[12];
```
leekillough, line 2535: `uint64_t ArgPtr = RegFile->GetX<uint64_t>(feature, 12);`

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+  unsigned long int NewTID = GetNewThreadID();
+  std::cout << "New TID = " << NewTID << std::endl;
+  CreateThread(NewTID, NewThreadPC, (void*)ArgPtr);
```
leekillough, line 2538: Prefer:

` CreateThread(NewTID, NewThreadPC, reinterpret_cast<void*>(ArgPtr));`

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+
+  mem->WriteMem(feature->GetHartToExec(), tidAddr, sizeof(unsigned long int), &NewTID, REVMEM_FLAGS(0x00));
```
leekillough, line 2540: Prefer:

 `mem->WriteMem(feature->GetHartToExec(), tidAddr, sizeof(NewTID), &NewTID, REVMEM_FLAGS(0));`

(i.e., don't pass a fixed size separately from the data being passed; pass the size of the data being passed.)

src/RevSysCalls.cc
```diff
@@ -0,0 +1,2570 @@
...
+  return RevProc::ECALL_status_t::SUCCESS;
+}
+
+// 1001, int rev_pthread_join(pthread_t thread, void **retval);
+RevProc::ECALL_status_t RevProc::ECALL_pthread_join(RevInst& inst){
+  RevProc::ECALL_status_t rtval = RevProc::ECALL_status_t::CONTINUE;
+  output->verbose(CALL_INFO, 2, 0, "ECALL: pthread_join called by thread %" PRIu32 "\n", GetActiveThreadID());
+  
+  rtval = RevProc::ECALL_status_t::SUCCESS;
```
leekillough, line 2549: `RevProc::` not needed inside `RevProc` class.

src/RevThread.cc
```diff
@@ -0,0 +1,154 @@
...
+  os << "|------|-------|---------------------|-----------------------|\n";
+
+  // Register aliases and descriptions
+  static constexpr const char* aliases[] = {"zero", "ra", "sp", "gp", "tp", "t0", "t1", "t2", "s0", "s1", "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7", "s2", "s3", "s4", "s5", "s6", "s7", "s8", "s9", "s10", "s11", "t3", "t4", "t5", "t6"};
+  static constexpr const char* info[] = {"Zero Register", "Return Address", "Stack Pointer", "Global Pointer", "Thread Pointer", "Temporary Register", "Temporary Register", "Temporary Register", "Callee Saved Register", "Callee Saved Register", "Arg/Return Register", "Arg/Return Register", "Argument Register", "Argument Register", "Argument Register", "Argument Register", "Argument Register", "Argument Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Callee Saved Register", "Temporary Register", "Temporary Register", "Temporary Register", "Temporary Register"};
```
leekillough, line 139: `constexpr` implies `const` for variables. No need to use `const`.

src/RevCPU.cc
```diff
@@ -301,7 +354,8 @@ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
   // Create the completion array
   Enabled = new bool [numCores];
   for( unsigned i=0; i<numCores; i++ ){
-    Enabled[i] = true;
+    // FIXME: This may break PAN
+    Enabled[i] = false;
```
donofrio, line 358: Soooo… does this break PAN?

src/RevCPU.cc
```diff
@@ -2437,12 +2521,246 @@ bool RevCPU::clockTick( SST::Cycle_t currentCycle ){
...
+  ThreadQueue.emplace_back(TID);
+}
+
+void RevCPU::AssignThread(uint32_t ThreadID, uint32_t ProcID){
+  output.verbose(CALL_INFO, 4, 0, "Assigning Thread %" PRIu32 " to Processor %" PRIu32 "\n", ThreadID, ProcID);
+  auto Thread = Threads.at(ThreadID);
+  
+  // Point the regfile of this thread's LSQ to the Proc's LSQ
+  Thread->GetRegFile()->LSQueue = Procs[ProcID]->GetLSQueue();
+
```
donofrio, line 2567: Why is the LSQueue assigned here? Wouldn’t this be part of the RegFile / Thread init?

src/RevCPU.cc
```diff
@@ -2437,12 +2521,246 @@ bool RevCPU::clockTick( SST::Cycle_t currentCycle ){
...
+                         Thread->GetThreadID(), ProcID);
+            break;
+          default: // Should DEFINITELY never happen
+            output.fatal(CALL_INFO, 99, "Error: Thread %u on Core %u is in an unknown state... This is a bug\n", 
+                         Thread->GetThreadID(), ProcID);
+            break;
+        }
+        // State change has been handled, so clear the bit
+        Changes[HartID] = false;
+      }
```
donofrio, line 2743: Some questions about this switch:
 - Can a Proc have more than one thread in RUNNING state at once? If it does, is that an error condition?
 - This statement transitions from BLOCKED to RUNNING - is this the only place that is possible?  The crux of the question is are there multiple places the threadstate can be updated, and is there potential for conflict, or multiple state changes in a single clock

include/RevCPU.h
leekillough: If you need a `uint32_t` random value, use `RevRand(0, UINT32_MAX)`

include/RevProc.h
rkabrick: Do you have any suggestions for a rename? HartUtilization perhaps?

include/RevThread.h
rkabrick: Pushed a fix

include/RevProc.h
donofrio: HartUtilization works for me

include/RevCPU.h
rkabrick: This was intentionally chosen because the SST RNG ensures no ThreadID collisions across the entire SST run. Using RevRand does not offer the same protection.

jleidel: SST has a standard randomization function that allows us to generate pseudorandom numbers that are unique across MPI ranks and/or CXX threads.  In order to support forthcoming scalable multithreaded system architecture simulation, we need to ensure thread ID uniqueness across instances of Rev.  Using the SST rand functions are the easiest way to do so

leekillough: > Using RevRand does not offer the same protection.

This can be fixed.

The same argument could be used to say that `RevRand`, and the old `rand()` code it replaced, could not ensure **k-distrbutivity** across multiple Rev instances (that's the technical term for what you mean when you say "no collisions"), in other parts of the code besides the thread code. The thread code is not special in this regard -- what you're asking for could be asked for by all other uses of `RevRand`.

leekillough: > SST has a standard randomization function that allows us to generate pseudorandom numbers that are unique across MPI ranks and/or CXX threads. In order to support forthcoming scalable multithreaded system architecture simulation, we need to ensure thread ID uniqueness across instances of Rev. Using the SST rand functions are the easiest way to do so

Then `RevRand` needs to be reimplemented using those methods, which should be easy. I do not want a half-dozen RNGs everyone reinvents for their own area of code.

leekillough: I wonder why these issues were not brought up in the review of `RevRand()`. Sure, all of the previous uses of PRNGs were relatively trivial, like fault injection or simulating indeterminate memory latency, but the issue remains.

leekillough: I will immediately work on reimplementing `RevRand()` to use SST's RNGs. In the meantime, @rkabrick I stand by my suggestion that you should use the `RevRand()` API. It should be the central place where all PRNG features / requirements are met.

leekillough: Personal note: I spent 4 years as a hobby studying PRNGs in the late 1980s and early 1990s, and developed the PRNG in HP's Fortran 90 compiler, which was thread-safe and thread-specific but deterministically repeatable within each thread.

leekillough: I have a new version of `RevRand()` which uses SST's generator. No interface changes are necessary. I'll submit a PR shortly. This PR should use `RevRand()`.

include/RevThread.h
```diff
@@ -0,0 +1,109 @@
...
+    uint32_t ThreadID;                      // Thread ID
+    uint32_t ParentThreadID;                // Parent ThreadID
+    uint64_t StackPtr;                      // Initial stack pointer
+    uint64_t FirstPC;                       // Initial PC
+    std::shared_ptr<MemSegment> ThreadMem;  // TLS and Stack memory
+    RevFeature* Feature;                    // Feature set for this thread
+    uint64_t ThreadPtr;                     // Thread pointer
+    ThreadState State = ThreadState::START; // Thread state (Initializes as )
+    RevRegFile RegFile;  // Register file
+    std::vector<uint32_t> ChildrenThreadIDs = {};
```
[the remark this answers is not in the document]
rkabrick: This is currently unused completely. However, I think your point stands should we use it in the future

src/RevCPU.cc
rkabrick: If it is assigned when the Thread is created in RevProc we don't know what Proc it will end up on (ie. What LSQueue it should be assigned to)

src/RevCPU.cc
rkabrick: As of now, RevProc will only ever change a RevThread's state to `DONE` or `BLOCKED`. The rest of State management is handled in RevCPU.

src/RevProc.cc
```diff
@@ -2052,329 +1945,426 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
-    output->fatal(CALL_INFO, -1,
-                  "Failed to load ctx w/ PID=%u into Hart=%u because PID does not exist in ThreadTable",
-                  PID, HartToExec);
-    return false;
-  }
+// TODO: Print Summary 
+void RevProc::PrintStatSummary(){
+  output->verbose(CALL_INFO, 2, 0, "Program execution complete\n");
+  Stats.percentEff = float(Stats.cyclesBusy)/Stats.totalCycles;
+  output->verbose(CALL_INFO, 2, 0,
```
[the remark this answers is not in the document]
rkabrick: While I do agree this goes beyond the scope of this PR.

src/RevCPU.cc
rkabrick: No! 🎉🥳

include/RevThread.h
```diff
@@ -0,0 +1,180 @@
...
+  uint32_t ParentThreadID;                             // Parent ThreadID
+  uint64_t StackPtr;                                   // Initial stack pointer
+  uint64_t FirstPC;                                    // Initial PC
+  std::shared_ptr<RevMem::MemSegment> ThreadMem;       // TLS and Stack memory
+  RevFeature* Feature;                                 // Feature set for this thread
+  uint64_t ThreadPtr;                                  // Thread pointer
+  ThreadState State = ThreadState::START;              // Thread state (Initializes as )
+  RevRegFile RegFile;                                  // Register file
+  std::unordered_set<uint32_t> ChildrenThreadIDs = {}; // Child thread IDs
+  std::set<int> fildes = {0, 1, 2  };      // Default file descriptors
```
leekillough, line 174: `std::unordered_set` would be better than `std::set`  unless you need to traverse the entries in order.

leekillough:
```c++
 // Set the thread pointer
  ThreadPtr = ThreadMem->getTopAddr();
  RegFile.SetX(Feature, 2, ThreadPtr);
```
Shouldn't that be `4`?
---
```c++
std::set<int> fildes = {0, 1, 2  };      // Default file descriptors
```
Shouldn't that be `std::unordered_set<int>` ?

---
```c++
// Add new file descriptor
void RevThread::AddFD(int fd){
  fildes.emplace(fd);
}
```
Should be:
```c++
// Add new file descriptor
void RevThread::AddFD(int fd){
  fildes.insert(fd);
}
```
---

```c++
// See if file descriptor exists/is owned by Ctx
bool RevThread::FindFD(int fd){
  // Check if the fd is owned by the current ctx
  auto it = std::find(fildes.begin(), fildes.end(), fd);
  if( it != fildes.end() ){
    return true;
  }
  return false;
}
```
Should be:
```c++
// See if file descriptor exists/is owned by Ctx
bool RevThread::FindFD(int fd) const {
 return fildes.count(fd);
}
```
---
```c++
// Remove file descriptor from Ctx (ie. rev_close)
bool RevThread::RemoveFD(int fd){
  // Check if the fd is owned by the current ctx
  auto it = std::find(fildes.begin(), fildes.end(), fd);

  if( it != fildes.end() ){
    // fd found, return true
    fildes.erase(it);
    return true;
  }
  // Not found, return false
  return false;
}
```
Should be:
```c++
// Remove file descriptor from Ctx (ie. rev_close)
bool RevThread::RemoveFD(int fd){
  return fildes.erase(fd);
}
```
---
```c++
bool RevThread::AddChildThreadID(uint32_t tid){
  if( std::find(ChildrenThreadIDs.begin(), ChildrenThreadIDs.end(), tid) != ChildrenThreadIDs.end() ){
    ChildrenThreadIDs.emplace(tid);
    return true;
  }
  return false;
}
```
Should be:
```c++
bool RevThread::AddChildThreadID(uint32_t tid){
  return ChildrenThreadIDs.insert(tid).second;
}
```
---
```c++
bool RevThread::RemoveChildThreadID(uint32_t tid){
  auto ChildToErase = std::find(ChildrenThreadIDs.begin(), ChildrenThreadIDs.end(), tid);
  if (ChildToErase != ChildrenThreadIDs.end() ){
    ChildrenThreadIDs.erase(ChildToErase);
    return true;
  }
  return false;
}
```
Should be:
```c++
bool RevThread::RemoveChildThreadID(uint32_t tid){
    return ChildrenThreadIDs.erase(tid);
}
```
If those functions' return value isn't used, they should be changed to `void` instead of `bool`.

include/RevThread.h
rkabrick: Yup Absolutely right!

include/RevThread.h
```diff
@@ -27,37 +27,69 @@ namespace SST::RevCPU{
...
+     FirstPC(inputFirstPC), ThreadMem(inputThreadMem), Feature(inputFeature) {
+  // Create the RegFile for this thread
+  RegFile = RevRegFile(); 
+  
+  // Set the stack pointer 
+  RegFile.SetX(Feature, 2, StackPtr );
+
+  // Set the thread pointer
+  ThreadPtr = ThreadMem->getTopAddr();
+  RegFile.SetX(Feature, 2, ThreadPtr);
```
leekillough, line 68: Isn't the thread pointer `X4`?

[Here is the ABI spec](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/releases/download/v1.0/riscv-abi.pdf).

include/RevThread.h
```diff
@@ -0,0 +1,200 @@
...
+  void AddFD(int fd){ fildes.insert(fd); }
+
+  // Remove file descriptor from this thread (ie. rev_close)
+  void RemoveFD(int fd){ fildes.erase(fd); }
+
+  // See if file descriptor exists/is owned by this thread
+  bool FindFD(int fd){ return fildes.count(fd); }
+
+  // Get the threads 
+  std::unordered_set<int> GetFildes(){ return fildes; }
```
leekillough, line 92: This `GetFildes` function is not used (the comment is confusing). If used, you should consider:
```c++
 const std::unordered_set<int>& GetFildes(){ return fildes; }
```
... to prevent expensive copying of the entire container.

leekillough (approved):

jleidel:
approved to merge

PR #131, opened 2023-09-26
Change RevRand to use SST random number generator

leekillough:
This creates a `RevRNG` wrapper class which implements the [UniformRandomBitGenerator](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) requirements by using the SST Mersenne RNG as the engine for `RevRand()`.

In C++11, a random number distribution is separate from a random number generator engine. An engine only needs to supply uniform random bits of an arbitrary but specific size, and then the engine can be used with C++11 distribution classes to create particular random number distributions.

This PR replaces the `std::mt19937_64` engine with a `SST::RNG`-based engine.

It still initializes the seed differently for different threads running on the same host machine, and keeps state separate for each thread to avoid mutexes, since, upon brief inspection of the SST code, it did not look like it was thread-specific.  (Is SST multithreaded or thread-safe on the host?)


jleidel (approved):
approved

PR #135, opened 2023-09-28
Fix for AMO handling logic and dependency checking

jleidel:
Fix for AMO handling logic and dependency checking

### DIFF
--- a/src/RevProc.cc
+++ b/src/RevProc.cc
@@ -1646,10 +1646,34 @@ bool RevProc::DependencyCheck(uint16_t HartID, const RevInst* I) const {
   const auto* E = &InstTable[I->entry];
   const auto* regFile = GetRegFile(HartID);
 
-    //chek LS queue for outstanding load - ignore r0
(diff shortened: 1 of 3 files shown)


src/RevProc.cc
```diff
@@ -1646,10 +1646,34 @@ bool RevProc::DependencyCheck(uint16_t HartID, const RevInst* I) const {
...
+                                            HartID))) > 0){
+    return true;
+  }
+
+  if((0 != I->rd) &&
+     (regFile->LSQueue->count(make_lsq_hash(I->rd,
+                                            E->rdClass,
+                                            HartID))) > 0){
+    return true;
+  }
```
leekillough, line 1676: Suggest loop for brevity and less repetition:
```c++
for (const auto& [reg, regClass] : { std::tie(I->rs1, E->rs1Class),
                                     std::tie(I->rs2, E->rs2Class), 
                                     std::tie(I->rs3, E->rs3Class),
                                     std::tie(I->rd,  E->rdClass ) }) {
    if(reg != 0 && regFile->LSQueue->count(make_lsq_hash(reg,
                                                         regClass,
                                                         HartId) > 0){
        return true;
     }
}

PR #137, opened 2023-09-29
Cleanup Whitespace

leekillough:
Remove trailing whitespace at the ends of lines
Add newline if it's missing at end of file
Reindent C++ code which may be off-indentation
Add spaces after commas in argument or initializer lists
Untabify indentations

All changes are whitespace-only (if you use `git diff -w` you see only see some blank lines which were removed, and a couple of places where a newline was added so that C++ indentation lines up binary digits perfectly).


leekillough:
Under "Files changed", select the settings wheel, and "Hide whitespace".

The only diffs shown will be newlines added/removed,

PR #140, opened 2023-10-01
VERY minor cmake & gitignore fixes

rkabrick:
1. Removed the previous `SYSCALL_PATH` variable that was used to indicate the directory that held all of the implementation files for each of the syscalls. Never really served a purpose back then and now that the new implementations of the system calls are in a single file `RevSysCalls.h` there's no reason to have it

2. Added `*.dylib` to the `.gitignore` so we have parity between git status on mac & linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-src/librevcpu.so
+*.so
+*.dylib
 *.exe
 *.csv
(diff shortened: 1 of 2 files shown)


leekillough (approved):
LGTM

PR #141, opened 2023-10-02
Update Minimum Cmake Requirement

rkabrick:
- `OUTPUT_ERROR_IS_FATAL` requires 3.19
- All support for CMake < 3.5 is being dropped by CMake itself shortly so this change is needed regardless of the 3.19 req


jleidel (approved):
approved

PR #142, opened 2023-10-02
Stop passing feature pointer when calling RevRegFile functions

leekillough:
Pass `RevFeature` pointer to `RevRegFile` constructor, and extract relevant feature data to store in `RevRegFile`

If you have a `RevRegFile` pointer, use `R->IsRV32` and `R->HasD` (`public` `const` data members) instead of calling feature functions `F->IsRV32()` and `F->HasD()`

Separate `RevRegFile` into its own header file `RevRegFile.h`

Move `_REV_NUM_REGS_` and `_REV_HART_COUNT_` into `common/include/RevCommon.h`

Rearrange `RevRegFile` data members to promote locality among small members and put most frequently-used members near the beginning

Move instruction helper functions in `RevInstTable.h` into a separate `RevInstHelpers.h` header file

Pass `RevRegFile` pointer to `RevExt::Execute` instead of storing it in `RevExt`

Make `RevExt::name` a `std::string_view` for less overhead

Make some data members of `RevExt` and `RevRegFile` `const`, since they shouldn't change after construction

Use move semantics to speed up initialization of instruction vector lists for `RevExt` derived classes

Create streaming IO function for `RevRegFile`, separating it from `RevThread`'s streaming IO function

Put stream IO functions into `.cc` files instead of `.h` files, since they're slow and inlining them just makes code larger

Remove `TRACE_EXECUTE`-enabled diagnostic messages, since they could not compile


include/RevThread.h
```diff
@@ -89,20 +85,19 @@ class RevThread {
   bool FindFD(int fd){ return fildes.count(fd); }
 
   // Get the threads
-  std::unordered_set<int> GetFildes(){ return fildes; }
+  const std::unordered_set<int>& GetFildes(){ return fildes; }
 
   // Register file operations
-  const RevRegFile* GetConstRegFile() const { return &RegFile; }
+  const RevRegFile* GetRegFile() const { return &RegFile; }
```
rkabrick, line 91: I definitely like getting rid of the `GetConstRegFile` function... So overloading this won't be problematic? I didn't realize you could do this with `const` versions of functions but if so then this is great

src/RevExt.cc
```diff
@@ -11,78 +11,31 @@
...
-#endif
-    // retrieve the function pointer for this instruction
-    func = table[Inst].func;
+  } catch(...) {
+    output->fatal(CALL_INFO, -1,
+                  "Error: instruction at index=%u does not exist in extension=%s",
+                  Inst,
+                  name.data());
+    throw;
   }
```
rkabrick, line 33: My limited knowledge of c++17 says `try-catch` is abysmal for performance. Is the reason this is acceptable here because it's only if `TRACE_EXECUTE`?

rkabrick (approved):
As long as `GetRegFile` will still function as expected I'm good to go

include/RevThread.h
leekillough: Yes, Just overload one function with `const` after the parameter list `(...)`, and it will apply to `const`-qualified objects (or non-`const` if a better overload isn't available). You don't have to name it differently. The `cv-qualifier` of `*this`, as well as its optional reference specifier( blank, `&` or `&&` in C++11 or later)  participate in overload resolution.

src/RevExt.cc
leekillough: No, `TRACE_EXECUTE` was removed.

I can remove the `try`-`catch` block too. The original code was very cumbersome by doing its own bounds checking, while I simplified it to use `.at(i)` instead of `[i]`, and catch if `at()` exceptions.

With this and the removal of the uncompilable `TRACE_EXECUTE` code, `RevExt::Execute` is much smaller.

I've heard performance is slow when exceptions occur, but as long as they don't occur, it should be fast. I can change it back, but I'd use a little different syntax.

Issue #134, opened 2023-09-27
Decoder does not recognize fcvt instructions correctly


ukasz:
Instruction fcvt is treated by instruction decoder as invalid instruction.
Some encodings of this instructions are working fine, other don't.
This is a regression, it used to work fine earlier.
Two examples of encodings not working correctly are:
```
fcvt.l.d a0, ft1, rup
fcvt.w.s a0, ft1, rup
```

**To Reproduce**
Compile the following program:
```
// riscv64-linux-gnu-gcc ex2.c -march=rv64g -O2 -o ex2.exe
int main() {
        asm(".word 0x00000513");//li a0, 0
        asm(".word 0xc220b553");//fcvt.l.d a0, ft1, rup
        asm("li a7, 93; li a0, 0; ecall"); //exit(0)
        return 0;
}
```
Run REV with config from ex2 test:
```
REV_VERBOSE_LEVEL=9 REV_EXE=/home/dev/riscv/off/rev3/test/tests_lit/../cpp/polymorphism.exe sst ../devel/test/ex2/rev-test-ex2.py
```

All issues must be accompanied by: 
- The version of SST utilized:
```
# sst --version
SST-Core Version (-dev)
# sst-config --CXX
g++
# sst-config --ELEMENT_CXXFLAGS
 -std=c++17 -fPIC -DHAVE_CONFIG_H -I/home/dev/riscv/sst-core-src/build/../sst-core-install/include
```
- The source code branch: devel
- The last commit hash: 577ac9c2c9aa13a51f6d2cc0b332e76ffc246107
- The compiler utilized to build Rev: gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04.1)
- The compiler utilized to build the test: riscv64tigre-unknown-unknown-elf 6c8d6606259aca093839a1d165ead48680c59066
- The python config file: test/ex2/rev-test-ex2.py

**Expected behavior**
Application ends without throwing invalid instruction error.
**Trace**
```
# REV_EXE=ex2.exe sst ../../devel/test/ex2/rev-test-ex2.py
WARNING: Building component "cpu" with no links assigned.
RevCPU[cpu:RevCPU:0]: Building Rev with 1 cores
RevCPU[cpu:LoadElf:0]: Symbol Table Entry [:0x0]
RevCPU[cpu:LoadElf:0]: Symbol Table Entry [$d:0x570]
[...]

PR #143, opened 2023-10-03
Use ABI names for registers instead of raw numbers; start working on FP status/rounding mode

leekillough:
Move `ZeroExt()` and `SignExt()` functions to `common/include/RevCommon.h` since they are `inline` `constexpr` functions usable everywhere, and do not belong to any particular part of Rev. They replace all of the former macros which did sign- and zero-extension.

Remove unused `FCSR_*` macros since the new design makes them unnecessary (not finalized yet -- another solution besides hard-coded macro constants will be used).

Add `enum` `class` `RndMode` describing FP rounding mode (static or dynamic). May help fix [FCVT problem](https://github.com/tactcomplabs/rev/issues/134).

Add `RevReg` `enum` `class` to describe registers by name instead of value (e.g. `RevReg::a0` instead of `10`; `RevReg::sp` instead of `2`).

Start developing `FCSR` class, which describes floating-point state as `fcsr` member of `RevRegFile`.

Broaden `RevRegFile::GetX` and `RevRegFile::SetX` to accept registers as either integers or `RevReg` enum types.

Same for `RevRegFile::GetFP32` and `RevRegFile::SetFP32`.

Use `RevReg::*` ABI register names instead of hard-coded numbers in the source code.

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -11,8 +11,10 @@
 #ifndef __REV_COMMON__
 #define __REV_COMMON__
 
-#include <functional>
+#include <cstddef>
 #include <cstdint>
+#include <functional>
+#include <type_traits>
 
 #ifndef _REV_NUM_REGS_
 #define _REV_NUM_REGS_ 32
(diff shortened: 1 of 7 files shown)


donofrio (approved):
Looks good - enum improves readability

rkabrick (approved):
Looks good to me

jleidel (approved):
good to merge

PR #145, opened 2023-10-04
Rename RevProc to RevCore

leekillough:
Rename `RevProc` to `RevCore`, to emphasize the hierarchy `RevCPU` **->** `RevCore` **->** `RevHart`

The original nomenclature `RevProc` is confusing -- does it refer to a physical **Proc**essor, OS **Proc**ess, some kind of **Proc**edure? And since `RevCoProc` refers to a `CoProcessor`, this makes `RevProc` even more ambiguous.

The proper RISC-V hierarchy is `RevCPU` **->** `RevCore` **->** `RevHart`

@rkabrick  and I already discussed this and agreed with this naming change.

### DIFF
--- a/documentation/ExtensionDevelopment.md
+++ b/documentation/ExtensionDevelopment.md
@@ -21,7 +21,7 @@ From the base Rev directory, all the source code resides in `src`.  The instruct
 | `src/RevInstTable.h` | Contains the base strucutures utilized to create each extension as well as functions to assist in instruction implementation. |
 | `src/RevFeature.h` | Contains the feature to extension mappings. |
 | `src/RevMem.h` | Contains all the interfaces for reading/writing memory. |
-| `src/RevProc.cc` | Contains the main simulation driver and instruction table loader. |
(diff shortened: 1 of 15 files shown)


jleidel:
we need to reject and close this PR in the short term.  we have immediate deliverables where this PR will confuse the end user/developer

rkabrick:
I definitely like this but like John said there's too many people looking at the code for the short term. Perhaps opening an issue will offer some informal documentation of our intention to change it in the future?

jleidel:
I don't mind leaving this as a draft as well.

leekillough:
Recreation steps, for later reference (took less than 15 minutes):
```
git clean -dfx 
git mv src/RevProc.cc src/RevCore.cc
git mv include/RevProc.h include/RevCore.h
find . -name .git -prune -o \( -type f -exec grep -q RevProc {} \; -exec perl -pi -e 's/RevProc/RevCore/g' {} \; \)
```

leekillough:
This will be merged after #256 is merged.

Issue #144, opened 2023-10-04
[P1] JAL instruction does behave strage 


marcinkoss:
**Describe the bug**
Discovered when run llvm/libcxx/tests -> llvm-project/libcxx/test/std/utilities/format/format.range/format.range.fmtset/format.functions.format.pass.cpp
Test, when build with LLVM toolchain generates:
```
  9b42c: 83 35 04 fb   ld      a1, -80(s0)
   9b430: 13 05 04 fc   addi    a0, s0, -64
   9b434: 93 06 84 fa   addi    a3, s0, -88
   9b438: ef 50 8b 35   jal     0x150790 <_ZNSt3__13setIcNS_4lessIcEENS_9allocatorIcEEEC2B7v160000ESt16initializer_listIcERKS2_>
```
and:
```
0000000000150790 <_ZNSt3__13setIcNS_4lessIcEENS_9allocatorIcEEEC2B7v160000ESt16initializer_listIcERKS2_>:
  150790: 13 01 01 fa   addi    sp, sp, -96
  150794: 23 3c 11 04   sd      ra, 88(sp)
  150798: 23 38 81 04   sd      s0, 80(sp)
```

Now, when binary is run on REV model, when PC reaches instruction at address 0x9b438 then JAL instruction does jump to 0x50790 instead of 0x150790, and finish with error "cannot decode instruction at address 0x50790)

I added few logs in REV code, and IMM = b5358 and PC is 9B438 (which gives 0x150790):
RevCPU[cpu:ClockTick:69000]: Core 0; Hart 0; Thread 1; Executing PC= 0x9b438
jal imm = b5358 imm sign ext = ffffb5358

I think it might some arithmetic error on U2 numbers inside REV and "1" is lost somewhere.

**To Reproduce**
All issues must be accompanied by: 
- The version of SST utilized (`sst --version`; `sst-config --CXX`; `sst-config --ELEMENT_CXXFLAGS`)
#sst --version
SST-Core Version (-dev)
# sst-config --CXX
g++
# sst-config --ELEMENT_CXXFLAGS
 -std=c++17 -fPIC -DHAVE_CONFIG_H -I<sst-core-path>/build/../sst-core-install/include

- The source code branch being used
devel 
- The last commit hash
sha e1c1c71b28cd5dee0439c1ef5d20a0e64846fb57

- The compiler utilized to build Rev
gcc/g++, 11.3.0

- The compiler utilized to build the test
LLVM clang 

[...]

PR #147, opened 2023-10-04
JAL fix

leekillough:
`JAL` instruction implementation was only considering the bottom 20 bits of the immediate offset, even though the decoder had already shifted the 20-bit immediate offset left one bit.

Added tests for all combinations of the top two most-significant bits of `JAL` offset.

Fixes https://github.com/tactcomplabs/rev/issues/144


jleidel (approved):
good to merge

PR #150, opened 2023-10-05
Add cpuinfo syscall

ukasz:
Add a new REV-specific syscall to check number of harts inside a REV CORE. This feature is needed to be able to implement functions like omp_get_num_threads
without hardcoding this inside a library.
Add a simple test validating new functionality.


rkabrick:
Everything in here looks fine. I appreciate you following all of the same styling. 

I'm currently working on making the number of HARTs configurable in the python (ie. `numHartsPerCore`). When I submit that PR would you want the value of `nproc` to be a product of `RevProc's (Cores) * the number of Harts`? Or would you prefer it just return the `numCores` value specified in the python?

ukasz:
> I'm currently working on making the number of HARTs configurable in the python (ie. `numHartsPerCore`). When I submit that PR would you want the value of `nproc` to be a product of `RevProc's (Cores) * the number of Harts`? Or would you prefer it just return the `numCores` value specified in the python?

Hmm, In that case it would have to be a product of cores * harts, but maybe we should go straight to something more generic, a syscall to fill a struct with multiple fields lets call it cpuinfo with things like cores, harts, riscv extension flags and other things that workloads may find useful, but do not change after simulation starts. If you're fine with that I'll change PR to go this way.

rkabrick:
That works for me. Just be sure to document all the possible arguments

ukasz:
Refactored/updated code

src/RevProc.cc
```diff
@@ -2382,6 +2382,7 @@ void RevProc::InitEcallTable(){
     { 438, &RevProc::ECALL_pidfd_getfd},            //  rev_pidfd_getfd(int pidfd, int fd, unsigned int flags)
     { 439, &RevProc::ECALL_faccessat2},             //  rev_faccessat2(int dfd, const char  *filename, int mode, int flags)
     { 440, &RevProc::ECALL_process_madvise},        //  rev_process_madvise(int pidfd, const struct iovec  *vec, size_t vlen, int behavior, unsigned int flags)
+    { 500, &RevProc::ECALL_cpuinfo},                //  rev_cpuinfo(struct cpuinfo *info)
```
ukasz, line 2385: Should be rev_cpuinfo

jleidel:
good to merge

PR #152, opened 2023-10-05
Fix EcallState to be a reference instead of a local variable

leekillough:
Fix `EcallState` to be a reference instead of a local variable so that it actually modifies the referenced state rather than a local copy of it.

Add `const` version of `GetThreadsThatChangedState()`  (there was a comment suggesting a `const` version should be available).

Add `break` to `switch` `case` to prevent fallthrough warning (=error). If fall-through is desired, then `[[fallthrough]]` [attribute can be specified](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) before next `case`.

Minor whitespace changes because my `emacs` configuration automatically removes trailing whitespace at the ends of lines.

### DIFF
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -125,8 +125,9 @@ public:
   RevMem& GetMem() const { return *mem; }
 
   // Get the hart ids of the threads that changed state
-  // TODO: Maybe ThreadIDs? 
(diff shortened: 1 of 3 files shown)


jleidel (approved):
approved to merge

PR #154, opened 2023-10-09
Add function to query whether any scoreboard bits are set

leekillough:
Switch to using `std::bitset` and single scoreboard registers (no `union`s).

The 32/64-bit scoreboard registers used the same space and size, so the `union`s were redundant.

With `std::bitset` there are built-in functions for testing whether any or all bits are set.

Also takes 1/8th the space of an array of `bool`.

### DIFF
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -690,6 +690,12 @@ private:
   /// RevProc: Determine next thread to execute
   uint16_t GetHartID() const;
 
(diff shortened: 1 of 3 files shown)


rkabrick (approved):
Looks good to me... Dave made the scoreboard so as long as he is cool with that works for me. Once this is merged I'll update the code in my PR to use it.

donofrio (approved):
Looks good - clear to merge

PR #155, opened 2023-10-10
Hart rework

rkabrick:
TL;DR - Added a `RevHart` class & added a `numHarts` config option to Rev's Python options. 

Currently `fmove` and `fcvt` tests are timing out but the rest is working splendidly. I will push fixes for these prior to merging. 

All the relevant design choices I am unsure about are marked `TODO: (BEFORE MERGE)` so before the bug fixes are pushed please check out those points in the code and leave your thoughts. 

I will also submit more of a formal write up of all the changes tomorrow

### DIFF
--- a/include/RevHart.h
+++ b/include/RevHart.h
@@ -25,12 +25,6 @@ class RevHart{
   ///< RevHart: State management object when a Hart is executing a system call
   std::unique_ptr<EcallState> Ecall;
 
-  ///< RevHart: Hart is ready decode next instruction
-  bool ReadyToDecode;
-  
-  ///< RevHart: Hart is ready to execute its decoded inst (ie. no depedenecies)
-  bool ReadyToExecute; 
-
   ///< RevHart: Thread currently executing on this Hart
   uint32_t AssignedThreadID;
 
@@ -39,8 +33,6 @@ public:
   RevHart(uint16_t id) 
       : ID(id), 
         Ecall(std::make_unique<EcallState>()), 
-        ReadyToDecode(false), 
-        ReadyToExecute(false), 
         AssignedThreadID(_INVALID_TID_) {
   }
 
@@ -50,38 +42,16 @@ public:
   ///< RevHart: Get the EcallState
   EcallState& GetEcallState(){ return *Ecall; }
 
-  ///< RevHart: Returns if a Hart is ready to decode its next instruction (CTS)
-  bool isReadyToDecode(){ return ReadyToDecode && (AssignedThreadID != _INVALID_TID_); }
-  
-  ///< RevHart: Returns if a Hart is ready to execute the decoded instruction (CTE)
-  bool isReadyToExecute() const { return ReadyToExecute && (AssignedThreadID != _INVALID_TID_); }
-
-  // ReadyToDecode == CTS
-  void SetReadyToDecode(bool rtd ){ ReadyToDecode = rtd; }
-
-  // ReadyToExecute == CTE
-  void SetReadyToExecute(bool rte){ ReadyToExecute = rte; }
-
   ///< RevHart: Get Hart's ID 
   const uint16_t& GetID(){ return ID; }
 
   ///< RevHart: Returns the ThreadID of the assigned thread
-  /// TODO: Potentially move the checking logic here for if a thread is not assigned? Right now it's handled elsewhere
   uint32_t GetAssignedThreadID(){ return AssignedThreadID; }
 
   ///< RevHart: Assigns a RevThread to this Hart
   void AssignThread(const uint32_t& ThreadID){ AssignedThreadID = ThreadID; }
 
-  ///< RevHart: Removed a RevThread from this Hart
-  // TODO: Maybe check if Thread has any reasons it can't be removed (ie. Outstanding Load)
-  void UnassignThread(){ AssignedThreadID = _INVALID_TID_; }
-
 };
 
-//EcallStatus RevHart::EcallLoadAndParseString(...){
-//    if(Ecall.buf) {
-//      // do something with Hart (*this). Its members can be written as ordinary variables.
-//    }
-//}
 };
 #endif
--- a/include/RevHart.h
+++ b/include/RevHart.h
@@ -1,0 +1,90 @@
+//
+// _RevHart_h_
+//
+// Copyright (C) 2017-2023 Tactical Computing Laboratories, LLC
+// All Rights Reserved
+// contact@tactcomplabs.com
+//
+// See LICENSE in the top level directory for licensing details
+//
+
+#ifndef _SST_REVCPU_REVHART_H_
+#define _SST_REVCPU_REVHART_H_
+
+// -- SST Headers
+#include "RevSysCalls.h"
+#include "RevThread.h"
+#include "SST.h"
+
+namespace SST::RevCPU{
+
+class RevHart{
+  ///< RevHart: Id for the Hart (0,1,2,3,etc)
+  uint16_t ID;
+
+  ///< RevHart: State management object when a Hart is executing a system call
+  std::unique_ptr<EcallState> Ecall;
+
+  ///< RevHart: Hart is ready decode next instruction
+  bool ReadyToDecode;
+  
+  ///< RevHart: Hart is ready to execute its decoded inst (ie. no depedenecies)
+  bool ReadyToExecute; 
+
+  ///< RevHart: Thread currently executing on this Hart
+  uint32_t AssignedThreadID;
+
+public:
+  ///< RevHart: Constructor
+  RevHart(uint16_t id) 
+      : ID(id), 
+        Ecall(std::make_unique<EcallState>()), 
+        ReadyToDecode(false), 
+        ReadyToExecute(false), 
+        AssignedThreadID(_INVALID_TID_) {
+  }
+
+  ///< RevHart: Destructor
+  ~RevHart(){}
+
+  ///< RevHart: Get the EcallState
+  EcallState& GetEcallState(){ return *Ecall; }
+
+  ///< RevHart: Returns if a Hart is ready to decode its next instruction (CTS)
+  bool isReadyToDecode(){ return ReadyToDecode && (AssignedThreadID != _INVALID_TID_); }
+  
+  ///< RevHart: Returns if a Hart is ready to execute the decoded instruction (CTE)
+  bool isReadyToExecute() const { return ReadyToExecute && (AssignedThreadID != _INVALID_TID_); }
+
+  // ReadyToDecode == CTS
+  void SetReadyToDecode(bool rtd ){ ReadyToDecode = rtd; }
+
+  // ReadyToExecute == CTE
+  void SetReadyToExecute(bool rte){ ReadyToExecute = rte; }
+
+  ///< RevHart: Get Hart's ID 
+  const uint16_t& GetID(){ return ID; }
+
+  ///< RevHart: Returns true if a thread is assigned to this Hart
+  bool HasThread(){ return AssignedThreadID != _INVALID_TID_; }
+
+  ///< RevHart: Returns the ThreadID of the assigned thread
+  /// TODO: Potentially move the checking logic here for if a thread is not assigned? Right now it's handled elsewhere
+  uint32_t GetAssignedThreadID(){ return AssignedThreadID; }
+
+  ///< RevHart: Assigns a RevThread to this Hart
+  void AssignThread(const uint32_t& ThreadID){ AssignedThreadID = ThreadID; }
+
+  ///< RevHart: Removed a RevThread from this Hart
+  // TODO: Maybe check if Thread has any reasons it can't be removed (ie. Outstanding Load)
+  void UnassignThread(){ AssignedThreadID = _INVALID_TID_; }
+
+};
+
+//EcallStatus RevHart::EcallLoadAndParseString(...){
+//    if(Ecall.buf) {
+//      // do something with Hart (*this). Its members can be written as ordinary variables.
+//    }
+//}
+};
+#endif
--- a/include/RevHart.h
+++ b/include/RevHart.h
@@ -65,9 +65,6 @@ public:
   ///< RevHart: Get Hart's ID 
   const uint16_t& GetID(){ return ID; }
 
-  ///< RevHart: Returns true if a thread is assigned to this Hart
-  bool HasThread(){ return AssignedThreadID != _INVALID_TID_; }
-
   ///< RevHart: Returns the ThreadID of the assigned thread
   /// TODO: Potentially move the checking logic here for if a thread is not assigned? Right now it's handled elsewhere
   uint32_t GetAssignedThreadID(){ return AssignedThreadID; }
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -120,44 +120,11 @@ public:
     uint64_t cyclesIdle_MemoryFetch;
   };
 
+  ///< RevProc: Returns this proc's stats
   RevProcStats GetStats() { Stats.memStats = mem->memStats; return Stats; }
 
+  ///< RevProc: Returns a pointer to the RevMem object
   RevMem& GetMem() const { return *mem; }
-
-  // Get the hart ids of the threads that changed state
-  // TODO: Maybe ThreadIDs?
-  std::queue<std::shared_ptr<RevThread>>& GetThreadsThatChangedState() { return ThreadsThatChangedState; }
-  const std::queue<std::shared_ptr<RevThread>>& GetThreadsThatChangedState() const { return ThreadsThatChangedState; }
-
-  /// RevProc: SpawnThread creates a new thread and returns its ThreadID
-  void CreateThread(uint32_t NewTid, uint64_t fn, void* arg);
-
-  /// RevProc: Returns the current HartToExec active pid
-  uint32_t GetActiveThreadID();
-
-  // TODO: Comment
-  bool ThreadCanBeRemoved(const uint32_t& ThreadID) ;
-
-  /// RevProc: Queue of threads to be spawned (RevProc creates the thread, RevCPU readies it for execution)
-  const std::queue<std::shared_ptr<RevThread>>& GetNewThreadInfo() { return NewThreadInfo; }
-
-  ///< RevProc: Holds the info for all new threads to be spawned
-  ///           Right now this is through the following functions:
-  ///           - rev_pthread_create
-  std::queue<std::shared_ptr<RevThread>> NewThreadInfo;
-
-  /// RevProc: Gets the thread that is currently executing on the HartToExec
-  /// TODO: Make this safe
-  const std::shared_ptr<RevThread>& GetThreadOnHart(uint16_t HartID){ return AssignedThreads.at(Harts.at(HartID)->GetAssignedThreadID()); }
-
-  // TODO: Make this safe
-  auto RegFileOnHart(uint16_t HartID){ return AssignedThreads.at(Harts.at(HartID)->GetAssignedThreadID())->GetRegFile(); }
-
-  ///< RevProc: Used for scheduling in RevCPU (if Utilization < 1, there is at least 1 unoccupied HART )
-  double GetHartUtilization() const { return (AssignedThreads.size() * 100.0) / Harts.size(); }
-
-  std::vector<bool> HART_CTS; ///< RevProc: Thread is clear to start (proceed with decode)
-  std::vector<bool> HART_CTE; ///< RevProc: Thread is clear to execute (no register dependencides)
 
   ///< RevProc: Get this Proc's feature
   RevFeature* GetRevFeature() const { return feature; }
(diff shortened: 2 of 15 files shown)


include/RevHart.h
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  ///< RevHart: Destructor
+  ~RevHart(){}
+
+  ///< RevHart: Get the EcallState
+  EcallState& GetEcallState(){ return *Ecall; }
+
+  ///< RevHart: Get Hart's ID 
+  const uint16_t& GetID(){ return ID; }
```
leekillough, line 46: I would just return a `uint16_t` instead of a `const uint16_t&`

include/RevProc.h
```diff
@@ -120,41 +120,50 @@ class RevProc{
     uint64_t cyclesIdle_MemoryFetch;
   };
 
+  ///< RevProc: Returns this proc's stats
   RevProcStats GetStats() { Stats.memStats = mem->memStats; return Stats; }
```
leekillough, line 125: If `RevProcStats` is a large structure, it might be faster to use a return type of `const RevProcStats&` instead of `RevProcStats` to avoid (or at least delay) expensive copies.

include/RevProc.h
```diff
@@ -120,41 +120,50 @@ class RevProc{
...
+  }
 
   ///< RevProc: Used for scheduling in RevCPU (if Utilization < 1, there is at least 1 unoccupied HART )
-  double GetHartUtilization() const { return (AssignedThreads.size() * 100.0) / _REV_HART_COUNT_; }
+  double GetHartUtilization() const { return (AssignedThreads.size() * 100.0) / Harts.size(); }
 
-  ///< RevProc: Get this Proc's feature
-  RevFeature* GetRevFeature() const { return feature; }
+  // Get the hart ids of the threads that changed state
+  std::queue<std::shared_ptr<RevThread>>& GetThreadsThatChangedState() { return ThreadsThatChangedState; }
```
leekillough, line 151: Probably should be `const std::queue<std::shared_ptr<RevThread>>& GetThreadsThatChangedState()` unless the returned reference is intended to be mutable. Can be abbreviated `const auto& GetThreadsThatChangedState()` for brevity.

include/RevProc.h
```diff
@@ -120,41 +120,50 @@ class RevProc{
...
+  void CreateThread(uint32_t NewThreadID, uint64_t fn, void* arg);
 
-  ///< RevProc: Get pointer to Load / Store queue used to track memory operations
-  std::shared_ptr<std::unordered_map<uint64_t, MemReq>> GetLSQueue(){ return LSQueue; }
+  /// RevProc: Returns the current HartToExec active pid
+  uint32_t GetActiveThreadID();
+
+  // TODO: (BEFORE MERGE) - Potentially name this something else? It currently returns true is the scoreboard of a regfile is empty
+  bool ThreadCanBeRemoved(const uint32_t& ThreadID); ///< RevProC: Used to check if a thread can be removed (ie. has no outstanding loads)
+
```
leekillough, line 161: I assume "BEFORE MERGE" is before https://github.com/tactcomplabs/rev/pull/154

include/RevProc.h
```diff
@@ -167,23 +176,32 @@ class RevProc{
...
   RevMem *mem;              ///< RevProc: memory object
   RevCoProc* coProc;        ///< RevProc: attached co-processor
   RevLoader *loader;        ///< RevProc: loader object
 
-  /// ThreadIDs assigned to this RevProc (AssignedThreads[i] will give you the RevThread executing on Hart i)
-  std::vector<std::shared_ptr<RevThread>>& AssignedThreads;
+  /// TODO: (BEFORE MERGE) Potentailly turn into std::unordered_set<std::shared_ptr<RevThread>> but I'm not sure 
+  std::unordered_map<uint32_t, std::shared_ptr<RevThread>>& AssignedThreads;
+
+  // TODO: (BEFORE MERGE) Potentially make inline?
```
leekillough, line 192: By being defined inside the class, it is `inline` _by definition_.

src/RevCPU.cc
```diff
@@ -2410,11 +2414,11 @@ bool RevCPU::clockTick( SST::Cycle_t currentCycle ){
         }
         UpdateCoreStatistics(i);
         Enabled[i] = false;
-        output.verbose(CALL_INFO, 5, 0, "Closing Processor %" PRIu64 " at Cycle: %" PRIu64 "\n",
+        output.verbose(CALL_INFO, 5, 0, "Closing Processor %zu at Cycle: %" PRIu64 "\n",
                        i, currentCycle);
```
leekillough, line 2418: `%zu` is correct for `size_t`

src/RevCPU.cc
```diff
@@ -2546,6 +2546,7 @@ void RevCPU::InitThread(std::shared_ptr<RevThread>& ThreadToInit){
 
 void RevCPU::AssignThread(uint32_t ThreadID, uint32_t ProcID){
   output.verbose(CALL_INFO, 4, 0, "Assigning Thread %" PRIu32 " to Processor %" PRIu32 "\n", ThreadID, ProcID);
+  // TODO: (BEFORE MERGE): Lee, do I need to make this an auto&
```
leekillough, line 2549: No, since you're only using it as a value, as a pointer in `Thread->`, and as a non-reference parameter in `... .emplace(ThreadID, Thread);`

src/RevCPU.cc
```diff
@@ -2628,17 +2631,17 @@ void RevCPU::SetupArgs(uint32_t ThreadIDToSetup, RevFeature* feature){
...
 void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
+  // print the thread queue
   // Get utilization info
   double Util = Procs[ProcID]->GetHartUtilization();
-  if( Util > 0.0 ){
-    output.verbose(CALL_INFO, 10, 0, "Core %" PRIu32 " utilization: %.2f%%\n", ProcID, Util);
-  }
+  //if( Util > 0.0 ){
+    output.verbose(CALL_INFO, 11, 0, "Core %" PRIu32 " utilization: %.2f%%\n", ProcID, Util);
+  //}
```
leekillough, line 2639: Is this intended to be checked-in as part of the PR, or is this leftover debugging output?

src/RevCPU.cc
```diff
@@ -2660,95 +2663,92 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
-        default: // Should DEFINITELY never happen
-          output.fatal(CALL_INFO, 99, "Error: Thread %" PRIu32 " on Core %" PRIu32 " is in an unknown state... This is a bug\n",
-                       Thread->GetThreadID(), ProcID);
-          break;
-        }
-        // State change has been handled, so clear the bit
-        Changes[HartID] = false;
+  // NOTE: At this point we handle EVERY thread that changed state
+  while( !Procs[ProcID]->GetThreadsThatChangedState().empty() ){
+    auto& Thread = Procs[ProcID]->GetThreadsThatChangedState().front();
```
leekillough, line 2668: FYI:
`auto& Thread` means a reference to an lvalue (non-temporary which has an address) which might be `const`

`auto Thread` means a local copy of a value

`auto&& Thread` means a reference to an lvalue or rvalue, depending on the RHS, which might be `const`. It is the "universal reference" which can bind to anything. If you don't know anything about the RHS, then `auto&&` is the most universal.

src/RevCPU.cc
```diff
@@ -2660,95 +2663,92 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
+        // -- 3a.
+        BlockedThreads.emplace(Thread->GetThreadID());
+
+        std::cout << Thread->to_string();
+        
+        // -- 3b.
+        AssignedThreads.at(ProcID).erase(Thread->GetThreadID());
+        
+        Threads.at(Thread->GetThreadID())->GetRegFile()->SetLSQueue(nullptr);
+        Threads.at(Thread->GetThreadID())->GetRegFile()->SetMarkLoadComplete(nullptr);
```
leekillough, line 2713: I might avoid searching twice:
```c++
RevRegFile* regFile = Threads.at(Thread->GetThreadID())->GetRegFile();
regFile->SetLSQueue(nullptr);
regFile->SetMarkLoadComplete(nullptr);
```

src/RevProc.cc
```diff
@@ -16,18 +16,19 @@ using MemSegment = RevMem::MemSegment;
...
   : Halted(false), Stalled(false), SingleStep(false),
     CrackFault(false), ALUFault(false), fault_width(0),
-    id(Id), HartToDecode(0), HartToExec(0), Retired(0x00ull),
-    opts(Opts), mem(Mem), loader(Loader), AssignedThreads(AssignedThreads),
-    GetNewThreadID(GetNewTID), output(Output), feature(nullptr),
-    PExec(nullptr), sfetch(nullptr) {
+    id(Id), HartToDecode(0), HartToExec(0), Retired(0x00ull), 
+    numHarts(NumHarts), opts(Opts), mem(Mem), loader(Loader), 
+    AssignedThreads(AssignedThreads), GetNewThreadID(GetNewTID), 
+    output(Output), feature(nullptr), PExec(nullptr), sfetch(nullptr) {
```
leekillough, line 31: By the way, you can use constructor parameter names the same as member names.
```c++
: numHarts(numHarts), opts(opts), mem(mem), loader(loader), 
AssignedThreads(AssignedThreads), GetNewThreadID(GetNewTID), 
output(output), feature(nullptr), PExec(nullptr), sfetch(nullptr) {
```
There is no need to distinguish the constructor parameter names from the member names. In the initialization list (the `: var1(value1), var2(value2), ...`), the member names are in the class scope, while their initialization expressions are in the constructor function scope, so if they are the same, there is no conflict.

src/RevProc.cc
```diff
@@ -46,6 +47,13 @@ RevProc::RevProc( unsigned Id,
     coProc = NULL;
   }
 
+  // Create the Hart Objects
+  for( size_t i=0; i<numHarts; i++ ){
+    Harts.emplace_back(std::make_unique<RevHart>(i));
+    HART_CTS.emplace_back(false);
+    HART_CTE.emplace_back(false);
+  }
```
leekillough, line 55: This explicit initialization of `HART_CTS` and `HART_CTE` are not necessary if the constructor above has:
```c++
HART_CTS(NumHarts), HART_CTE(NumHarts)
```
A vector constructed with a size argument value-initializes (zeros or default-constructs) its members. For a vector of `bool`, this means initializing a vector of `false`.

Right now `HART_CTS` and `HART_CTE` are default-initialized to size-0 vectors since they are not listed in the constructor, and then they are appended on. But you can initialize them to `numHarts` vectors of `false` by initializing them with their `numHarts` size.

src/RevProc.cc
```diff
@@ -1564,6 +1574,31 @@ void RevProc::HandleALUFault(unsigned width){
                   "FAULT:ALU: ALU fault injected into next retire cycle\n");
 }
 
+bool RevProc::ThreadCanBeRemoved(const uint32_t& ThreadID) {
+  auto* regFile = AssignedThreads.at(ThreadID)->GetRegFile();
+
+  // Check the if regfile scoreboard is anything but clear 
+  // if not, this thread has outstanding requests and can't be removed
+  // TODO: (PRE-MERGE) Do this in a more ~Lee~ fashion
+  if( feature->IsRV32() ){
```
leekillough, line 1583: `AnyDependency()` takes an `isFloat` parameter, but it can be changed to detect **any** dependency by ORing the RV and FP scoreboards, and then all of this can be replaced with `AnyDependency()`

src/RevProc.cc
```diff
@@ -1635,22 +1670,25 @@ void RevProc::DependencySet(uint16_t HartID, uint16_t RegNum,
   }
 }
 
-uint16_t RevProc::GetHartID()const{
-  if(HART_CTS.none()) { return HartToDecode;};
+uint16_t RevProc::GetHartID() const {
+  // TODO: (BEFORE MERGE) --- Potentially make this a bitset that is MAX HARTS long
+  if( std::none_of(HART_CTS.begin(), HART_CTS.end(), [](bool x){ return true; } )){
```
leekillough, line 1675: I'm not sure I understand the predicate, always `true`

src/RevProc.cc
```diff
@@ -1635,22 +1670,25 @@ void RevProc::DependencySet(uint16_t HartID, uint16_t RegNum,
...
 
   uint16_t nextID = HartToDecode;
   if(HART_CTS[HartToDecode]){
     nextID = HartToDecode;
   }else{
-    for(int tID = 0; tID < _REV_HART_COUNT_; tID++){
+    for(size_t tID = 0; tID < Harts.size(); tID++){
       nextID++;
-      if(nextID >= _REV_HART_COUNT_){
+      if(nextID >= Harts.size()){
```
leekillough, line 1685: I hated this function. In my initial rework, it printed out messages every time, so I just backed out my changes. I think there is a better more elegant way to handle it.

src/RevProc.cc
```diff
@@ -1696,16 +1734,24 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
+    if( HartHasThread(HartID) ){
+      HART_CTS[HartID] = (ThreadOnHart(HartID)->GetRegFile()->cost == 0);
+    } else {
+      HART_CTS[HartID] = false;
+    }
   }
 
-  if( HART_CTS.any() && (!Halted)) {
+  // TODO: (BEFORE MERGE) --- Either leave this or switch to bitset that's length is always "MAX HARTS" and then use HART_CTS.any()
+  if( std::any_of(HART_CTS.begin(), HART_CTS.end(), [](bool x){ return true; }) && (!Halted)) {
```
leekillough, line 1746: Predicate?

src/RevProc.cc
```diff
@@ -1893,16 +1940,16 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
     // and the HART is halted
     output->verbose(CALL_INFO, 9, 0,
-                    "Core %u ; No available thread to exec PC= 0x%" PRIx64 "\n",
-                    id, ExecPC);
+                    "Core %" PRIu16 " ; Hart %" PRIu16 "; No available hart to exec PC= 0x%" PRIx64 "\n",
+                    id, HartToDecode, ExecPC);
     rtn = true;
     Stats.cyclesIdle_Total++;
-    if(HART_CTE.any()){
+    if( std::any_of(HART_CTE.begin(), HART_CTE.end(), [](bool x){ return true; })) {
```
leekillough, line 1952: ?

Seems like all of these could be replaced by:
```c++
if( !HART_CTE.empty() ){
```

src/RevProc.cc
```diff
@@ -1911,16 +1958,23 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
+                      "Core %" PRIu16 "; Hart %" PRIu16 "; Thread %" PRIu32 "; Retiring PC= 0x%" PRIx64 "\n",
+                      id, HartID, ThreadOnHart(HartID)->GetThreadID(), ExecPC);
       Retired++;
-      DependencyClear(tID, &(Pipeline.front().second));
+
+      // Only clear the dependency if there is no outstanding load
+      if((Pipeline.front().second.rd != 0) && 
+        (RegFile->GetLSQueue()->count(make_lsq_hash(Pipeline.front().second.rd,
+                                                                 InstTable[Pipeline.front().second.entry].rdClass,
+                                                                 HartID))) == 0){
```
leekillough, line 1973: THIS?

src/RevProc.cc
```diff
@@ -2011,11 +2061,15 @@ void RevProc::PrintStatSummary(){
 }
 
 RevRegFile* RevProc::GetRegFile(uint16_t HartID) const{
-  if( AssignedThreads.size() < HartID ){
+  if( !HartHasThread(HartID) ){
+    std::cout << "ThreadMemSegs: " << std::endl;
+    for( auto Seg : mem->GetThreadMemSegs() ){
+      std::cout << *Seg << std::endl;
+    }
```
leekillough, line 2068: Debug messages intended in merge?

src/RevProc.cc
```diff
@@ -2369,94 +2424,67 @@ void RevProc::InitEcallTable(){
...
- * supported exceptions at this point there is no need just yet.
- */
+// This is the function that is called when an ECALL exception is detected inside ClockTick
+// Currently the only way to set this exception is by Ext->Execute(....) an ECALL instruction
+// Eventually this will be integrated into a TrapHandler however since ECALLs are the only
+// supported exceptions at this point there is no need just yet.
 void RevProc::ExecEcall(RevInst& inst){
   // a7 register = ecall code
-  auto EcallCode = RegFile->GetX<uint64_t>(RevReg::a7);
+  auto EcallCode = RegFile->GetX<uint64_t>(17);
```
leekillough, line 2433: `RevReg::a7`

src/RevProc.cc
```diff
@@ -2369,94 +2424,67 @@ void RevProc::InitEcallTable(){
...
-    }
-    else {
-      output->fatal(CALL_INFO, 1, "HartToDecode = %" PRIu32 " but there are only %" PRIu64 " threads assigned to this Proc. This might be a bug\n", HartToDecode, AssignedThreads.size());
+// TODO: (BEFORE MERGE) Potentially add a faster way to find available Harts
+// Also I don't think we ever get here without first knowing that there is a Hart available
+// so maybe we can just assign the thread to the first available hart and be done with it
+void RevProc::AssignThread(const uint32_t& ThreadID){
+  bool ThreadAssigned = false;
+  // Check for available Hart 
+  for( auto Hart : Harts ){
```
leekillough, line 2460: For performance, it should be `auto& Hart : Harts`.

As it is written now, it causes the atomic reference count of each `std::shared_ptr` element to be incremented and decremented at each loop iteration.

include/RevProc.h
rkabrick: It needs to be mutable because as the RevCPU handles each thread in there based on their change of state it pops the thread from this queue.

include/RevProc.h
rkabrick: Nope, In the initial text for this PR I said anything I'd like you, Dave, and John to look at is marked `TODO: (BEFORE MERGE)` so that we talk about it before this PR is merged into `devel`

src/RevCPU.cc
rkabrick: All `output.verbose` lines are not debugging... If you ever see `std::cout` in a PR then definitely point that out. Anything using the SST::Output object is meant to be there. I'll remove the surrounding `if` comments though

src/RevCPU.cc
rkabrick: I definitely don't want a local copy. If I know the pointer isn't `const` is the single `&` sufficient? Or should I use the `&&`?

src/RevCPU.cc
```diff
@@ -2660,95 +2663,92 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
-        // This thread didn't experience a state change
-        continue;
+      else { // -- 3.
+        output.verbose(CALL_INFO, 8, 0, "Thread %" PRIu32 " on Core %" PRIu32 " was waiting on thread %u which has not yet completed so it remains blocked\n",
+                        Thread->GetThreadID(), ProcID, Thread->GetWaitingToJoinTID());
+        Thread->SetState(ThreadState::BLOCKED);
+        // -- 3a.
+        BlockedThreads.emplace(Thread->GetThreadID());
+
+        std::cout << Thread->to_string();
```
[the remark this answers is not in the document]
rkabrick: Nope! Good catch

src/RevProc.cc
rkabrick: Ah okay. Well that's good to know.

rkabrick: I only updated my `numHarts` one but a future PR can feel free to clean that up

src/RevProc.cc
rkabrick: Yeah, there's definitely a better way to do it. If I remember in your changes there was a scope that got muddied so you were seeing the message printed every time even if the hart didn't switch (or something similar) 

I'm also not crazy about the name because `GetHartID` implies you're getting the current Hart's ID when in reality you're getting the next Hart that is ready to decode. 

I think we can potentially modify the logic now that we have a RevHart object which can potentially allow to us to remove HART_CTE & HART_CTS as separate data structures in the future and perhaps cycle through each RevHart and check if its ready, then stop once you find one, but for now I'd say just deal with it.

src/RevProc.cc
rkabrick: The number of entries in the vector doesn't go to 0... My intention with this `any_of` is to check if any of the elements inside of the vector are non-zero. I'm guessing by your confusion that's not what I'm doing 😂

src/RevProc.cc
rkabrick: Yup. This was the fix for not clearing the scoreboard when there is a memreq in flight. Ugly as sin but with the way the LSQueue works I think this is the best way to do it. Open to ideas 🤷

src/RevProc.cc
rkabrick: Nope

src/RevProc.cc
rkabrick: Fixed!

rkabrick:
I am closing this PR as it is no longer passing multithreading tests. Will rollback some of my changes and open a new one

src/RevCPU.cc
leekillough: It was the comments I was referring to as "debugging code", like you were turning it always on for debugging.

src/RevCPU.cc
leekillough: If you know the RHS is an lvalue, or you want to enforce it as an lvalue, then `auto&` is correct. `auto&&` is only needed in variable contexts where the RHS can be either a lvalue or rvalue and might change. This lvalue/rvalue distinction is independent of `const`ness, since `auto` reference can deduce `const`.

For example, `auto&& [a, b, c]` is a structured binding with a universal reference which will work when the RHS has a tuple-like mixture of lvalues and rvalues.

src/RevProc.cc
leekillough: > The number of entries in the vector doesn't go to 0... My intention with this `any_of` is to check if any of the elements inside of the vector are non-zero. I'm guessing by your confusion that's not what I'm doing 😂

Correct, your predicate was always returning `true`, while it should have returned `x` (I think there is an `identity` function defined in one of the versions of C++, but I don't remember which one).

But switching to using `std::bitset` is simpler and faster.

PR #156, opened 2023-10-10
Rev Hart Rework

rkabrick:
This PR introduces two main things surrounding Harts in Rev:

## RevHart Class 
A `RevHart` currently has:
- `id`: Self-explanatory
- `EcallState`: Required to allow multithreading on a Proc otherwise two threads would trounce on each other's EcallState state if it was only 1 per Proc
- `AssignedThreadID`: This corresponds to the ThreadID that is assigned to the RevHart

## Configurable Number of Harts via Python
Specifying the new `numHarts` option inside of the rev python configuration will allow users to specify the number of harts PER core. The maximum number of harts supported is 65,536 which I *think* is reasonable. I can make it infinite but that prevents us from being able to use a `std::bitset` for the existing barrel logic inside of `RevProc` (ie. `HART_CTS` & `HART_CTE`). Until we make any fundamental changes to how that logic works I think enforcing a maximum number of Harts is sufficient. 

### RevProc
*RevProc* used to have a `std::vector<RevThread>& AssignedThreads` where the index into that vector represented the RevThread assigned to the Hart w/ ID = index. Now, RevProc has a `std::unordered_map<uint32_t ThreadID, std::shared_ptr<RevThread> AssignedThreads` which encompasses all threads assigned to that RevProc at any time. This is treated as the single source of truth for what threads are currently on a RevProc during execution. It is worth noting that while a `RevHart` may have an entry in its `AssignedThreadID` field, you have to use `bool RevProc::HartHasThread(HartID)` to check if the thread the Hart has assigned to it is actually still present inside the `AssignedThreads` map on that Proc. I'm not in love with this design but it seemed like the least problematic compared to having to iterate over every RevHart object and remove the matching shared_ptr<RevThread> every time a thread is removed from a Proc (ie. There existing a shared_ptr<RevThread> inside of `RevHart` and `AssignedThreads` on that Proc that may not reflect the same state at any one point). I am open to suggestions though.

*NewThreadInfo queue removed*:
Instead of having a separate data structure for new threads, the `ThreadsThatChangedState` vector is used for new threads that are created from `rev_pthread_create` and when RevCPU is 'handling' them, the new thread has a state of `START` which indicates this is a new thread

# Outstanding Issues 
Everything is passing with the exception of `fcvt` & `fmove` which both time out. I will investigate further however the rest of the code is in what I would consider to be a final state.


include/RevHart.h
```diff
@@ -0,0 +1,60 @@
...
+  ~RevHart(){}
+
+  ///< RevHart: Get the EcallState
+  EcallState& GetEcallState(){ return *Ecall; }
+
+  ///< RevHart: Get Hart's ID 
+  uint16_t GetID(){ return ID; }
+
+  ///< RevHart: Returns the ThreadID of the assigned thread
+  uint32_t GetAssignedThreadID(){ return AssignedThreadID; }
```
leekillough, line 49: `GetEcallState()`, `GetID()`, and `GetAssignedThreadID()` should be `const` (`const` after the closing `)` before the opening `{`). This allows them to be called on objects which are `const` (or references to them).

include/RevProc.h
```diff
@@ -156,6 +155,9 @@ class RevProc{
   ///< RevProc: Get pointer to Load / Store queue used to track memory operations
   std::shared_ptr<std::unordered_map<uint64_t, MemReq>> GetLSQueue(){ return LSQueue; }
 
+  // Find a Hart to assign ThreadID to
+  void AssignThread(const uint32_t& ThreadID);
```
leekillough, line 159: A reference is not required for a trivial type like `uint32_t`, and causes extra memory operations.

include/RevProc.h
```diff
@@ -167,23 +169,26 @@ class RevProc{
   uint64_t ExecPC;          ///< RevProc: executing PC
   uint16_t HartToDecode;    ///< RevProc: Current executing ThreadID
   uint16_t HartToExec;      ///< RevProc: Thread to dispatch instruction
+  std::vector<std::shared_ptr<RevHart>> Harts; ///< RevProc: vector of Harts
   uint64_t Retired;         ///< RevProc: number of retired instructions
 
+  unsigned numHarts;        ///< RevProc: Number of Harts for this core
```
leekillough, line 175: You used `uint16_t` elsewhere; best to be consistent; if you think it might change, then
```c++
   using numHarts_t = uint16_t;
```
can be defined in a header file, and then everywhere a `numHarts` appears, it should be declared `numHarts_t`.

(Note: `using type1 = type2` is preferred to `typedef type2 type1` in C++11 and later. It can also be templatized.)

include/RevProc.h
```diff
@@ -696,6 +674,15 @@ class RevProc{
     return isFloat ? regFile->FP_Scoreboard.any() : regFile->RV_Scoreboard.any();
```
leekillough, line 674: As I said, if you need to test for both float and general registers, you can remove `isFloat` and change this function to return `regFile->RV_Scoreboard.any() || regFile->FP_Scoreboard.any()`.  (I put RV first because it's more likely in practice to be true, so short-circuiting of `||` will be faster.)

include/RevProc.h
```diff
@@ -696,6 +674,15 @@ class RevProc{
...
   }
 
+  bool ThreadHasDependencies(const uint32_t& ThreadID) const {
+    // Make sure this ThreadID is in AssignedThreads
+    if (AssignedThreads.find(ThreadID) == AssignedThreads.end()) {
+      output->fatal(CALL_INFO, -1, "Error: Tried to check if thread %" PRIu32 " has dependencies but that thread was not found in AssignedThreads. This is a bug\n", ThreadID);
+    }
+    const RevRegFile* regFile = AssignedThreads.at(ThreadID)->GetRegFile();
+    return regFile->FP_Scoreboard.any() || regFile->RV_Scoreboard.any();
+  }
```
leekillough, line 684: See above. I'd switch order of `FP` and `RV`. This might make the above function redundant; please remove it if it's not needed. It's just another untested possible failure point if it's not used.

src/RevCPU.cc
```diff
@@ -2651,95 +2654,102 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
+        Thread->SetState(ThreadState::RUNNING);
       }
+      else { // -- 3.
+        output.verbose(CALL_INFO, 8, 0, "Thread %" PRIu32 " on Core %" PRIu32 " was waiting on thread %u which has not yet completed so it remains blocked\n",
+                        Thread->GetThreadID(), ProcID, Thread->GetWaitingToJoinTID());
+        Thread->SetState(ThreadState::BLOCKED);
+        // -- 3a.
+        BlockedThreads.emplace(Thread->GetThreadID());
+
+        std::cout << Thread->to_string();
```
leekillough, line 2698: Debug?

src/RevProc.cc
```diff
@@ -16,24 +16,25 @@ using MemSegment = RevMem::MemSegment;
...
                   SST::Output *Output )
   : Halted(false), Stalled(false), SingleStep(false),
     CrackFault(false), ALUFault(false), fault_width(0),
-    id(Id), HartToDecode(0), HartToExec(0), Retired(0x00ull),
-    opts(Opts), mem(Mem), loader(Loader), AssignedThreads(AssignedThreads),
-    GetNewThreadID(GetNewTID), output(Output), feature(nullptr),
-    PExec(nullptr), sfetch(nullptr) {
+    id(Id), HartToDecode(0), HartToExec(0), Retired(0x00ull), 
+    numHarts(NumHarts), opts(Opts), mem(Mem), loader(Loader), 
+    AssignedThreads(AssignedThreads), GetNewThreadID(GetNewTID),
```
leekillough, line 30: `GetNewThreadID(GetNewTID)` should be `GetNewThreadID(std::move(GetNewTID))`

This "moves" `GetNewTID`, which has already been constructed as a constructor parameter, into `GetNewThreadID`. If you don't use `std::move`, then the compiler has to create a new copy of `GetNewTID` in `GetNewThreadID`, allocating memory and copying data, and then it destroys `GetNewTID`. By using `std::move`, you are moving it rather than copying it, so it's much faster.

See: https://github.com/tactcomplabs/rev/pull/115 ("Exploit Move Semantics")

These kinds of "moves" instead of copies are not needed for native types (`int`, `double`, etc.), or for very small trivial classes like C-style structs. For large C-style structs, a `const <class>&` parameter should be used in the constructor, so that it gets passed by reference and doesn't get copied twice.

src/RevProc.cc
```diff
@@ -1564,6 +1566,31 @@ void RevProc::HandleALUFault(unsigned width){
...
 
+bool RevProc::ThreadCanBeRemoved(const uint32_t& ThreadID) {
+  auto* regFile = AssignedThreads.at(ThreadID)->GetRegFile();
+
+  // Check the if regfile scoreboard is anything but clear 
+  // if not, this thread has outstanding requests and can't be removed
+  // TODO: (PRE-MERGE) Do this in a more ~Lee~ fashion
+  if( feature->IsRV32() ){
+    for( unsigned i=0; i<_REV_NUM_REGS_; i++ ){
+      if( regFile->RV_Scoreboard[i] ){
```
leekillough, line 1577: Should be replaced with `regFile->RV_Scoreboard.any()`.

There's no need to differentiate with `feature->IsRV32()` anymore, since the scoreboards were merged.

Also, what about `regFile->FP_Scoreboard`?

src/RevProc.cc
```diff
@@ -1680,18 +1708,27 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
 
-  for( size_t HartID=0; HartID<AssignedThreads.size(); HartID++ ){
-    HART_CTS[HartID] = (AssignedThreads.at(HartID)->GetRegFile()->cost == 0);
+  for( size_t HartID=0; HartID<Harts.size(); HartID++ ){
+    if( HartHasThread(HartID) ){
+      HART_CTS[HartID] = (GetThreadOnHart(HartID)->GetRegFile()->cost == 0);
+    }
+    else {
+      HART_CTS[HartID] = false;
+    }
```
leekillough, line 1717: I'd just write the whole `if`-`else` statement as:
```c++
 HART_CTS[HartID] = HartHasThread(HartID) && GetThreadOnHart(HartID)->GetRegFile()->cost == 0;
```

src/RevCPU.cc
```diff
@@ -2651,95 +2654,102 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
+      ThreadQueue.emplace_back(Thread->GetThreadID());
+      break;
+
+    case ThreadState::RUNNING:
+      output.verbose(CALL_INFO, 11, 0, "Thread %" PRIu32 " on Core %" PRIu32 " is RUNNING\n", Thread->GetThreadID(), ProcID);
+      break;
+
+    case ThreadState::READY:
+      // If this happens we are not setting state when assigning thread somewhere
+      output.fatal(CALL_INFO, 99, "Error: Thread %" PRIu32 " on Core %" PRIu32 " is assigned but is in START state... This is a bug\n",
```
donofrio, line 2728: Should this read `START state..` or `READY state...`

include/RevHart.h
jleidel: +1

include/RevSysCalls.h
```diff
@@ -0,0 +1,55 @@
...
+    string.clear();
+    path_string.clear();
+    bytesRead = 0;
+    buf[0] = '\0';
+  }
+  EcallState() {
+    buf[0] = '\0';
+  }
+};
+
```
jleidel, line 51: this is a nit... remove unnecessary whitespace

src/RevCPU.cc
jleidel: as in debug state/mode? That might be handy for single stepping within the simulator

src/RevCPU.cc
```diff
@@ -2651,95 +2654,102 @@ void RevCPU::UpdateThreadAssignments(uint32_t ProcID){
...
-    output.verbose(CALL_INFO, 8, 0, "Core %" PRIu32 " has new threads\n", i);
-    // There are new thread(s) to create
-    for( size_t j=0; j<Procs[i]->GetNewThreadInfo().size(); j++ ){
-      auto NewThread = Procs[i]->NewThreadInfo.front();
-      Procs[i]->NewThreadInfo.pop();
-      InitThread(NewThread);
-    }
-  }
-}
+//void RevCPU::CheckForNewThreads(uint32_t i){
```
jleidel, line 2743: if this is commented out then we should remove it

include/RevProc.h
jleidel: +1, use the same type of the `Id` as well.  It's defined as an `unsigned` but printed as a `uint16_t`

src/RevProc.cc
```diff
@@ -46,14 +47,19 @@ RevProc::RevProc( unsigned Id,
...
+
   LSQueue = std::make_shared<std::unordered_map<uint64_t, MemReq>>();
   LSQueue->clear();
 
   featureUP = std::make_unique<RevFeature>(Machine, output, MinCost, MaxCost, Id);
   feature = featureUP.get();
   if( !feature )
     output->fatal(CALL_INFO, -1,
-                  "Error: failed to create the RevFeature object for core=%u\n", id);
+                  "Error: failed to create the RevFeature object for core=%" PRIu16 "\n", id);
```
jleidel, line 62: make sure you type use is consistent for `id`

src/RevCPU.cc
rkabrick: Ah yeah, that shouldn't be there.

src/RevProc.cc
rkabrick: Whoops forgot to remove this function. I replaced it with `ThreadHasDependencies`

src/RevCPU.cc
rkabrick: Sure should!

src/RevProc.cc
rkabrick: Well aren't you fancy

include/RevProc.h
rkabrick: Yup above function is removed now. I switched the order in this one

src/RevCPU.cc
leekillough: > as in debug state/mode? That might be handy for single stepping within the simulator

"Debug?" meant "Is this a debug print statement left over which shouldn't be merged?"

rkabrick:
Alright, all of your feedback has been addressed in the recent pushes. I still have to figure out what's happening with the `fcvt` and `fcmp` tests.

jleidel (approved):
good to merge

rkabrick:
Just spoke with @donofrio ... He was able to nail down the bug in the `fcvt` and `fcmp` tests and said he will push a change shortly. Once that's up, you're good to merge

PR #157, opened 2023-10-11
Pipeline scoreboard and stall mgmt from co-processor

donofrio:
**Summary**

- Adding new functionality to allow `RevCoProc` to Set / Clear scoreboard bits
- Adding new functionality to allow `RevCoProc` to stall the pipeline for a given Hart for an indefinite amount of time
- Both of the above functions are achieved with new public (but not that public) member functions on RevProc
- `RevCoProc` now has a pointer back to its associated `RevProc`
- Changing constructors for `RevProc` and `RevCoProc` to allow a future feature of one `RevProc` with many `RevCoProcs` to be implemented more easily

**Details**

The scoreboards are private members of `RevProc` and `RevCoProc` operates as a independent subcomponent and cannot directly access these fields. `Friends` was a cumbersome model and would have ended up requiring `RevCoProc` to have complete access to all `RevProcs` data members.  To get around this, I used the Passkey pattern. There is a new class (and header) `RevProcPasskey<T>`.  A specialized version of this class can _only_  be constructed by RevCoProc or its derived classes and is a parameter to the new public functions allowing scoreboard manipulation and pipeline stalling.

The constructor changes swap the relationship between CoProc and Proc. You now require a `RevProc` pointer when constructing a `RevCoProc` - which makes sense as you would never have a CoProc without a Proc.   `RevProc`  no longer takes a pointer to a single `RevCoProc` but is instead assigned through a Set function. This functionality could be extended to have `RevProc` maintain a `vector` of `RevCoProc` objects to support multiple attached co-processors.

### DIFF
--- a/include/RevCoProc.h
+++ b/include/RevCoProc.h
@@ -31,14 +31,16 @@
 #include "RevFeature.h"
 #include "RevMem.h"
 #include "RevInstTable.h"
+#include "RevProcPasskey.h"
+#include "RevProc.h"
 
 namespace SST::RevCPU{
-
+class RevProc;
 // ----------------------------------------
 // RevCoProc
 // ----------------------------------------
-struct RevCoProc : SST::SubComponent {
-
+class RevCoProc : public SST::SubComponent {
+public:
   SST_ELI_REGISTER_SUBCOMPONENT_API(SST::RevCPU::RevCoProc);
   SST_ELI_DOCUMENT_PARAMS({ "verbose", "Set the verbosity of output for the attached co-processor", "0" });
 
@@ -51,9 +53,17 @@ struct RevCoProc : SST::SubComponent {
   virtual bool Reset() = 0;                           /// RevCoProc: Reset - called on startup
   virtual bool Teardown() = 0;                        /// RevCoProc: Teardown - called when associated RevProc completes
   virtual bool ClockTick(SST::Cycle_t cycle) = 0;     /// RevCoProc: Clock - can be called by SST or by overriding RevCPU
+//  virtual void DependencySet(uint16_t HartID, uint16_t RegNum, bool isFloat, bool value = true); 
+//  virtual void DependencyClear(uint16_t HartID, uint16_t RegNum, bool isFloat){
+//    DependencySet(HartID, RegNum, isFloat, false);
+//  }
+
+  void SetParent(RevProc* parentPtr){parent = parentPtr;}
+  RevProcPasskey<RevCoProc> CreatePasskey(){return RevProcPasskey<RevCoProc>();}
 
 protected:
   SST::Output *output;                                ///< RevCoProc: sst output object
+  RevProc* parent;
 }; // class RevCoProc
 
 // ----------------------------------------
@@ -110,6 +120,7 @@ public:
   ///                   also signal to SST that the co-processor is done if ClockTick is registered
   ///                   to SSTCore vs. being driven by RevCPU
   virtual bool Teardown() { return Reset(); };
+
 
 private:
   struct RevCoProcInst {
--- a/include/RevCoProc.h
+++ b/include/RevCoProc.h
@@ -44,26 +44,24 @@ public:
   SST_ELI_REGISTER_SUBCOMPONENT_API(SST::RevCPU::RevCoProc);
   SST_ELI_DOCUMENT_PARAMS({ "verbose", "Set the verbosity of output for the attached co-processor", "0" });
 
-  /// RevCoProc: default constructor
+  /// RevCoProc: Constructor
   RevCoProc( ComponentId_t id, Params& params);
 
   /// RevCoProc: default destructor
-  virtual ~RevCoProc();                               /// RevCoProc: Destructor
+  virtual ~RevCoProc();                               ///< RevCoProc: Destructor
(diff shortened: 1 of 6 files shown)


include/RevCoProc.h
```diff
@@ -31,29 +31,35 @@
...
+class RevProc;
 // ----------------------------------------
 // RevCoProc
 // ----------------------------------------
-struct RevCoProc : SST::SubComponent {
-
-  SST_ELI_REGISTER_SUBCOMPONENT_API(SST::RevCPU::RevCoProc);
+class RevCoProc : public SST::SubComponent {
+public:
+  SST_ELI_REGISTER_SUBCOMPONENT_API(SST::RevCPU::RevCoProc, RevProc*);
```
jleidel, line 44: Would using the fully qualified namespace `SST::RevCPU::RevProc*` allow you to remove the forward decl of `class RevProc` from above?

include/RevProc.h
```diff
@@ -48,20 +48,22 @@
...
 #include "RevThread.h"
 #include "RevRand.h"
+#include "RevProcPasskey.h"
 #include "../common/syscalls/SysFlags.h"
 #include "../common/include/RevCommon.h"
 
 #define _PAN_FWARE_JUMP_            0x0000000000010000
 
 namespace SST::RevCPU{
+class RevCoProc;
```
jleidel, line 58: Is the forward decl necessary?

rkabrick:
So I think John's last comment is especially relevant because based on the Rev hierarchy, a RevProc is synonymous with a Core (ie. `numCores` == number of Procs). I'm not sure tying it to a specific "Core" makes sense from a typical design perspective. I've never heard of a core-based coprocessor but if the functionality for what we need it for requires that then that's fine. I'd just amend Lee's draft PR for changing `RevProc` to `RevCore` in the future to potentially name `RevCoProc` something different.

rkabrick:
I definitely think an approach that doesn't force RevThreads to execute on a certain Proc is closer to what an actual system would do. I think something that creates the CoProc interface for all Procs is the better approach.

include/RevProc.h
```diff
@@ -156,6 +158,37 @@ class RevProc{
...
+  void ExternalDepSet(RevProcPasskey<RevCoProc>, uint16_t HartID, uint16_t RegNum, bool isFloat, bool value = true){
+    DependencySet(HartID, RegNum, isFloat, value);
+  }
+
+  ///< RevProc: Allow a co-processor to manipulate the scoreboard by clearing a bit. Note the RevProcPassKey may only
+  ///  be created by a RevCoProc (or a class derived from RevCoProc) so this funciton may not be called from even within
+  ///  RevProc 
+  void ExternalDepClear(RevProcPasskey<RevCoProc>, uint16_t HartID, uint16_t RegNum, bool isFloat){
+    DependencyClear(HartID, RegNum, isFloat);
+  }
```
leekillough, line 177: Note that `DependencyClear` is written in terms of `DependencySet(..., value = true)` by calling `DependencySet(..., false)`, to avoid copy-and-paste code duplication.

include/RevProc.h
donofrio: Forward decl here is required for the RevProcPasskey<RevCoProc>

donofrio:
To address the comments from @jleidel and @rkabrick - this gets into creation of a heterogeneous CPU.  I view the hierarchy as: `RevCPU` --> `RevProc` --> `RevCoProc`.  The code as written does not allow a `RevCoProc` to exist without a `RevProc` just as a `RevProc` cannot exist without a `RevCPU`.  We could rename `RevCoProc` to `RevRoCC` as it functions very similar to that interface, but that may be a touch misleading as this is not an explicit RoCC interface per se.  We could create a generic `RevAccelerator` or something similar, but the connections between those and RevProc probably should be managed in a different way, likely through the python config. 

Having a fixed binding between `RevCoProc` and `RevProc` is not ideal, I think this is a broader topic of creating a heterogeneous `RevCPU` where some cores have different numbers and types of co-processors.  The code here preserves the previous 1:1 `RevProc` / `RevCoProc` relationship. I think we _do_ need better support for heterogeneity within a CPU, but that is a longer design discussion.

jleidel:
> To address the comments from @jleidel and @rkabrick - this gets into creation of a heterogeneous CPU. I view the hierarchy as: `RevCPU` --> `RevProc` --> `RevCoProc`. The code as written does not allow a `RevCoProc` to exist without a `RevProc` just as a `RevProc` cannot exist without a `RevCPU`. We could rename `RevCoProc` to `RevRoCC` as it functions very similar to that interface, but that may be a touch misleading as this is not an explicit RoCC interface per se. We could create a generic `RevAccelerator` or something similar, but the connections between those and RevProc probably should be managed in a different way, likely through the python config.
> 
> Having a fixed binding between `RevCoProc` and `RevProc` is not ideal, I think this is a broader topic of creating a heterogeneous `RevCPU` where some cores have different numbers and types of co-processors. The code here preserves the previous 1:1 `RevProc` / `RevCoProc` relationship. I think we _do_ need better support for heterogeneity within a CPU, but that is a longer design discussion.

ok, echo'ing back the implications in a different way.  That implies that a `RevCoProc` can only be utilized by one `RevProc` at a time.  I think this is ok in the short term for our immediate goals, but a `RevROCC` or `RevAccelerator` interface will be a big feature request in the future.

donofrio:
Yup - totally agree on that restriction, 1:1 relationship between CoProc and Proc and the  two components are tightly bound together.

jleidel (approved):
good to merge

PR #163, opened 2023-10-12
REV: Add clock_gettime syscall

ukasz:
Wire clock_gettime to the cycle counter to be able to use it to measure time intervals in various
segments of workloads. Since REV core is not aware of the clock it is running at we have to propagate timeConverter to each of the cores in RevProc, to
convert cycles to real time. Clockid argument is not supported - we simply ignore it.

Add a test with usage example of the "new" syscall.

### DIFF
--- a/src/RevCPU.cc
+++ b/src/RevCPU.cc
@@ -225,6 +225,10 @@ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
     }
   }
   #endif
+  // Setup timeConverter
+  for( size_t i=0; i<Procs.size(); i++){
+    Procs[i]->SetTimeConverter(timeConverter);
+  }
 
   // Initial thread setup
   uint32_t MainThreadID = id+1; // Prevents having MainThreadID == 0 which is reserved for INVALID
--- a/test/syscalls/clock_gettime/clock_gettime.c
+++ b/test/syscalls/clock_gettime/clock_gettime.c
@@ -1,0 +1,24 @@
+#include "../../../common/syscalls/syscalls.h"
+#define assert(x) if(!(x)) { asm(".byte 0; .byte 0; .byte 0; .byte 0"); }
+
+int main(int argc, char *argv[]) {
+  int sum = 0;
+  struct __kernel_timespec s, e;
+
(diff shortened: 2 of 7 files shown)


src/RevCPU.cc
```diff
@@ -272,6 +272,11 @@ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
     }
   }
 
+  // Setup timeConverter
+  for( unsigned i=0; i<Procs.size(); i++){
```
leekillough, line 276: `unsigned` should be `size_t`

ukasz: Fixed, thank you!

test/syscalls/clock_gettime/clock_gettime.c
```diff
@@ -0,0 +1,17 @@
+#include "../../../common/syscalls/syscalls.h"
+#define assert(x) if(!(x)) { asm(".byte 0; .byte 0; .byte 0; .byte 0"); }
+
+int main() {
+  struct __kernel_timespec s, e;
+
+  int ret = rev_clock_gettime(0, &s);
+  assert(ret == 0);
+
+  for(int i = 0; i < 100; i++);
```
leekillough, line 10: At `-O2` I'm sure the compiler can optimize away the loop, and even with `volatile`, it takes an infinitesimal amount of time. I'd normally use `nanosleep()`, but that's another system call which would need to be added.

ukasz: This is just for demonstrative purposes. Even if I throw it away it still will pass because an assert at the end contains =0 case. 
Do you want me to remove it from the test/example or add a comment?

leekillough: Maybe add `volatile` and a comment that this is for illustration purposes.

ukasz: I doubt volatile will prevent compiler from removing dead code, instead I used argc which is not known during compilation time and the worst case is that the loop will be changed to multiplication, which is fine I guess.

ukasz:
Resolved merge conflicts, no functional changes.

jleidel (approved):
good to merge

PR #164, opened 2023-10-12
Add lg() function for base-2 logarithm of integers

leekillough:
This adds a `lg()` function which returns the base-2 logarithm of integers, using the `__builtin_clz*` functions.

~~(Also increases `big_loop` test timeout to make CI tests pass.)~~

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -48,19 +48,16 @@ constexpr auto SignExt(T val, size_t bits){
 /// Base-2 logarithm of integers
 template<typename T>
 constexpr int lg(T x){
-  // We select the __builtin_clz which takes integers the same size as x
-  if(x){
-    if constexpr(sizeof(T) == sizeof(int)){
-      return 8*sizeof(T)-1 - __builtin_clz(x);
-    }else if constexpr(sizeof(T) == sizeof(long)){
-      return 8*sizeof(T)-1 - __builtin_clzl(x);
-    }else if constexpr(sizeof(T) == sizeof(long long)){
-      return 8*sizeof(T)-1 - __builtin_clzll(x);
-    }else{
-      static_assert(sizeof(T) == 0, "Unhandled type in lg()");
-    }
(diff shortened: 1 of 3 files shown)


rkabrick:
Timeout should not be increased for big_loop... I will take a look at how I can fix the slowdown introduced by my RevHart rework changes this weekend.

common/include/RevCommon.h
```diff
@@ -45,6 +45,21 @@ constexpr auto SignExt(T val, size_t bits){
...
+constexpr int lg(T x){
+  static_assert(std::is_integral_v<T>);
+
+  // We select the __builtin_clz which takes integers no smaller than x
+  if constexpr(sizeof(x) <= sizeof(int)){
+    return x ? 8*sizeof(int)-1 - __builtin_clz(x) : -1;
+  }else if constexpr(sizeof(x) <= sizeof(long)){
+    return x ? 8*sizeof(long)-1 - __builtin_clzl(x) : -1;
+  }else{
+    return x ? 8*sizeof(long long)-1 - __builtin_clzll(x) : -1;
```
donofrio, line 59: should the case of x == 0 throw rather than return -1? it's a sort-of-NaN.

leekillough: We don't want to return `0` because that coincides with the valid case of `lg(1) = 0`. Returning `-1` seems to be the best way of indicating error. We don't need to throw exceptions or call error routines -- this is supposed to be a simple helper function. Also it's `constexpr`, and by implication, `inline` (when I tried a simple test case, the constant was propagated and the function was fully evaluated at compile-time). The caller can check the argument and raise errors if the argument is invalid.

The x86 `bsf` and `bsr` instructions (as well as `__builtin_ctz()`) produce undefined behavior if the input is `0`, which is why we test for it.

Also see [Find first set](https://en.wikipedia.org/wiki/Find_first_set).

donofrio:
Good to merge

PR #138, opened 2023-09-29
Privatize and unionize RevRegFile

leekillough:
This privatizes the data members of `RevRegFile`, providing Getters/Setters for them. A few `friend` functions and classes are declared, to allow access to private data members directly.

`RevRegFile` is modified to use unnamed (anonymous) unions to make `RV32` and `RV64`, and other 32/64-bit duplicated registers, take up the same space, but still accessible by name in the `RevRegFile` class. This cuts the data size of `RevRegFile` by around 1/3.

PR #166, opened 2023-10-13
[RFC] Register tracing

ukasz:
For easier debugging on loglevel 9 and above we log RV32/64 register changes after each instruction executed. Additionaly on loglevel 10 we log the state
of every RV32/64 register.
To avoid slowing down execution too much we skip logging code by comparing loglevel early.

This is probably not the best way to implement that feature, but I wanted to start a discussion about possible options. Another way could be to move that to Ext->Execute, or even RegFile->SetX function if we can set up logging there. From the downsides current implementation does not cover syscalls.

### DIFF
--- a/src/RevProc.cc
+++ b/src/RevProc.cc
@@ -1682,6 +1682,10 @@ void RevProc::MarkLoadComplete(const MemReq& req){
 }
 
 bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
+  union {
+    uint32_t RV32b[_REV_NUM_REGS_];
+    uint64_t RV64b[_REV_NUM_REGS_];
+  };
   RevInst Inst;
 
   bool rtn = false;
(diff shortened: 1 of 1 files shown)


src/RevProc.cc
```diff
@@ -1682,6 +1682,8 @@ void RevProc::MarkLoadComplete(const MemReq& req){
 }
 
 bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
+  uint32_t RV32b[_REV_NUM_REGS_];
+  uint64_t RV64b[_REV_NUM_REGS_];
```
leekillough, line 1686: In `RevRegFile.h`, we used anonymous `union`s to join the `RV32` and `RV64` arrays.

Please see https://github.com/tactcomplabs/rev/pull/138

ukasz: Good point, changed. Thank you!

PR #167, opened 2023-10-17
Syscall fixes

rkabrick:
There were several bugs in the system call implementations. The `EcallLoadAndParseString` function was not properly accounting for the memory request being cleared, the write system call was consistently writing 8 bytes short because it did not use the updated templatized version of `ReadVal`. This fixes the `file-io` and `write` tests with mem hierarchy

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -81,7 +81,7 @@ enum class MemOp : uint8_t {
   MemOpAMO         = 9,
 };
 
-inline uint64_t make_lsq_hash(uint8_t destReg, RevRegClass regType, uint16_t HartID){
+inline uint64_t make_lsq_hash(uint16_t destReg, RevRegClass regType, unsigned HartID){
   return static_cast<uint64_t>(regType) << (16 + 8) | static_cast<uint64_t>(destReg) << 16 | HartID;
 };
 
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -51,7 +51,6 @@ EcallStatus RevProc::EcallLoadAndParseString(RevInst& inst,
                   req,
                   REVMEM_FLAGS(0));
       EcallState.bytesRead = 1;
-      std::cout << "Setting dependency for HartToExec == " << HartToExec << std::endl;
       DependencySet(HartToExec, 10, false);
       rtval = EcallStatus::CONTINUE;
     }
@@ -482,7 +481,7 @@ EcallStatus RevProc::ECALL_openat(RevInst& inst){
 
   // commented out to remove warnings
   // auto flags = RegFile->GetX<int>(RevReg::a2);
-  // auto mode = RegFile->GetX<int>(RevReg::a3);
+  auto mode = RegFile->GetX<int>(RevReg::a3);
 
   /*
    * NOTE: this is currently only opening files in the current directory
@@ -497,8 +496,8 @@ EcallStatus RevProc::ECALL_openat(RevInst& inst){
 
   auto action = [&]{
     // Do the openat on the host
-    dirfd = open(std::filesystem::current_path().c_str(), O_RDWR);
-    int fd = openat(dirfd, EcallState.string.c_str(), O_RDWR);
+    dirfd = open(std::filesystem::current_path().c_str(), mode);
+    int fd = openat(dirfd, EcallState.string.c_str(), mode);
 
     // Add the file descriptor to this thread
     Thread->AddFD(fd);
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -616,7 +616,6 @@ EcallStatus RevProc::ECALL_write(RevInst& inst){
   auto nleft = nbytes - ECALL.string.size();
   if(nleft == 0 && LSQueue->count(lsq_hash) == 0){
     int rc = write(fd, ECALL.string.data(), ECALL.string.size());
-    fsync(fd);
     RegFile->SetX(RevReg::a0, rc);
     ECALL.clear();
     DependencyClear(HartToExec, 10, false);
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -476,6 +476,7 @@ EcallStatus RevProc::ECALL_fchown(RevInst& inst){
 
 // 56, rev_openat(int dfd, const char  *filename, int flags, umode_t mode)
 EcallStatus RevProc::ECALL_openat(RevInst& inst){
+  output->verbose(CALL_INFO, 2, 0, "ECALL: openat called by thread %" PRIu32 " on hart %" PRIu32 "\n", GetActiveThreadID(), HartToExec);
   auto dirfd = RegFile->GetX<int>(RevReg::a0);
   auto pathname = RegFile->GetX<uint64_t>(RevReg::a1);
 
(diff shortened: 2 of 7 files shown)


src/RevSysCalls.cc
```diff
@@ -16,41 +16,45 @@ EcallStatus RevProc::EcallLoadAndParseString(RevInst& inst,
...
+      //We are in the middle of the string - read one byte
+      MemReq req{straddr + EcallState.string.size(), 10, RevRegClass::RegGPR, HartToExec, MemOp::MemOpREAD, true, [=](const MemReq& req){this->MarkLoadComplete(req);}};
+      LSQueue->insert({make_lsq_hash(req.DestReg, req.RegType, req.Hart), req});
+      mem->ReadVal(HartToExec,
+                  straddr + EcallState.string.size(),
+                  EcallState.buf.data(),
+                  req,
+                  REVMEM_FLAGS(0));
+      EcallState.bytesRead = 1;
+      std::cout << "Setting dependency for HartToExec == " << HartToExec << std::endl;
```
donofrio, line 54: Spurious cout

src/RevSysCalls.cc
```diff
@@ -486,25 +491,31 @@ EcallStatus RevProc::ECALL_openat(RevInst& inst){
...
+  auto& Thread = GetThreadOnHart(HartToExec);
+
+  auto& EcallState = Harts.at(HartToExec)->GetEcallState();
 
   auto action = [&]{
     // Do the openat on the host
-    dirfd = open(std::filesystem::current_path().c_str(), O_RDONLY);
-    int fd = openat(dirfd, Harts.at(HartToExec)->GetEcallState().string.c_str(), O_RDWR);
+    dirfd = open(std::filesystem::current_path().c_str(), O_RDWR);
+    int fd = openat(dirfd, EcallState.string.c_str(), O_RDWR);
```
donofrio, line 502: Do we ignore any other `openat` syscall parameters and always open a file RDWR?

common/include/RevCommon.h
```diff
@@ -81,7 +81,7 @@ enum class MemOp : uint8_t {
   MemOpAMO         = 9,
 };
 
-inline uint64_t make_lsq_hash(uint8_t destReg, RevRegClass regType, uint16_t HartID){
+inline uint64_t make_lsq_hash(uint16_t destReg, RevRegClass regType, unsigned HartID){
```
leekillough, line 84: `HartID` is `unsigned` in some places and `uint16_t` in others, maybe others. 

I think a `HartID_t` typedef should be defined and used everywhere.

rkabrick: HartID should always be unsigned

src/RevSysCalls.cc
```diff
@@ -595,65 +598,55 @@ EcallStatus RevProc::ECALL_read(RevInst& inst){
...
+  }
+
+  if (LSQueue->count(lsq_hash) == 0) {
+    MemReq req (addr + ECALL.string.size(), 10, RevRegClass::RegGPR, HartToExec, MemOp::MemOpREAD, true, RegFile->GetMarkLoadComplete());
+    LSQueue->insert({lsq_hash, req});
+
+    // Reverted to original approach for type-specific reads
     if(nleft >= 8){
-      readfunc(reinterpret_cast<uint64_t*>(ECALL.buf.data()));
+      mem->ReadVal<uint64_t>(HartToExec, addr+ECALL.string.size(), reinterpret_cast<uint64_t*>(ECALL.buf.data()), req, REVMEM_FLAGS(0));
```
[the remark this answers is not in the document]
leekillough: But the `LSQueue` stuff might have needed to be moved outside of `readfunc` and called unconditionally regardless of size.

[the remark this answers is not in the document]
rkabrick: Oh? I wonder why it was never reading the correct amount then. Is there anything wrong with the implementation as it is now?

src/RevSysCalls.cc
rkabrick: That's how I did it originally way back when but I suppose there's no reason to keep it that way

src/RevSysCalls.cc
```diff
@@ -595,65 +598,55 @@ EcallStatus RevProc::ECALL_read(RevInst& inst){
...
+      ECALL.bytesRead = 8;
     } else if(nleft >= 4){
-      readfunc(reinterpret_cast<uint32_t*>(ECALL.buf.data()));
+      mem->ReadVal<uint32_t>(HartToExec, addr+ECALL.string.size(), reinterpret_cast<uint32_t*>(ECALL.buf.data()), req, REVMEM_FLAGS(0));
```
leekillough, line 634: The explicit template argument `ReadVal<uint32_t>` is unnecessary. You can call `ReadVal()` without the template argument. It is automatically deduced from `reinterpret_cast<uint32_t*>(ECALL.buf.data())`. Same for all of the other cases.

leekillough:
```c++
  template <typename T>
  bool ReadVal( unsigned Hart, uint64_t Addr, T *Target,
                const MemReq& req,
                StandardMem::Request::flags_t flags){
    return ReadMem(Hart, Addr, sizeof(T), Target, req, flags);
  }
```
The third `ReadVal` parameter is `T *Target`. Passing `reintrepret_cast<U*>(ECALL.buf.data())` as its third argument causes `T` to be automatically deduced to `U`. You do not need to pass `T` explicitly.

rkabrick: Gotcha... I removed it

donofrio (approved):
Good to merge on my end

rkabrick:
rebased onto current devel

src/RevSysCalls.cc
```diff
@@ -16,41 +16,44 @@ EcallStatus RevProc::EcallLoadAndParseString(RevInst& inst,
...
+      // from the caller, such as performing a syscall using EcallState.string.
+      action();
+
+      EcallState.string.clear();   //reset the ECALL buffers
+      EcallState.bytesRead = 0;
+
+      DependencyClear(HartToExec, 10, false);
+      rtval = EcallStatus::SUCCESS;
+    }else{
+      //We are in the middle of the string - read one byte
```
jleidel, line 45: you guys really need to start paying attention to the 80 character limits.  Code like this is illegible on anything except VSCode with a 100" monitor

jleidel (approved):
good to merge

src/RevSysCalls.cc
leekillough: I generally try to keep lines under ~80-100 characters.

One tip: if a line is too long because it has a long string, the string can be split into adjacent strings, like:
```c++
"very long strings "
"can be split across lines"
```
I usually put the separating space at the ends of the split string(s).

I could make a pass at re-wrapping the lines intelligently with `emacs`, or more fancy with `clang-format`, but it would be disruptive and cause merge conflicts with orhers' work in progress, so it would need to done at a strategic time, such as after a deliverable.

In rocBLAS we used a pre-commit script which called `clang-format` on the committed file, and rejected the commit if `clang-format` produced differences. I could develop a similar script for Rev. It would incrementally reformat the source files at each commit, rather than reformatting the whole source tree. I will create a `clang-format` file which keeps most of the style used today.

rkabrick: I use neovim and have a 27" monitor 😭... I'll go ahead and make the lines 80 wide

rkabrick: Alright all the code lines are now no longer than 80-90 chars long

PR #168, opened 2023-10-17
tracer pr

kpgriesser:
This merge enables the compact execution tracer which optionally integrations with the spike disassembler library.
See Documentation/Tracer.md for details.

### DIFF
--- a/include/RevTracer.h
+++ b/include/RevTracer.h
@@ -142,7 +142,7 @@ namespace SST::RevCPU{
       /// RevTracer: assign disassembler. Returns 0 if successful
       int SetDisassembler(std::string machine);
       /// RevTracer: assign trace symbol lookup map
-      int SetTraceSymbols(std::map<uint64_t,std::string>* TraceSymbols);
+      void SetTraceSymbols(std::map<uint64_t,std::string>* TraceSymbols);
       /// RevTracer: assign cycle where trace will start (user param)
       void SetStartCycle(uint64_t c);
       /// RevTracer: assign maximum output lines (user param)
@@ -182,19 +182,19 @@ namespace SST::RevCPU{
       /// RevTracer: pointer to output stream
       SST::Output *pOutput;
       /// RevTracer: control whether output is printed or not ( sampling continues )
-      bool outputEnabled;
+      bool outputEnabled = false;
       /// RevTracer: Special affecting trace output
       TraceEvents_t events;
       /// RevTracer: buffer for captured states
       std::vector<TraceRec_t> traceRecs;
       /// RevTracer: saved program counter
-      uint64_t pc;
+      uint64_t pc = 0;
       /// RevTracer: previous program counter for branch determination
-      uint64_t lastPC; 
+      uint64_t lastPC = 0; 
       /// RevTracer: saved instruction
-      uint32_t insn;
+      uint32_t insn = 0;
       /// RevTracer: map of instruction addresses to symbols
-      std::map<uint64_t,std::string>* traceSymbols;
+      std::map<uint64_t,std::string>* traceSymbols = nullptr;
       /// RevTracer: Array of supported "NOP" instructions avaible for trace controls
       uint32_t nops[NOP_COUNT];
       /// RevTracer: Check current state against user settings and update state
@@ -206,11 +206,11 @@ namespace SST::RevCPU{
       /// RevTracer: Format data associated with memory access
       void fmt_data(unsigned len, uint64_t data, std::stringstream& s);
       /// RevTracer: Generate string from captured state
-      std::string RenderOneLiner(std::string& fallbackMnemonic);
+      std::string RenderOneLiner(const std::string& fallbackMnemonic);
       /// RevTracer: User setting: starting cycle of trace (overrides programmtic control)
-      uint64_t startCycle;
+      uint64_t startCycle = 0;
       /// RevTracer: User setting: maximum number of lines to print
-      uint64_t cycleLimit;
+      uint64_t cycleLimit = 0;
       /// RevTracer: support for trace control push/pop 
       std::vector<bool> enableQ;
       /// RevTracer: current pointer into trace controls queue
@@ -218,9 +218,9 @@ namespace SST::RevCPU{
       /// RevTracer: wraparound limit for trace controls queue
       const unsigned MAX_ENABLE_Q = 100;
       /// RevTracer: count of lines rendered
-      uint64_t traceCycles;
+      uint64_t traceCycles = 0;
       /// RevTracer: Hard disable for output
-      bool disabled;
+      bool disabled = 0;
 
     }; // class RevTracer
 
--- a/src/RevTracer.cc
+++ b/src/RevTracer.cc
@@ -18,10 +18,7 @@
 
 using namespace SST::RevCPU;
 
-RevTracer::RevTracer(std::string Name, SST::Output *o)
-    : name(Name), pOutput(o), outputEnabled(false), lastPC(0), insn(0), 
-    traceSymbols(nullptr), startCycle(0), cycleLimit(0), traceCycles(0), 
-    disabled(0) {
+RevTracer::RevTracer(std::string Name, SST::Output *o): name(Name), pOutput(o) {
     
     enableQ.resize(MAX_ENABLE_Q);
     enableQ.assign(MAX_ENABLE_Q,0);
@@ -70,10 +67,9 @@ int RevTracer::SetDisassembler(std::string machine)
     #endif
 }
 
-int RevTracer::SetTraceSymbols(std::map<uint64_t, std::string> *TraceSymbols)
+void RevTracer::SetTraceSymbols(std::map<uint64_t, std::string> *TraceSymbols)
 {
     traceSymbols = TraceSymbols;
-    return 0;
 }
 
 void RevTracer::SetStartCycle(uint64_t c)
@@ -105,7 +101,7 @@ void RevTracer::CheckUserControls(uint64_t cycle)
 {
     // bail out early if disabled
     if (disabled) return;
-    if (cycleLimit and traceCycles>cycleLimit) {
+    if (cycleLimit && traceCycles>cycleLimit) {
         disabled = true;
         if (outputEnabled) {
             outputEnabled = false;
@@ -116,7 +112,7 @@ void RevTracer::CheckUserControls(uint64_t cycle)
     
     // Using a startCycle will override programmatic controls.
     if (startCycle>0) {
-        bool enable = startCycle and cycle > startCycle;
+        bool enable = startCycle && cycle > startCycle;
         if (enable != outputEnabled) {
             outputEnabled = enable;
             events.f.trc_ctl = 1;
@@ -157,7 +153,7 @@ void RevTracer::SetFetchedInsn(uint64_t _pc, uint32_t _insn)
 
 bool RevTracer::OutputOK()
 {
-    return outputEnabled or events.f.trc_ctl;
+    return outputEnabled || events.f.trc_ctl;
 }
 
 void RevTracer::regRead(size_t r, uint64_t v)
@@ -176,9 +172,7 @@ void RevTracer::memWrite(uint64_t adr, size_t len,  const void *data)
     uint64_t d = *((uint64_t*) data);
     if (len<8) {
         // zero out garbage bytes
-        unsigned shift = (8-len)*8;
-        uint64_t mask = (~0ULL) >> shift;
-        d = d & mask;
+        d &= ~(~0llu << 8*len);
     }
     traceRecs.emplace_back(TraceRec_t(MemStore,adr,len,d));
 }
@@ -210,10 +204,10 @@ void RevTracer::InstTrace(size_t cycle, unsigned id, unsigned hart, unsigned tid
     Reset();
 }
 
-std::string RevTracer::RenderOneLiner(std::string& fallbackMnemonic)
+std::string RevTracer::RenderOneLiner(const std::string& fallbackMnemonic)
 {
     // Flow Control Events
-    std::stringstream ss_events;
+    std::stringstream ss_events; 
     if (events.v) {
         if (events.f.trc_ctl) {
             EVENT_SYMBOL e = outputEnabled ? EVENT_SYMBOL::TRACE_ON : EVENT_SYMBOL::TRACE_OFF;


documentation/Tracer.md
```diff
@@ -0,0 +1,201 @@
...
+4.4 Register to Memory (Store)
+------------------------------
+  |  PC  : INSN    |     Disassembly    |
+  0x10220:fef42623  sw      a5, -20(s0)	 
+
+  |                    Effects                        |
+    0x3ffffb60<-s0 0x64<-a5 [0x3ffffb4c,4]<-0x00000064
+
+  Effects interpretation:
+   data<-reg  : Source register read
```
ukasz, line 156: This format is a bit confusing, in my opinion "reg<-data" would be better, or at least "data->reg"

test/tracer/tracer.c
```diff
@@ -0,0 +1,101 @@
...
+ * Copyright (C) 2017-2023 Tactical Computing Laboratories, LLC
+ * All Rights Reserved
+ * contact@tactcomplabs.com
+ *
+ * See LICENSE in the top level directory for licensing details
+ *
+ */
+
+#include <stdlib.h>
+#include <cstdint>
```
ukasz, line 15: For me clang complains about this include, I had to change it to stdint.h

documentation/Tracer.md
rkabrick: Perhaps a prefix indicator would be better? 

R(reg): data
W(data): reg 

At the very least I definitely agree with differentiating the direction of the arrows if we don't go with a prefix.

include/RevInstHelpers.h
```diff
@@ -76,15 +76,16 @@ bool load(RevFeature *F, RevRegFile *R, RevMem *M, RevInst Inst) {
...
+    req.Set(rs1 + Inst.ImmSignExt(12),
             Inst.rd, RevRegClass::RegGPR,
             F->GetHartToExec(),
             MemOp::MemOpREAD,
             true,
             R->GetMarkLoadComplete());
     R->LSQueueInsert({make_lsq_hash(Inst.rd, RevRegClass::RegGPR, F->GetHartToExec()), req});
     M->ReadVal(F->GetHartToExec(),
-               R->GetX<uint64_t>(Inst.rs1) + Inst.ImmSignExt(12),
+               rs1 + Inst.ImmSignExt(12),
```
rkabrick, line 88: What's the reason for this change? @leekillough can you confirm if this is okay or should be something else

rkabrick: Oh wait I'm dumb. This is the same thing broken up into two stages so the tracer can use the value (I think)

src/RevTracer.cc
```diff
@@ -0,0 +1,337 @@
...
+
+#include <sstream>
+#include <iomanip>
+#include <string>
+
+#include "../include/RevTracer.h"
+#include "RevTracer.h"
+
+using namespace SST::RevCPU;
+using namespace std;
```
rkabrick, line 20: Is this necessary? Seems like you prefix most of what you use from std namespace with `std::` anyways

src/RevTracer.cc
```diff
@@ -0,0 +1,337 @@
...
+    if (diasm) delete diasm;
+    if (isaParser) delete isaParser;
+    #endif
+}
+
+int SST::RevCPU::RevTracer::SetDisassembler(std::string machine)
+{
+    #ifdef REV_USE_SPIKE
+    try {
+        // TODO privelege level options
```
rkabrick, line 62: Is this TODO outstanding?

src/RevTracer.cc
```diff
@@ -0,0 +1,337 @@
...
+{
+    return outputEnabled or events.f.trc_ctl;
+}
+
+void SST::RevCPU::RevTracer::regRead(uint8_t r, uint64_t v)
+{
+    traceRecs.emplace_back(TraceRec_t(RegRead,r,v));
+}
+
+void SST::RevCPU::RevTracer::regWrite(uint8_t r, uint64_t v)
```
rkabrick, line 169: Perhaps change these register arguments to use the new `RevReg` enum class (ie. `RevReg::a0`)

src/RevTracer.cc
```diff
@@ -0,0 +1,337 @@
...
+{
+    traceRecs.emplace_back(TraceRec_t(RegWrite,r,v));
+}
+
+void SST::RevCPU::RevTracer::memWrite(uint64_t adr, size_t len,  const void *data)
```
rkabrick, line 174: I think your namespace should be `SST::RevCPU` already so the prefix should be unecessary

rkabrick (changes requested):
Everything looks good to me aside from the few comments I left. Assuming those get addressed I'm good to get this merged in.

documentation/Tracer.md
leekillough: I agree that arrows should go right to left, but I think the LHS of an arrow should always be a register or memory, and not the value being written.

For the value being written,
```
Reg <- Reg (= 0x12345678)
Mem <- Reg (= 0x12345678)
Reg <- 0x12345678
```
should be used, similar to how `gdb` prints values when examining variables / registers.

include/RevRegFile.h
```diff
@@ -214,21 +218,28 @@ class RevRegFile {
...
     if( IsRV32 ){
-      RV32[size_t(rd)] = RevReg(rd) != RevReg::zero ? uint32_t(val) : 0;
+      res = RevReg(rd) != RevReg::zero ? uint32_t(val) : 0;
+      RV32[size_t(rd)] = res;
     }else{
-      RV64[size_t(rd)] = RevReg(rd) != RevReg::zero ? uint64_t(val) : 0;
+      res = RevReg(rd) != RevReg::zero ? uint64_t(val) : 0;
+      RV64[size_t(rd)] = res;
     }
+    TRACE_REG_WRITE(rd, res);
```
leekillough, line 242: Should the trace be in the type of the source value (`T`), or should it be in the type of the destination register (i.e., split into the two `if` conditions above)?

include/RevRegFile.h
```diff
@@ -245,8 +256,10 @@ class RevRegFile {
   void SetPC(T val) {
     if( IsRV32 ){
       RV32_PC = static_cast<uint32_t>(val);
+      TRACE_PC_WRITE(RV32_PC);
     }else{
       RV64_PC = static_cast<uint64_t>(val);
+      TRACE_PC_WRITE(RV64_PC);
```
leekillough, line 262: See above. Here it's split between the destination register types. I think I favor this approach.

include/RevRegFile.h
```diff
@@ -255,8 +268,10 @@ class RevRegFile {
   void AdvancePC(T bytes) {
     if ( IsRV32 ) {
       RV32_PC += bytes;
+      TRACE_PC_WRITE(RV32_PC);
     }else{
       RV64_PC += bytes;
+      TRACE_PC_WRITE(RV64_PC);
```
leekillough, line 274: If you want to avoid unnecessary tracer output advancing `PC` at the end of every instruction, we can refactor this so that `AdvancePC` only advances one instruction and avoids excessive tracer output, and use `SetPC` for all other `PC` modifications.

include/RevTracer.h
```diff
@@ -0,0 +1,229 @@
...
+#else
+#define TRACE_REG_READ(R,V)
+#define TRACE_REG_WRITE(R,V)
+#define TRACE_PC_WRITE(PC)
+#define TRACE_MEM_WRITE(ADR, LEN, DATA)
+#define TRACE_MEM_READ(ADR, LEN, DATA)
+#endif
+
+namespace SST{
+  namespace RevCPU{
```
leekillough, line 56: `namespace SST::RevCPU{` shorthand can be used instead of nested namespaces in C++17 and later.

include/RevTracer.h
```diff
@@ -0,0 +1,229 @@
...
+  // SRLI   rd=x0 2^10   srli  x0, x0, ?    0x00?05013  //
+  // SRAI   rd=x0 2^10   srai  x0, x0, ?    0x40?05013  //
+  // SLT    rd=x0 2^10   slt   x0, x0, x?   0x00?02033  // Not supported
+  // SLTU   rd=x0 2^10   sltu  x0, x0, x?   0x00?03033  // Not supported
+  const std::map<std::string, uint32_t> s2op {
+    {"slti",  0x00002013}, 
+    {"sltiu", 0x00003013},
+    {"slli",  0x00001013},
+    {"srli",  0x00005013},
+    {"srai",  0x40005013}
```
leekillough, line 73: A trailing comma on the last entry is allowed and encouraged, to ease adding new entries or shuffling existing entries, or for automatic code generation.

include/RevTracer.h
```diff
@@ -0,0 +1,229 @@
...
+  const std::string TRC_OP_DEFAULT = "slli";
+  const int TRC_OP_POS = 20;
+
+  // Position of fully formed instruction in 'nops' array
+  enum TRC_CMD_IDX : unsigned {
+    TRACE_OFF = 0,
+    TRACE_ON = 1,
+    TRACE_PUSH_OFF = 2,
+    TRACE_PUSH_ON = 3,
+    TRACE_POP = 4
```
leekillough, line 84: See trailing comma comment above

include/RevTracer.h
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  constexpr unsigned NOP_COUNT = TRC_CMD_IDX::TRACE_POP + 1;
+  
+  enum class EVENT_SYMBOL : unsigned {
+    OK = 0x0,
+    STALL = 0x1,
+    FLUSH = 0x2,
+    TRACE_ON = 0x100,
+    TRACE_OFF = 0x102,
```
leekillough, line 94: Like this.

include/RevTracer.h
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  const std::map<EVENT_SYMBOL,char> event2char {
+    {EVENT_SYMBOL::OK,' '},
+    {EVENT_SYMBOL::STALL,'#'},
+    {EVENT_SYMBOL::FLUSH,'!'},
+    {EVENT_SYMBOL::TRACE_ON,'+'},
+    {EVENT_SYMBOL::TRACE_OFF,'-'}
+  };
```
leekillough, line 103: See above

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+#include <iomanip>
+#include <string>
+
+#include "../include/RevTracer.h"
+#include "RevTracer.h"
+
+using namespace SST::RevCPU;
+
+RevTracer::RevTracer(std::string Name, SST::Output *o)
+    : name(Name), pOutput(o), outputEnabled(false), lastPC(0), insn(0),
```
leekillough, line 22: Optimization tip:  `name(std::move(Name))` makes it a shallow instead of deep copy.

Also, constructor parameter names don't have to be different from the member names they initialize. `name(std::move(name))` is perfectly fine; the initialized member `name` is in class scope; the initializer expression `std::move(name)`'s `name` is in block (function) scope.

documentation/Tracer.md
donofrio: I like the <- notation as is. It is overloaded, but conveys a lot of information succinctly. Keeping the arrows in one direction allows the focus to be on the values next to the arrows.

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+
+#include "../include/RevTracer.h"
+#include "RevTracer.h"
+
+using namespace SST::RevCPU;
+
+RevTracer::RevTracer(std::string Name, SST::Output *o)
+    : name(Name), pOutput(o), outputEnabled(false), lastPC(0), insn(0), 
+    traceSymbols(nullptr), startCycle(0), cycleLimit(0), traceCycles(0), 
+    disabled(0) {
```
leekillough, line 24: Also generally, using default member initializers is clearer and less error-prone than listing out all constant initializations in a constructor. (Fill in types and put this in `RevTracer.h`; then the constructor(s) don't need to initialize these members.)

```c++
    traceSymbols = nullptr;
    startCycle = 0;
    cycleLimit = 0;
    traceCycles = 0;
    disabled = 0;
```

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+    #else
+    return 1;
+    #endif
+}
+
+int RevTracer::SetTraceSymbols(std::map<uint64_t, std::string> *TraceSymbols)
+{
+    traceSymbols = TraceSymbols;
+    return 0;
+}
```
leekillough, line 77: Unless this is a part of a family of functions which return `int`, the return type should be `void`

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+    unsigned cmd_template = s2op.at(cmd);
+    for (unsigned i=0;i<NOP_COUNT;i++)
+         nops[i]= cmd_template | (i << TRC_OP_POS);
+}
+
+void RevTracer::CheckUserControls(uint64_t cycle)
+{
+    // bail out early if disabled
+    if (disabled) return;
+    if (cycleLimit and traceCycles>cycleLimit) {
```
leekillough, line 108: `&&`

You don't see `and` too often

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+        if (outputEnabled) {
+            outputEnabled = false;
+            events.f.trc_ctl = 1;
+        }
+        return;
+    }
+    
+    // Using a startCycle will override programmatic controls.
+    if (startCycle>0) {
+        bool enable = startCycle and cycle > startCycle;
```
leekillough, line 119: `&&`

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+
+void RevTracer::SetFetchedInsn(uint64_t _pc, uint32_t _insn)
+{
+    insn = _insn;
+    pc = _pc;
+}
+
+bool RevTracer::OutputOK()
+{
+    return outputEnabled or events.f.trc_ctl;
```
leekillough, line 160: `||`

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+
+void RevTracer::regWrite(uint8_t r, uint64_t v)
+{
+    traceRecs.emplace_back(TraceRec_t(RegWrite,r,v));
+}
+
+void RevTracer::memWrite(uint64_t adr, size_t len,  const void *data)
+{
+    // Only tracing the first 64 bytes. Retaining pointer in case we change that.
+    uint64_t d = *((uint64_t*) data);
```
leekillough, line 176: Might return bus error if not aligned properly. Pointer type punning is generally undefined behavior in C/C++ unless the finally dereferenced pointer and the data at the address are the same type (with exceptions for pointer to `char`).

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+    if (len<8) {
+        // zero out garbage bytes
+        unsigned shift = (8-len)*8;
+        uint64_t mask = (~0ULL) >> shift;
+        d = d & mask;
```
leekillough, line 181: `d &= ~(~0llu << 8*len);`

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+    if (OutputOK()){
+        pOutput->verbose(CALL_INFO, 5, 0,
+                         "Core %" PRIu32 "; Hart %" PRIu32 "; Thread %" PRIu32 "]; *I %s\n",
+                         id, hart, tid, RenderOneLiner(fallbackMnemonic).c_str());
+    }
+    Reset();
+}
+
+std::string RevTracer::RenderOneLiner(std::string& fallbackMnemonic)
+{
```
leekillough, line 209: TODO: I need to review this more thoroughly.

test/tracer/gold.trc
```diff
@@ -0,0 +1,488 @@
+rm -Rf tracer.exe tracer.dis
```
leekillough, line 1: Is this file intended to be checked in?

lt appears to be a log of commands on MacOS.

If it's expected trace output, it will need to be different under Linux.

test/tracer/tracer.c
```diff
@@ -0,0 +1,101 @@
...
+#define assert TRACE_ASSERT
+
+// inefficient calculation of r-s
+int long_sub(int r, int s) {
+  for (int i=0;i<s;i++)
+    r--;
+  return r;
+}
+
+volatile int check_push_on(int x, int y) {
```
leekillough, line 27: `volatile` is not meaningful on function return values (but it could be important on the referents of references, or the pointees of pointers, of function return references or pointers, e.g., `volatile int& check_push_on(...)`

src/RevTracer.cc
leekillough: `SetX` etc. were changed so that a general template argument of either `enum` or integer could be passed, and it cast it to `size_t` to get the raw register number, whether it was passed as an `enum` or as an integer. As long as it's legally cast to `size_t()`, it can be used. Otherwise you're likely to get compilation errors.

This was the simplest way, but if you want to get more stringent, template metaprogramming can be used.

src/RevTracer.cc
leekillough: **NEVER** use `using namespace std;`!!!

test/tracer/tracer.c
leekillough: If lang==C, it's `<stdwhatever.h>`

If lang==C++, it's `<cstdwhatever>` (without `.h`)

.gitignore
```diff
@@ -4,6 +4,7 @@
 *.csv
 .cache/clangd/index
 .vscode
+*.dylib
```
leekillough, line 7: I thought we already had that in `devel`. I wonder why it keeps disappearing.

jleidel (approved):
good to merge

src/RevTracer.cc
leekillough: Okay, but keep in mind the `RevReg enum` is a scoped enum, meaning it is not implicitly convertible to an integer type like `size_t`. It must be explicitly cast (whether with `static_cast`, C-style cast, or function-style cast).

Using scoped enums (`enum class` instead of plain `enum`) is preferred in new code unless their own scope is limited (like inside a single function or class) _and_ using the scope specifier like `RevReg::a0`  is deemed too verbose.

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+    CheckUserControls(cycle);
+    if (OutputOK()){
+        pOutput->verbose(CALL_INFO, 5, 0,
+                         "Core %" PRIu32 "; Hart %" PRIu32 "; Thread %" PRIu32 "]; *I %s\n",
+                         id, hart, tid, RenderOneLiner(fallbackMnemonic).c_str());
+    }
+    Reset();
+}
+
+std::string RevTracer::RenderOneLiner(std::string& fallbackMnemonic)
```
leekillough, line 209: Can this be `const std::string& fallbackMnemonic`?

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+void RevTracer::Reset()
+{
+    events.v = 0;
+    // save processing time and only clear the essentials.
+    // pc = 0;
+    // insn = 0;
+    traceRecs.clear();
+}
+
+void RevTracer::fmt_reg(uint8_t r, std::stringstream& s)
```
leekillough, line 311: `void fmt_reg` should be `std::stringstream& fmt_reg()`  (returning `s`) so it can be used inline like `s << r1 << fmt_reg() << r2`

src/RevTracer.cc
```diff
@@ -0,0 +1,336 @@
...
+        s<<xpr_name[r]; // defined in disasm.h 
+        return;
+    }
+    s << "?" << (unsigned)r;
+    #else
+    s << "x" << std::dec << (uint16_t) r; // Use SST::RevCPU::RevReg?
+    #endif
+}
+
+void RevTracer::fmt_data(unsigned len, uint64_t d, std::stringstream &s)
```
leekillough, line 324: As above, should return `std::stringstream&` so it can be used inline between `<<`'s.

Issue #159, opened 2023-10-11
LR/SC Decode Fault


jleidel:
**Describe the bug**
LR/SC decodes incorrectly for RV64A

**To Reproduce**
1165c: 2f b5 06 10   lr.d  a0, (a3)

   11660: 63 16 b5 00   bne a0, a1, 0x1166c <test_single_thread+0x1f0>

   11664: 2f b6 e6 18   sc.d  a2, a4, (a3)

   11668: e3 1a 06 fe   bnez  a2, 0x1165c <test_single_thread+0x1e0>

produces a fault in 0x1165c having rs1 = 0, 

**Expected behavior**
See above

**Trace**
```
struct Rev64AInstDefaults : RevInstDefaults {

    static constexpr uint8_t     opcode   = 0b0101111;

    static constexpr uint8_t     funct3   = 0b011;

    static constexpr RevRegClass rs2Class = RevRegClass::RegUNKNOWN;

  };

  std::vector<RevInstEntry> RV64ATable = {

    {RevInstEntryBuilder<Rev64AInstDefaults>().SetMnemonic("lr.d %rd, (%rs1)"          ). SetFunct7(0b00010).Setrs1Class(RevRegClass::RegUNKNOWN).Setrs2Class(RevRegClass::RegUNKNOWN).SetImplFunc(&lrd ).InstEntry},

    {RevInstEntryBuilder<Rev64AInstDefaults>().SetMnemonic("sc.d %rd, %rs1, %rs2"      ).SetFunct7(0b00011                        ).Setrs2Class(RevRegClass::RegUNKNOWN).SetImplFunc(&scd ).InstEntry},
```
 

 

…. should change to the following (basically delete the yellow) so that rs1 of lr and rs2 of sc would be properly decoded:

 
```
  struct Rev64AInstDefaults : RevInstDefaults {

    static constexpr uint8_t     opcode   = 0b0101111;

    static constexpr uint8_t     funct3   = 0b011;

  };

  std::vector<RevInstEntry> RV64ATable = {

    {RevInstEntryBuilder<Rev64AInstDefaults>().SetMnemonic("lr.d %rd, (%rs1)"          ).SetFunct7(0b00010).Setrs2Class(RevRegClass::RegUNKNOWN).SetImplFunc(&lrd ).InstEntry},

[...]

PR #170, opened 2023-10-19
adding feature to disable manual coprocessor clocking

jleidel:
This adds a `disable_coproc_clock` parameter that when set to `true` disables the manual clocking of coprocessor subcomponents from the top level RevCPU.  Also note that this PR fixes Issue #159 and removes the explicit deletion of the ClockHandler object as the SST-Core garbage collects this item.

### DIFF
--- a/include/RevCPU.h
+++ b/include/RevCPU.h
@@ -118,6 +118,7 @@ public:
     {"RDMAPerCycle",    "Number of RDMA messages per cycle to inject",  "1"},
     {"testIters",       "Number of PAN test messages to send",          "255"},
     {"splash",          "Display the splash logo",                      "0"},
+    {"disable_coproc_clock",  "Disables the manual clocking of coprocessor subcomponents", "0"},
     )
 
   // -------------------------------------------------------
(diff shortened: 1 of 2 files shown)


include/RevCPU.h
```diff
@@ -118,6 +118,7 @@ class RevCPU : public SST::Component{
     {"RDMAPerCycle",    "Number of RDMA messages per cycle to inject",  "1"},
     {"testIters",       "Number of PAN test messages to send",          "255"},
     {"splash",          "Display the splash logo",                      "0"},
+    {"disable_coproc_clock",  "Disables the manual clocking of coprocessor subcomponents", "0"},
```
rkabrick, line 121: kind of a nitpick but should we keep with camel case for rev options?

rkabrick (approved):
Logic wise nothing is wrong... Only thing that's ultimately up to you is if we keep rev python options as camel case or not

include/RevCPU.h
jleidel: good call; fixed in 3c3ac59

include/RevCPU.h
```diff
@@ -118,6 +118,7 @@ class RevCPU : public SST::Component{
     {"RDMAPerCycle",    "Number of RDMA messages per cycle to inject",  "1"},
     {"testIters",       "Number of PAN test messages to send",          "255"},
     {"splash",          "Display the splash logo",                      "0"},
+    {"disableCoprocClock",  "Disables the manual clocking of coprocessor subcomponents", "0"},
     )
```
donofrio, line 122: Can we call this "IndependentCoProcClock" or something similar? This sounds like the CoProc receives no clock at all

jleidel: changed in 9d29794

PR #171, opened 2023-10-19
Make AdvancePC only work on advancing PC past the current instruction

leekillough:
Make `AdvancePC()` only work on advancing `PC` past the current instruction; require using `SetPC()` for all other `PC` changes.

This prevents tracer output from excessively outputting every `PC` update, while still tracing events such as _taken_ branches and the initialization of `PC` during thread creation.

The signature of `AdvancePC()` has been changed, to prevent using it the old way, and to allow the new way to only pass `Inst` instead of `Inst.instSize`.

Also bring the documentation up to date with the latest function signatures.

**Note:**

> The template definition of `AdvancePC()` is used so that it can be defined before `RevInst`, which is defined in another header, is a _complete type_. It is predeclared only as `struct RevInst;` in `RevRegFile.h`. If the `Inst` parameter was not _dependent_ on a template parameter such as `T`, then `Inst.instSize` would require `Inst` to be a _complete type_ at the time of `AdvancePC()`'s definition. `RevInst` is defined in `RevInstTable.h`, and `RevInstTable.h` `#include`'s `RevRegFile.h` before `RevInst` is defined. `RevInstTable` depends much more on `RevRegFile.h` than the converse -- `RevRegFile.h` only depends on `RevInst` being forward-declared and allowed to be an incomplete type while `RevRegFile.h` is parsed.

### DIFF
--- a/documentation/ExtensionDevelopment.md
+++ b/documentation/ExtensionDevelopment.md
@@ -341,9 +341,8 @@ There are several helper functions to make generating instruction implementation
   void SetPC(const RevFeature* F, T val);
 ```
 ```c++
-  /// AdvancePC: Advance the program counter a certain number of bytes
-  template<typename T>
-  void AdvancePC(const RevFeature* F, T bytes);
+  /// AdvancePC: Advance the program counter to the next instruction
+  void AdvancePC(const RevInst& Inst);
 ```
 ```c++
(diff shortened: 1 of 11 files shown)


jleidel (approved):
Solid changes.  Verified the support for compressed instructions.

PR #172, opened 2023-10-19
Use templatized GetFP()/SetFP() instructions for both float and double, similar to GetX()/SetX()

leekillough:
This provides `GetFP()` and `SetFP()` templated functions for getting/setting the Floating-Point registers, in the same way as `GetX()`/`SetX()` get/set the integer registers.

`GetFP<float>(reg)` and `SetFP(reg, float)` replace the old `GetFP32(reg)` and `SetFP32(reg, float)`.

`GetFP<double>(reg)` and `SetFP(reg, double)` alleviate the need to access `RegFile->DPF[reg]` directly except in memory loads/stores. `RV32D` and `RV64D` no longer need to be `friend` classes to `RevRegFile`.

The new functions will allow tracing of getting/setting FP registers to be easier.

### DIFF
--- a/documentation/ExtensionDevelopment.md
+++ b/documentation/ExtensionDevelopment.md
@@ -345,12 +345,14 @@ There are several helper functions to make generating instruction implementation
   void AdvancePC(const RevInst& Inst);
 ```
 ```c++
-  /// GetFP32: Get the 32-bit float value of a specific FP register
-  float GetFP32(const RevFeature* F, size_t rs) const;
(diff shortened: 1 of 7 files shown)


leekillough:
Hold on... I think I need to add a template parameter and a cast, to allow `RevReg::` `enum`s to be used.

leekillough:
> Hold on... I think I need to add a template parameter and a cast, to allow `RevReg::` `enum`s to be used.

Never mind. It was fine.

jleidel (approved):
good to merge

PR #173, opened 2023-10-23
adding send/get raw data methods for coprocessor base API

jleidel:
Adds two interfaces to the base coproc API: 
```
virtual bool sendRawData(std::vector<uint8_t> Data);
virtual const std::vector<uint8_t> getRawData();
```

Note that these are not `pure` methods as I wanted to avoid breaking all the current external coprocessor implementations. 

I also fixed up the doxygen comments in the base API


leekillough (approved):

donofrio (approved):
Looks good

PR #174, opened 2023-10-23
Fix coProc stall when back to back instructions are issued

donofrio:
Quick fix to prevent instructions from being decoded when the co-processor stalls the pipeline. This prevents an infinite stall from occurring when back to back co-proc instructions are issued.  Updated the coproc test to issue multiple instructions


jleidel (approved):
good to merge

PR #175, opened 2023-10-24
pan extraction

kpgriesser:
There has been a request to support PAN functionality separately. A separate branch (pan) has been created to maintain PAN functionality and likely refactor it so it is better decoupled from the core. The PR removes all PAN code resulting in a significant code reduction.

### DIFF
--- a/include/RevCoProc.h
+++ b/include/RevCoProc.h
@@ -44,16 +44,51 @@ public:
   SST_ELI_REGISTER_SUBCOMPONENT_API(SST::RevCPU::RevCoProc, RevProc*);
   SST_ELI_DOCUMENT_PARAMS({ "verbose", "Set the verbosity of output for the attached co-processor", "0" });
 
+  // --------------------
+  // Virtual methods
+  // --------------------
+
   /// RevCoProc: Constructor
   RevCoProc( ComponentId_t id, Params& params, RevProc* parent);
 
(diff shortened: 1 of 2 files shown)


src/RevProc.cc
```diff
@@ -1720,9 +1719,8 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
                     id, HartToExec, GetActiveThreadID(), ExecPC);
     #endif
 
-    // attempt to execute the instruction as long as it is NOT
-    // the firmware jump PC
-    if( ExecPC != _PAN_FWARE_JUMP_ ){
+
+    if ( true ){
```
donofrio, line 1724: I see the motivation to preserve the logic of the previous implementation, but I think we can safely remove this :)

src/RevProc.cc
```diff
@@ -1920,40 +1918,11 @@ bool RevProc::ClockTick( SST::Cycle_t currentCycle ){
...
-          break;
-        case PanExec::QValid:
-        case PanExec::QError:
-          // done = true;
-        default:
-          break;
-        }
-      }
-    }else if( GetPC() == 0x00ull ) {
+   if( GetPC() == 0x00ull ) {
```
donofrio, line 1925: Is this redundant with line 1921?

PR #183, opened 2023-10-26
Add perf_stats syscall

ukasz:
Implement a new syscall to expose internal performance counters of REV core to the application. Add an example of usage where cycles and number of instructions executed of a simple loop is measured.

Fixed a comment in cpuinfo syscall.


rkabrick (approved):
I see nothing wrong with the code. @jleidel if you're cool with adding it (ie. it provides universal benefit) then you're good to merge on my end

jleidel:
I'm good with it.  Lets merge.

PR #186, opened 2023-10-27
Fix Formatted Output of MemOp enum class

leekillough:
When `_REV_DEBUG_` is defined, some `MemOp` values are outputted. There was no method defined for streaming `MemOp`, which is a `enum class`. In the past it was probably an unscoped enum (`enum` non-`class`) which could be implicit converted to integer and printed as such.

I've added an `operator<<` to output `MemOp` by the tag names.

Unfortunately argument-dependent lookup will not work if the function is defined in a namespace higher in scope than the argument type's definition (`MemOp`). 

`RevMemCtrl.cc` was `using namespace SST::RevCPU;` but was defining everything in global namespace. I have switched it to defining inside `namespace SST:RevCPU { }`, so that it can see argument-dependent lookup from types defined in `SST::RevCPU`. There were a couple of `using namespace SST;` declarations which are redundant now that we define inside of `SST::RevCPU`, which already has other `SST` namespaces imported or searched automatically as parent namespaces.

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -14,6 +14,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <ostream>
 #include <type_traits>
 
 #ifndef _REV_NUM_REGS_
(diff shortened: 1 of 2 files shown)


jleidel (approved):
good to merge

donofrio (approved):

PR #187, opened 2023-10-27
Pipeline deque

rkabrick:
This PR makes the Pipeline in `RevProc` a `std::deque`. We don't do anything with the Pipeline that requires it to be a vector and I'm fairly sure that when we call `Pipeline.erase(Pipeline.front)` that ends up being an `O(n)` operation because you have to shift every element down. 

Unless there's future functionality that requires Pipeline to be a `std::vector` I think this would be a free performance gain.

### DIFF
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -176,7 +176,7 @@ public:
 
   ///< RevProc: Allow a co-processor to manipulate the scoreboard by clearing a bit. Note the RevProcPassKey may only
   ///  be created by a RevCoProc (or a class derived from RevCoProc) so this funciton may not be called from even within
-  ///  RevProc 
+  ///  RevProc
   void ExternalDepClear(RevProcPasskey<RevCoProc>, uint16_t HartID, uint16_t RegNum, bool isFloat){
     DependencyClear(HartID, RegNum, isFloat);
   }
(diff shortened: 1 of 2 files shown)


leekillough:
(`using namespace std` in comments below -- don't do that in code!)

Yes, `deque` is faster if you're inserting the back and removing the front, because with `vector` removing from the front requires O(n) shifting of values.

`queue` is a wrapper container which is usually on top of `deque` and provides a subset of its operations. `priority_queue` is another wrapper container. "Wrapper container" means they aren't true containers in the self-contained sense, but implement on top of another container like `deque` or `list`.

The `ThreadsThatChangedState` did not need a `queue` or `deque` because we were not adding to the back and removing from the front in FIFO fashion -- we were just adding to the back and at some synchronizing points in `RevCPU`, took the whole list and flushed it. A `vector` was fine for that since `push_back` is O(1) amortized and the algorithm doesn't need to `new` / `delete` each element separately. Also the contiguity of the `vector` gives it better cache performance.

leekillough (approved):

donofrio:
This is fine for now, as we create more sophisticated pipeline models with bypass or other features it may be necessary to go back to vector, but at this point a fixed latency model is fine

leekillough:
I don't understand the requirement for a `vector`. A pipeline, as I understand it, would be a FIFO queue structure, where `pop_front` and `push_back` would be the major operations, not arbitrary random access. A `deque` seems much more appropriate than a `vector` for this.

A C++ `deque` is sometimes implemented as a series of contiguous sub-arrays to trade off speed vs memory requirements; see:

https://stackoverflow.com/questions/71784137/stddeque-is-contiguous-memory-container-or-not

https://stackoverflow.com/questions/33165319/why-doesnt-the-c-standard-library-have-a-version-of-deque-with-contiguous-mem

As consumers, we do not need to get involved in the internal implementation details of `deque`. We just need to choose the best data structure / C++ container for the problem at hand.

rkabrick:
Closing as this was included in the `unique-ptrs` PR that was just merged

PR #193, opened 2023-11-01
Custom Stack Allocation

rkabrick:
This PR introduces the ability to customize: 
- The default stack size given to new threads 
- The address range where stacks will be allocated for a given instance of Rev (thus preventing address collisions across multi-rev simulations)

I'm keeping this as a draft until I get back from vacation because I haven't pushed on it enough however everything appears to be working. The only thing I think is left is adding adequate error handling if the user specifies address ranges & stack sizes that are unreasonable. 

The default behavior for rev if these two parameters aren't specified remain how they currently are in devel (`StackSize == 1MB` & stacks begin at `memSize-1024`)


src/RevLoader.cc
```diff
@@ -10,6 +10,7 @@
 
 #include "RevLoader.h"
 #include "RevMem.h"
+#include <_types/_uint32_t.h>
```
leekillough, line 13: Why this strange header instead of `<cstdint>`?

rkabrick: I like to be different

rkabrick: Jk changed it

Issue #130, opened 2023-09-22
fmadd.d returns invalid value


ukasz:
**Describe the bug**
Values returned by fma.double are either not valid or cannot be compared with expected value.
We observe similar problems with other math functions, so a bug comparing double data type is plausible.
Consider the following c code:
```
#define assert(x) if (!(x)) { asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); asm(".byte 0x00"); }
#define T double
int main() {
        volatile T ad = 2.0;
        volatile T bd = 3.0;
        volatile T cd = 5.0;

        T rd = ad * bd + cd;
        assert(rd >= 10.99);
        assert(rd <= 11.01);
		return 0;
}
```
The above program hits an assert on double data type (on float works fine).

The problem is in fma instruction itself or with comparison on double type, which can be
reproduced by the code below:
```
        volatile T ad = 2.0;
        volatile T bd = 3.0;
        volatile T cd = 5.0;

        T rd = (ad * bd) + cd;
        assert(rd > 1000); //fails
        assert(rd < 1000); //fails
        assert(rd != 1000);//ok
``

Disassembly of the test:
```
0000000000011158 <main>:
   11158: 13 01 01 fc   addi    sp, sp, -64
   1115c: 23 3c 11 02   sd      ra, 56(sp)
   11160: 23 38 81 02   sd      s0, 48(sp)
   11164: 13 04 01 04   addi    s0, sp, 64
   11168: 13 05 00 00   li      a0, 0
   1116c: 23 26 a4 fe   sw      a0, -20(s0)
   11170: 13 05 10 00   li      a0, 1
   11174: 13 15 e5 03   slli    a0, a0, 62
   11178: 23 30 a4 fe   sd      a0, -32(s0)
   1117c: 37 15 80 00   lui     a0, 2049
   11180: 13 15 75 02   slli    a0, a0, 39
   11184: 23 3c a4 fc   sd      a0, -40(s0)
   11188: 37 55 00 01   lui     a0, 4101
   1118c: 13 15 65 02   slli    a0, a0, 38
   11190: 23 38 a4 fc   sd      a0, -48(s0)
   11194: 07 30 04 fe   fld     ft0, -32(s0)
   11198: 87 30 84 fd   fld     ft1, -40(s0)
   1119c: 07 31 04 fd   fld     ft2, -48(s0)
   111a0: 43 70 10 12   fmadd.d ft0, ft0, ft1, ft2
[...]

PR #195, opened 2023-11-02
Fix the decoding of Reg-4 instructions

leekillough:
Fix the decoding of Reg-4 instructions (`fmadd`, `fmsub`, `fnmadd`, `fnmsub`) by renaming `Funct7` to `Funct2or7` and filling it in with `Funct2` during Reg-4 instructions, because in Reg-4 instruction encodings, `Funct7` overloads `rs3:Funct2`, where `rs3` is the third source register and `Funct2` is the precision. `Funct2or7` represents either `Funct2` in Reg-4 instructions, or `Funct7` in all other non-compressed instructions. It is distinct from the `Funct2` field used in compressed instructions, whose handling is unchanged by this.

Previously, the `Funct2` field in Reg-4 instructions was ignored, causing `fmadd.d` and similar double-precision instructions to fail.

There was a `fmt` field for floating-point format (precision) in `RevInst`, but it was never filled or used, and using it instead of `Funct2or7` would prevent the precision from being used during instruction table lookup, because there is no more room for the instruction encoding unless another field (like `Funct2or7`) is used. So the `fmt` field in `RevInst` has been removed.

Also templatize (`fmadd`, `fmsub`, `fnmadd`, `fnmsub`) so that the same code is used for single and double precisions.

Add a test case which fails before this change, and passes now.

@ukasz Fixes https://github.com/tactcomplabs/rev/issues/130


src/RevProc.cc
```diff
@@ -1337,31 +1330,34 @@ RevInst RevProc::DecodeInst(){
...
+      Funct2or7 = ((Inst >> 25) & 0b1111111);
       //Atomics have a smaller funct7 field - trim out the aq and rl fields
       if(Opcode == 0b0101111){
-        Funct7 = (Funct7 &0b01111100) >> 2;
+        Funct2or7 = (Funct2or7 &0b01111100) >> 2;
       }
     }
+  }else if((inst65== 0b10) && (inst42 < 0b100)){
+    // R4-Type encodings -- we store the Funct2 precision field in Funct2or7
+    Funct2or7 = DECODE_FUNCT2(Inst);
```
leekillough, line 1345: This is where `Funct2or7` is set to `Funct2` (the precision), when `Inst 4:2` is either `0b00`, `0b01`, `0b10`, or `0b11` (less than `0b100`), corresponding to the 4 REV-4 instructions `fmadd`, `fmsub`, `fnmadd`, `fnmsub`.

jleidel (approved):
good to merge

PR #199, opened 2023-11-06
Clean up RevFlag

leekillough:
Clean up the `enum class` `RevFlag`
- Add a `F_NONE` enum tag for no flags, and use it instead of `0`, `static_cast<RevFlag>(0)`, `REVMEM_FLAGS(0)`, etc.
- Ensure that `RevFlag` has the same underlying type as `StandardMem::Request::flags_t` since they are used interchangeably
- Add casts from `RevFlag` to `StandardMem::Request::flags_t`  when calling SST or outputting streaming IO
- Remove the `IS_ATOMIC` macro constant, replacing it with a `F_ATOMIC` `enum` tag which is computed from the `AMO` tags (except acquire and release)
- Add a `RevFlagHas()` function for testing for the existence of flags, such as `RevFlagHas(flag, RevFlag:F_AQ)`
- Remove the unused function `IsAMOOp()`, to be replaced with `RevFlagHas(flag, RevFlag::F_ATOMIC)` (this is more succinct and avoids the need to cast everywhere a test is made)
- Remove unnecessary `RevCPU::` and similar namespace specifiers such as in `RevCPU::RevFlag::F_AMOADD`, since we're already inside `SST::RevCPU` namespace
- Use `RevFlag` parameters and variables in all Rev functions, instead of `uint32_t` or `StandardMem::Request::flags_t`
- Remove the `REVMEM_FLAGS()` macro which is just a static cast to `RevFlag` -- use `RevFlag::_name_` instead of casting
- Guard against a division-by-zero error when `_REV_DEBUG_` is defined
- Simplify some loops which test flags, by using structured bindings instead of `std::pair::first` and `std::pair::second`

### DIFF
--- a/include/RevInstHelpers.h
+++ b/include/RevInstHelpers.h
@@ -68,9 +68,8 @@ template<typename T>
 bool load(RevFeature *F, RevRegFile *R, RevMem *M, RevInst Inst) {
   MemReq req{};
   if( sizeof(T) < sizeof(int64_t) && R->IsRV32 ){
-    static constexpr auto flags = sizeof(T) < sizeof(int32_t) ?
-      REVMEM_FLAGS(std::is_signed_v<T> ? RevCPU::RevFlag::F_SEXT32 : RevCPU::RevFlag::F_ZEXT32) :
-      REVMEM_FLAGS(0);
+    static constexpr RevFlag flags = sizeof(T) < sizeof(int32_t) ?
+      std::is_signed_v<T> ? RevFlag::F_SEXT32 : RevFlag::F_ZEXT32 : RevFlag::F_NONE;
     uint64_t rs1 = R->GetX<uint64_t>(Inst.rs1); // read once for tracer
     req.Set(rs1 + Inst.ImmSignExt(12),
             Inst.rd, RevRegClass::RegGPR,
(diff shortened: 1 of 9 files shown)


leekillough:
`RevFlag`, `uint2_t`, and `StandardMem::Request::flags_t` used interchangeably in the past, so I cleaned it up to always use `RevFlag` inside of Rev and to cast to `StandardMem::Request::flags_t` when calling SST, and I put a `static_assert` to make sure they are the same type:
```c++
// Ensure RevFlag is same underlying type as StandardMem::Request::flags_t
static_assert( std::is_same_v<StandardMem::Request::flags_t,
               std::underlying_type_t<RevFlag> > );
```
Ours is an `enum` while theirs is a `uint32_t`. I made `uint32_t` the "underlying type" of the `enum`, and used `static_assert` to verify that the SST type and the underlying type of our enum are the same. It won't compile if they differ.

Our `RevFlag` is an "extension" of `StandardMem::Request::flags_t` in that it defines bits above the bottom 16 bits, which are [reserved by SST](https://sst-simulator.org/SSTDoxygen/13.0.0_docs/html/classSST_1_1Interfaces_1_1StandardMem_1_1Request.html#a223993d8380f25efcd58cfba2cca3d20). Our flags are only in the upper 16 bits.

jleidel (approved):
good to merge

PR #200, opened 2023-11-07
Fix mmap test

ukasz:
Size parameter in mmap test was smaller than it should be. Change rev_mmap declaration to return 64bit value, so allocating memory above 2GB is possible.


jleidel (approved):
good to merge

PR #202, opened 2023-11-07
Fix RevRegFile formatted output

leekillough:
Supersedes https://github.com/tactcomplabs/rev/pull/201 (`std::hex` lasted after more than one item; would need to reset the stream with `std::dec`; need to justify the text with the `0x` string prepended).

Minor reformatting to make the column separators line up in source.


jleidel (approved):
good to merge

PR #204, opened 2023-11-10
Fixing the off-by-one bug in mmap

rkabrick:
I was an idiot when making the mmap code. there was an unnecessary `+1` in the call to `AllocMem`. This should fix it.


jleidel (approved):
good to merge

PR #208, opened 2023-11-16
Clean up dependency checks

leekillough:
Fix `DependencySet()` regression caused by 7b12e50a -- it was returning early based on the always-true condition:
```c++
   if( RegNum < _REV_NUM_REGS_ || RegNum != 0 ){ return; }
```
Fix bug which ignored register `0` even when `regClass != RevRegClass::RegGPR` -- only general register `x0` needs to be treated as constant.

Create individual `LSQCheck()` and `ScoreboardCheck()` functions to separate out the concerns, since they are not always used as a pair.

Change `DependencyCheck()` to use new `LSQCheck()` and `ScoreboardCheck()` functions -- now a simple `||` of tests can be used, instead of using more abstract "structured bidings" to iterate through lists of tuples.

Add ability of coprocessor to check dependencies, superseding https://github.com/tactcomplabs/rev/pull/177

### DIFF
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -153,7 +153,7 @@ public:
   void MarkLoadComplete(const MemReq& req);
 
   ///< RevProc: Get pointer to Load / Store queue used to track memory operations
-  auto GetLSQueue() const { return LSQueue; }
+  std::shared_ptr<std::unordered_map<uint64_t, MemReq>> GetLSQueue() const { return LSQueue; }
 
   ///< RevProc: Add a co-processor to the RevProc
   void SetCoProc(RevCoProc* coproc);
--- a/include/RevProc.h
+++ b/include/RevProc.h
@@ -153,13 +153,21 @@ public:
   void MarkLoadComplete(const MemReq& req);
 
   ///< RevProc: Get pointer to Load / Store queue used to track memory operations
-  std::shared_ptr<std::unordered_map<uint64_t, MemReq>> GetLSQueue(){ return LSQueue; }
+  auto GetLSQueue() const { return LSQueue; }
 
   ///< RevProc: Add a co-processor to the RevProc
   void SetCoProc(RevCoProc* coproc);
 
-
   //--------------- External Interface for use with Co-Processor -------------------------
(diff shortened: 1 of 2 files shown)


include/RevProc.h
```diff
@@ -153,13 +153,21 @@ class RevProc{
   void MarkLoadComplete(const MemReq& req);
 
   ///< RevProc: Get pointer to Load / Store queue used to track memory operations
-  std::shared_ptr<std::unordered_map<uint64_t, MemReq>> GetLSQueue(){ return LSQueue; }
+  auto GetLSQueue() const { return LSQueue; }
```
donofrio, line 156: prefer the explicit declaration of the return type (I know it's long, but nice to be explicit about what it is)

src/RevProc.cc
```diff
@@ -1530,43 +1530,21 @@ void RevProc::HandleALUFault(unsigned width){
...
+  const RevRegFile* regFile = GetRegFile(HartID);
+  const RevInstEntry* E = &InstTable[I->entry];
+
+  return
+    // check LS queue for outstanding load
+    LSQCheck(HartID, regFile, I->rs1, E->rs1Class) ||
+    LSQCheck(HartID, regFile, I->rs2, E->rs2Class) ||
+    LSQCheck(HartID, regFile, I->rs3, E->rs3Class) ||
+    LSQCheck(HartID, regFile, I->rd , E->rdClass) ||
+
```
donofrio, line 1542: we were checking `rd` previously, but not sure that is necessary in all cases? We may be leaning on it for atomics in an unexpected way, but we should experiment with removing in as in the general case a WAW hazard is non blocking

donofrio (approved):
Overall looks good to merge

include/RevProc.h
leekillough: I didn't do it for shorter names, but for single-points-of-definition (of the type of `LSQueue`). Since `GetLSQueue()` is a simple accessor which just returns `LSQueue`, we shouldn't need to repeat the type.

src/RevProc.cc
leekillough: WAW could be a problem with atomics with a threaded simulation or any out-of-order processing (do we atomically increment or decrement first?).

PR #210, opened 2023-11-20
Tracer support for Mem Hierarchy

kpgriesser:
Enhanced the tracer to properly show load requests and load completion when using mem hierarchy model.  The Tracer.md file under the documentation area has been updated. Use ccmake to configure for using spike disassembler or the internal one to rev core.  To view mem vs memh and rv32i vs rv64g run the tracer test as follows:

cd test/tracer
make -j

The following trace shows a load request and it's completion after another independent instruction has executed.

RevCPU[cpu0:Render:412322000]: Core 0; Hart 0; Thread 1; *I 0x14f48:fec42783    lw      a5, -20(s0)      0x3ffffbfc<-s0 a5<-[0x3ffffbe8,4]
RevCPU[cpu0:Render:412323000]: Core 0; Hart 0; Thread 1; *I 0x14f4c:01400593    li      a1, 20   0x0<-zero a1<-0x14
RevCPU[cpu0:Render:412423000]: Hart 0; *A 0x000007d0<-[0x3ffffbe8,4] a5<-0x000007d0

### DIFF
--- a/include/RevTracer.h
+++ b/include/RevTracer.h
@@ -139,7 +139,6 @@ namespace SST::RevCPU{
     uint64_t a;  // reg             adr                     adr
     uint64_t b;  // value           len                     len
     uint64_t c;  // origin(TODO)    data (limited 8 bytes)  reg
-    TraceRec_t() {};
     TraceRec_t(TraceKeyword_t Key, uint64_t A, uint64_t B, uint64_t C=0)
       : key(Key), a(A), b(B), c(C) {};
   };
@@ -152,7 +151,7 @@ namespace SST::RevCPU{
     std::string defaultMnem = "?";
     std::string& fallbackMnemonic = defaultMnem;
     bool valid = false;
-    void set(size_t _cycle, unsigned _id, unsigned _hart, unsigned _tid, std::string& _fallback) {
+    void set(size_t _cycle, unsigned _id, unsigned _hart, unsigned _tid, const std::string& _fallback) {
       cycle = _cycle;
       id = _id;
       hart = _hart;
@@ -172,6 +171,13 @@ namespace SST::RevCPU{
     uint64_t data; // first 8 bytes max
     bool isFloat = false;
     bool wait4Completion = false;
+    CompletionRec_t(unsigned int hart, uint16_t destReg, size_t len, uint64_t addr, 
+      void *data, RevRegClass regClass) 
+      : hart(hart), destReg(destReg), len(len), addr(addr),
+        isFloat(regClass == RevRegClass::RegFLOAT), wait4Completion(true) 
+    {
+      memcpy(&this->data, data, len > sizeof(this->data) ? sizeof(this->data) : len);
+    }
   };
 
   class RevTracer{
@@ -210,7 +216,7 @@ namespace SST::RevCPU{
       /// RevTracer: capture 64-bit program counter
       void pcWrite(uint64_t newpc);
       /// RevTracer: render instruction execution trace to output stream and update tracer state
-      void Exec(size_t cycle, unsigned id, unsigned hart, unsigned tid, std::string& fallbackMnemonic);
+      void Exec(size_t cycle, unsigned id, unsigned hart, unsigned tid, const std::string& fallbackMnemonic);
       /// RevTracer: Render captured states
       void Render(size_t cycle);
       /// RevTracer: control whether to render captured states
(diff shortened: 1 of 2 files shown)


include/RevTracer.h
```diff
@@ -114,24 +119,61 @@ namespace SST::RevCPU{
...
     TraceKeyword_t key;
-                 // register        memory
-    uint64_t a;  // reg             adr
-    uint64_t b;  // value           len
-    uint64_t c;  // origin(TODO)    data (limited to show 8 bytes)
+                 // register        memory                  memh
+    uint64_t a;  // reg             adr                     adr
+    uint64_t b;  // value           len                     len
+    uint64_t c;  // origin(TODO)    data (limited 8 bytes)  reg
     TraceRec_t() {};
```
leekillough, line 142: Are there instances where you want `TraceRec_t` to be uninitialized, and initialize it later (e.g. with aggregate initialization `tracerec = {key, a, b, c}`?) To avoid uninitialized members, I usually use default member initialization, e.g.
```c++
TraceKeyword_t key = -1;
uint64_t a = 0;
uint64_t b = 0;
uint64_t c = 0;
```
Or just remove `TraceRec_t() {}` default constructor, since the other constructor initializes all members.

include/RevTracer.h
```diff
@@ -114,24 +119,61 @@ namespace SST::RevCPU{
...
 
+  struct InstHeader_t {
+    size_t cycle;
+    unsigned id;
+    unsigned hart;
+    unsigned tid;
+    std::string defaultMnem = "?";
+    std::string& fallbackMnemonic = defaultMnem;
+    bool valid = false;
+    void set(size_t _cycle, unsigned _id, unsigned _hart, unsigned _tid, std::string& _fallback) {
```
leekillough, line 155: Does `_fallback`  need to be a non-`const` `std::string&`?

include/RevTracer.h
```diff
@@ -159,18 +201,28 @@ namespace SST::RevCPU{
...
+      /// RevTracer: data returning from memory read (memh)
+      void memReadResponse(size_t len, void *data, const MemReq* req);
       /// RevTracer: capture 32-bit program counter
       void pcWrite(uint32_t newpc);
       /// RevTracer: capture 64-bit program counter
       void pcWrite(uint64_t newpc);
-      /// RevTracer: render trace to output stream and update tracer state
-      void InstTrace(size_t cycle, unsigned id, unsigned hart, unsigned tid, std::string& fallbackMnemonic);
+      /// RevTracer: render instruction execution trace to output stream and update tracer state
+      void Exec(size_t cycle, unsigned id, unsigned hart, unsigned tid, std::string& fallbackMnemonic);
```
leekillough, line 213: Does `fallbackMnemonic` need to be a non-`const` `std::string&`?

src/RevTracer.cc
```diff
@@ -182,6 +182,25 @@ void RevTracer::memRead(uint64_t adr, size_t len, void *data)
...
+
+void RevTracer::memReadResponse(size_t len, void *data, const MemReq* req)
+{
+    if (req->DestReg==0) return;
+    CompletionRec_t c;
+    c.hart = req->Hart;
+    c.destReg = req->DestReg;
+    c.len = len;
+    c.addr = req->Addr;
+    c.data = *((uint64_t*) data); // First 8 bytes only
```
leekillough, line 198: This may cause a bus error for unaligned data, and violate strict aliasing rules. Use:
```c++
memcpy(&c.data, data, sizeof(c.data));
```
(The `sizeof` is used so that if `c.data` changes size, the code adapts automatically.)

src/RevTracer.cc
```diff
@@ -182,6 +182,25 @@ void RevTracer::memRead(uint64_t adr, size_t len, void *data)
...
 
+void SST::RevCPU::RevTracer::memhSendRead(uint64_t adr, size_t len, uint16_t reg)
+{
+    traceRecs.emplace_back(TraceRec_t(MemhSendLoad, adr, len, reg));
+}
+
+void RevTracer::memReadResponse(size_t len, void *data, const MemReq* req)
+{
+    if (req->DestReg==0) return;
+    CompletionRec_t c;
```
leekillough, line 193: You might want to put all of the assignments into one constructor call, e.g.
```c++
 CompletionRec_t c{req->Hart, req->DestReg, len, req->Addr, data, req->RegType};
 completionRecs.emplace_back(c);
```
and then
```c++
CompletionRec_t(unsigned hart, uint32_t destReg, size_t len, uint64_t addr, void* data, RevRegClass regClass) 
: hart(hart), destReg(destReg), len(len), addr(addr), isFloat(regClass == RevRegClass::RegFLOAT), wait4Completion(true) {
  memcpy(&this->data, data, len > sizeof(this->data) ? sizeof(this->data) : len);
}
```

PR #211, opened 2023-11-22
Better testing

rkabrick:
# Summary of Changes
  ## add_rev_test macro
  - Added `add_rev_test` macro to `test/CMakeLists.txt`
    This macro has the following parameters:
    - `test_name` - Name of the test
    - `test_dir` - directory containing the test files
    - `timeout` - Same as `add_test` timeout
    - `labels` - Same as `add_test` labels
  - OPTIONALLY: `SCRIPT <script>` - Script to run the test if you want special options (Effectively preserves old behavior)

     The advantage of this macro is it will automatically run every test with and without memHierarchy UNLESS you specify the `SCRIPT` option

## Rev Model Options Python File (`test/rev-model-options-config.py`)
   File uses the `argparse` library to parse command line arguments and setup the relevant parts of rev

  Currently there is support for the following options:
  
- "--numCores": Number of Rev Cores per RevCPU
   - default=1
   
- "--numHarts": Number of HARTs per Rev Core
   - default=1
   
- "--program": Program for rev to run
   - default="a.out"
   
- "--enableMemH": Whether to enable memHierarchy
   - choices=[0, 1]
   - default=0
   
- "--verbose": Verbosity level
   - default=2

- "--machine": Machine type/configuration
   - default="[0:RV64GC]"

- "--args": Command line arguments to pass to the target executable
   - default=""
   
   You call it like so: `sst rev_model_options_config.py -- --numCores 2 --numHarts 2... etc.`

# Outstanding Questions
  - @donofrio, idk if you want me to make the `isa` tests use this macro or not
  - Is there anything else you want supported? I could probably extend it to do 32-bit support but I'm not sure that's a priority
  - We could potentially make a generic makefile that compiles all c files in the test's directory but I'm not sure if that's worth it
  - If anyone else has tests that should be added to this, let me know

  *Final note: This is currently failing the same tests as @donofrio's PR. This PR should not be merged until that one is merged and the tests are passing.*


rkabrick:
I think there was an issue with my rebase because there shouldn't be any changes to RevInstHelpers or any of the RV{32|64}{I|M|F...etc} files.

leekillough:
> `riscv64-unknown-elf-objdump -D --source`

Might also want to include `-C` (demangle C++ names)

donofrio:
while we are adding our favorite switches... I'll add `-Mno-aliases` so you get the compressed instructions

donofrio:
We can leave the isa tests as they are for now

test/big_loop/run_big_loop.sh
```diff
@@ -5,8 +5,8 @@ make clean && make
 
 # Check that the exec was built...
 if [ -f big_loop.exe ]; then
-  sst --add-lib-path=../../build/src/ ./big_loop.py
+	sst --add-lib-path=../../build/src/ ../rev-model-options-config.py -- --program="big_loop.exe" --numHarts=1 --numCores=1
```
leekillough, line 8: `[[ -x big_loop.exe ]]`

test/cache_2/run_cache_2.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+#Build the test
+make clean && make
+
+# Check that the exec was built...
+if [ -f cache_2.exe ]; then
```
leekillough, line 7: `[[ -x cache_2.exe ]]`

test/coproc_ex/run_coproc_ex.sh
```diff
@@ -4,10 +4,10 @@
 make
 
 # Check that the exec was built...
-if [ -f ex1.exe ]; then
-  sst-info revcpu
-  sst --add-lib-path=../../build/src/ ./rev-test-coproc_ex.py
+if [ -f coproc_ex.exe ]; then
```
leekillough, line 7: `[[ -x coproc_ex.exe ]]`

test/syscalls/vector/Makefile
```diff
@@ -14,7 +14,7 @@
 
 EXAMPLE=vector
 #CC=riscv64-unknown-linux-gnu-gcc
-CC="${RVCC}"
+CC="${RVCXX}"
```
leekillough, line 17: Why are you setting `CC` to `"${RVCXX}"` instead of `"${RVCC}"`?

leekillough (changes requested):

.gitignore
```diff
@@ -9,3 +9,5 @@
 *~
 .gdb_history
 *.out
+build/test/*
```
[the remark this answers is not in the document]
rkabrick: Ah yeah I don't like including the build folder and I just get frustrated when I delete the build folder to do a clean build and then the .gitignore is deleted. If we have to keep the empty build directory in the repo then I'd prefer to keep these lines in the gitignore

test/syscalls/vector/Makefile
rkabrick: ah that was a quick fix while I was testing. good catch

.gitignore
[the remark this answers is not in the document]
leekillough: I would use `build/*` then, instead of `build/test/*`, in this `.gitignore`.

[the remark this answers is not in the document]
leekillough: I'm not sure what difference `build/*` would have over simply `build` or `build/`, except that I think that `build/.gitignore` would be ignored with `build` or `build/`, but not with `build/*`, which is what we want (so that `build/.gitignore` changes are still tracked).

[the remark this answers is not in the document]
leekillough: So I guess that I'm saying that `build/*` should be used. That would cover your case while not including `build/.gitignore`, because `*` does not include filenames beginning with `.`.

rkabrick:
# New Label Based Approach

  I made this a little more extensible than before... now most of the test generation is done based on the labels specified in the test

## Label-Based Configurations

### `rv32` Label

- **Machine Configuration**: Sets the `machine_args` parameter to `"[CORES:RV32GC]"`. If not present, the default is `"[CORES:RV64GC]"`.

### `multihart` Label

- **Number of HARTs**: Sets the `numHartsList` to `1 2 4`, allowing tests to run with multiple HARTs (Hardware Threads).

### `multicore` Label

- **Number of Cores**: Sets the `numCoresList` to `1 2 4`, enabling tests to run with multiple cores.

### `pthreads` (or `multithreading`) Label

- **Multi-threading Configuration**: Similar to `multihart` and `multicore`, sets both `numHartsList` and `numCoresList` to `1 2 4`.

### `memh` Label

- **Memory Hierarchy**: If this label is present, additional tests with memHierarchy enabled are added (`--enableMemH=1`).

### `cpp` Label

- **Start Symbol**: If this label is present, the `startSymbol` is set to `"[0:_start]"`. Otherwise, it defaults to `"[0:main]"`.

*Note*: These labels _stack_... meaning if you specify a test with `multihart;multicore;memh` you will be generating 3*3*2 tests

leekillough (approved):

PR #215, opened 2023-11-30
Changes to the compressed test fix

leekillough:
~Revert  1d693169fb which causes `AMOADD_CXX` and `AMOADD_CXX_MEMH` failures (why?).~

Merge all of the latest changes from `devel`.

In compressed instructions where `rs1 = rd`, make the decoding code assign to both `rs1` and `rd` in the same statement, to avoid mistakes and to prevent creating a dependency on whether `rs1` or `rd` is initialized first (some of the previous code assigned to one and then used it in a later statement to assign to the other, which is confusing -- this makes it clear that both are being assigned the same value at once).

Rename `make_lsq_hash()` to `LSQHash()`, keeping in line with CamelCase.

Clean up `MemReq` class to only have a constructor, and not a `.Set()` function to change it after it's created (it still has an assignment `operator=` if really necessary). Directly initialize `MemReq` objects using this constructor, instead of default-initializing a value and using `.Set()` afterwards.

Since they are common operations, add `LSQHash()` and `LSQHashPair()` functions to `MemReq`. `LSQHash()` returns the namespace-scope `LSQHash()` on selected `MemReq` members, and `LSQHashPair()` returns `{LSQHash(), *this}`, a pair consisting of the hash and the request, for storing in `std::unordered_map`.

Make data members of `MemReq` private, only visible to `MemReq` and friend classes `RevTracer`, `RevMem` and `RevProc`.

Instead of `MarkLoadComplete()` being a member functor of `MemReq` which expects a `MemReq` argument, change `MarkLoadComplete()` to be a member function of `MemReq` which takes no arguments and calls a `MarkLoadCompleteFunc` functor with `*this` as its argument, so that `req.MarkLoadComplete(req)` can simply be called as `req.MarkLoadComplete()`.

Use `std::function<void(const MemReq&)>` consistently throughout the code. Previously `std::function<void(MemReq)>` was used in some places. While it may work because of conversions, it is not as clean.

Using the new helper functions, replace the common idiom `R->LSQueue->insert({make_lsq_hash(Inst.rd, RevRegClass::RegGPR, F->GetHartToExecID()), req});` with the shorter ` R->LSQueue->insert(req.LSQHashPair());`

In many `.cpp` files, `using namespace SST::RevCPU;` was used to pull in symbols, but the rest of the file still defined functions in global namespace unless they were qualified names (using `::`). Change all of those files to use `namespace SST::RevCPU{ ... }` instead of `using namespace SST::RevCPU;`, so that everything in the file is defined in the `SST:RevCPU` namespace.


donofrio:
Change https://github.com/tactcomplabs/rev/commit/1d693169fb6616c42161b25789cece4234c31882 caused the AMO tests to fail because now the AMO tests run to their proper length. The code there fixes a case where we improperly decode a c.jr instruction. We set the return to x0, which sets the PC to 0x0 which triggers an early exit. Without this change we "compete" the test when we return from a call to a function at sim time 47,000 with the return address incorrectly set to 0x00. With this change in place we continue running until sim time 425,000 when we finally hit the error. So this fix exposes an additional, yet unknown issue with atomics

Issue #207, opened 2023-11-13
Double addition ends as NaN


ukasz:
**Bug description**

This is a new weird FP related problem I found.
This is reproducible only on CLANG/LLVM, I have not been able to reproduce it on GCC in any way.
Consider the following test case, where we perform a simple addition of a constant to float variable.
Depending on the compiler this can be treated as addition of two float type variables, or be promoted to a double data type before addition and then back to float in the end.
```
// clang bug.c -o bug.exe -static -nostdlib -march=rv64imafd

#define assert(x) if (!(x)) { asm(".dword 0x00000000"); }
float kek(float x) {
	return x + 1.0;
}

int main() {
	float rf = kek(3.0);
	assert (rf > 3.0);
	assert (rf < 5.0);
	return 0;
}
```

Both asserts should pass, but instead both can fail - rf value is NaN.
The issue will go away if we change "1.0" to "1.0f" which will make both a constant and the addition of a float type, instead of a double.
The difference in disassembly is listed below:
```
float kek(float x) {
	return x + 1.0;
}

0000000000011244 <kek>:
   11244: 13 01 01 fe   addi    sp, sp, -32
   11248: 23 3c 11 00   sd      ra, 24(sp)
   1124c: 23 38 81 00   sd      s0, 16(sp)
   11250: 13 04 01 02   addi    s0, sp, 32
   11254: 27 26 a4 fe   fsw     fa0, -20(s0)
   11258: 87 27 c4 fe   flw     fa5, -20(s0)
   1125c: d3 87 07 42   fcvt.d.s        fa5, fa5
   11260: 37 25 01 00   lui     a0, 18
   11264: 07 37 85 2e   fld     fa4, 744(a0)
   11268: d3 f7 e7 02   fadd.d  fa5, fa5, fa4
   1126c: 53 f5 17 40   fcvt.s.d        fa0, fa5
   11270: 83 30 81 01   ld      ra, 24(sp)
   11274: 03 34 01 01   ld      s0, 16(sp)
   11278: 13 01 01 02   addi    sp, sp, 32
   1127c: 67 80 00 00   ret


```
Float constant:
```
float kek(float x) {
	return x + 1.0f;
}

0000000000011244 <kek>:
   11244: 13 01 01 fe   addi    sp, sp, -32
   11248: 23 3c 11 00   sd      ra, 24(sp)
   1124c: 23 38 81 00   sd      s0, 16(sp)
[...]

PR #227, opened 2023-12-16
Fix FCVT instructions

leekillough:
- The instruction table had a wrong entry for `fcvt.d.s`, and the `fcvtsd` and `fcvtds` "execution" functions were implemented backwards.

- More FCVT tests were added. One had a NaN test but I commented it out because the compiler was removing the `isnan()` test and always executing the assert.

- FMV and FSW / FSD instructions need to load a FP register in raw mode instead of unboxing the NaN and verifying that the upper bits are all 1 (see section 12.2 of the unprivleged spec).

- A new raw mode `FMV_FS` template parameter has been added to `RevRegFile::GetFP<...>`, but it defaults to false (the old behavior). It's only used when loading a FP register for a FMV, FSW or FSD. (Most of the text differences in `GetFP` are simply to get it to assign to `fp32` once and return that, instead of having multiple return points.)

- Some compressed decoding functions still had comments which said `Removed - set in decode`.

Fixes https://github.com/tactcomplabs/rev/issues/207


jleidel (approved):
good to merge

PR #229, opened 2023-12-19
Load/Store Queue Update - allowing multiple loads per register target

donofrio:
This converts the `unordered_map` implementation of the `LSQueue` to a `multimap.` This allows multiple outstanding loads per register destination to be spawned. All load requests are now cleared based on the address read.  In addition, the pre-fetcher now tracks outstanding loads in an internal vector. On each pipeline clock the prefetcher checks the head of the vector against the outstanding loads and if there is a match this indicates that the instruction fill has not yet completed.  This scheme allows fill requests to complete in any order.

When the pre-fetcher makes a `Fill()` request all addresses are added to the internal vector. The `MarkLoadComplete` function in `RevProc` now will also call into the pre-fetcher if the destination address is `x0`.  (only load destinations spawned by the prefetcher should be targeting `x0`).  If the returned load matches an address in the vector then it is erased. 

For _all_ loads a dependency will only be cleared if _all_ outstanding loads for a given register have been returned.

### DIFF
--- a/include/RevHart.h
+++ b/include/RevHart.h
@@ -26,7 +26,7 @@ class RevHart{
   EcallState Ecall{};
 
   ///< RevHart: Pointer to the Proc's LSQueue
-  const std::shared_ptr<std::unordered_map<uint64_t, MemReq>>& LSQueue;
+  const std::shared_ptr<std::multimap<uint64_t, MemReq>>& LSQueue;
 
   ///< RevHart: Pointer to the Proc's MarkLoadCompleteFunc
   std::function<void(const MemReq&)> MarkLoadCompleteFunc;
@@ -40,7 +40,7 @@ class RevHart{
 
 public:
   ///< RevHart: Constructor
-  RevHart(unsigned ID, const std::shared_ptr<std::unordered_map<uint64_t, MemReq>>& LSQueue,
+  RevHart(unsigned ID, const std::shared_ptr<std::multimap<uint64_t, MemReq>>& LSQueue,
           std::function<void(const MemReq&)> MarkLoadCompleteFunc)
     : ID(ID), LSQueue(LSQueue), MarkLoadCompleteFunc(std::move(MarkLoadCompleteFunc)) {}
 
--- a/include/RevHart.h
+++ b/include/RevHart.h
@@ -26,7 +26,7 @@ class RevHart{
   EcallState Ecall{};
 
   ///< RevHart: Pointer to the Proc's LSQueue
-  const std::shared_ptr<std::multimap<uint64_t, MemReq>>& LSQueue;
+  const std::shared_ptr<std::unordered_multimap<uint64_t, MemReq>>& LSQueue;
 
   ///< RevHart: Pointer to the Proc's MarkLoadCompleteFunc
   std::function<void(const MemReq&)> MarkLoadCompleteFunc;
@@ -40,7 +40,7 @@ class RevHart{
 
 public:
   ///< RevHart: Constructor
-  RevHart(unsigned ID, const std::shared_ptr<std::multimap<uint64_t, MemReq>>& LSQueue,
+  RevHart(unsigned ID, const std::shared_ptr<std::unordered_multimap<uint64_t, MemReq>>& LSQueue,
           std::function<void(const MemReq&)> MarkLoadCompleteFunc)
     : ID(ID), LSQueue(LSQueue), MarkLoadCompleteFunc(std::move(MarkLoadCompleteFunc)) {}
 
--- a/include/RevPrefetcher.h
+++ b/include/RevPrefetcher.h
@@ -43,7 +43,7 @@ public:
   bool IsAvail(uint64_t Addr);
 
   /// RevPrefetcher: Mark Instruction fill as complete
-  void MarkInstructionLoadComplete(MemReq& req);
+  void MarkInstructionLoadComplete(const MemReq& req);
 
 private:
   RevMem *mem;                                ///< RevMem object
--- a/include/RevPrefetcher.h
+++ b/include/RevPrefetcher.h
@@ -31,7 +31,7 @@ public:
   RevPrefetcher(RevMem *Mem, RevFeature *Feature, unsigned Depth,
                 std::shared_ptr<std::multimap<uint64_t, MemReq>> lsq,
                 std::function<void(const MemReq&)> func)
-    : mem(Mem), feature(Feature), depth(Depth), LSQueue(lsq), MarkLoadAsComplete(func){}
+    : mem(Mem), feature(Feature), depth(Depth), LSQueue(lsq), MarkLoadAsComplete(func), OutstandingFetchQ(){}
 
   /// RevPrefetcher: destructor
   ~RevPrefetcher() = default;
@@ -42,6 +42,9 @@ public:
   /// RevPrefetcher: determines in the target instruction is already cached in a stream
   bool IsAvail(uint64_t Addr);
 
+  /// RevPrefetcher: Mark Instruction fill as complete
+  void MarkInstructionLoadComplete(MemReq& req);
+
 private:
   RevMem *mem;                                ///< RevMem object
   RevFeature *feature;                        ///< RevFeature object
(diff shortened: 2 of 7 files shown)


include/RevHart.h
```diff
@@ -26,7 +26,7 @@ class RevHart{
   EcallState Ecall{};
 
   ///< RevHart: Pointer to the Proc's LSQueue
-  const std::shared_ptr<std::unordered_map<uint64_t, MemReq>>& LSQueue;
+  const std::shared_ptr<std::multimap<uint64_t, MemReq>>& LSQueue;
```
leekillough, line 29: There's an `std::unordered_multimap` if you want to retain `O(1)` amortized lookups. I would do a performance comparison between `std:multimap` and `std::unordered_multimap`.

include/RevPrefetcher.h
```diff
@@ -42,14 +42,18 @@ class RevPrefetcher{
   /// RevPrefetcher: determines in the target instruction is already cached in a stream
   bool IsAvail(uint64_t Addr);
 
+  /// RevPrefetcher: Mark Instruction fill as complete
+  void MarkInstructionLoadComplete(MemReq& req);
```
leekillough, line 46: All other `MarkLoadComplete` functions have taken a `const Memreq&`.

include/RevProc.h
```diff
@@ -26,13 +26,13 @@
 #include <iostream>
 #include <list>
 #include <map>
+#include <unordered_map>
```
leekillough, line 29: I had my editor sort the headers alphabetically.

src/RevPrefetcher.cc
```diff
@@ -64,6 +74,17 @@ bool RevPrefetcher::IsAvail(uint64_t Addr){
   return false;
 }
 
+void RevPrefetcher::MarkInstructionLoadComplete(MemReq& req){
+  auto it = OutstandingFetchQ.begin();
+  while(!OutstandingFetchQ.empty() && (it != OutstandingFetchQ.end())){
```
leekillough, line 79: Is testing `OutstandingFetchQ.empty()` necessary? I thought that if it were empty, `it == OutstandingFetchQ.end()`.

src/RevProc.cc
```diff
@@ -1638,15 +1638,32 @@ unsigned RevProc::GetNextHartToDecodeID() const {
...
+  auto it = LSQueue->equal_range(req.LSQHash());      // Find all outstanding dependencies for this register
+  bool addrMatch = false;
+  if( it.first != LSQueue->end()){                  
+    for (auto i = it.first; i != it.second; ++i){    // Iterate over all outstanding loads for this reg (if any)
+        if(i->second.Addr == req.Addr){
+          if(LSQueue->count(req.LSQHash()) == 1){   // Only clear the dependency if this is the LAST outstanding load for this register
+            DependencyClear(i->second.Hart, i->second.DestReg, (i->second.RegType == RevRegClass::RegFLOAT));
+          }
+          MemReq temp = req;
+          sfetch->MarkInstructionLoadComplete(temp);
```
leekillough, line 1650: Why do we need to create a `temp` and send a non-const reference to it? All the other uses of MarkLoadComplete use `const MemReq&` -- see https://github.com/tactcomplabs/rev/pull/219

> Use `std::function<void(const MemReq&)>` consistently throughout the code. Previously `std::function<void(MemReq)>` was used in some places. While it may work because of conversions, it is not as clean.

src/RevPrefetcher.cc
donofrio: It appeared to be running off the end without that check

src/RevProc.cc
donofrio: At one point there was an issue w/ a what appeared to be a double-delete of req when it was being erased from multiple functions. This appears to no longer be necessary. Updated to just use req

include/RevProc.h
donofrio: sorted

include/RevPrefetcher.h
donofrio: updated

include/RevHart.h
donofrio: equivalent functionality between unordered_multimap and multimap for this application so if we get better performance that works. updated

src/RevPrefetcher.cc
jleidel: does `LSQHash()` return `LSQueue.end()`? That might be why the `empty()` is required.

leekillough: `LSQHash()`, whether as the namespace function accepting 3 parameters, or the `MemReq` member function depending on 3 `MemReq` properties, returns a `uint64_t` hash, which if used as a key in `find()` for a `std::unordered_multimap`, returns an iterator which should equal `end()` if there was no match.

https://github.com/tactcomplabs/rev/blob/1468d6d672bd69abe2f33273c13fb9455ee76a7c/common/include/RevCommon.h#L88

In this case `begin()` is used, which should equal `end()` if the map is empty. You can erase entries, which invalidates the erased iterator and its pointed-to-data, but the next iterator should be okay, even if it equals `end()`. It might be necessary to compute the next iterator before `erase()` is called on the current iterator.

leekillough:
```c++
OutstandingFetchQ.erase(it++);
```
Instead of:
```c++
OutstandingFetchQ.erase(it);
it++;
```
... because `it++`, which is executed before `erase()` is called, will set `it` to the next iterator before `erase()` is called, and then even if `erase()` invalidates the old value of `it` and what it pointed to, `it` will already be incremented to the next value before `erase()` is called, and will be set to `end()` if the map is finished being traversed.

donofrio: fixed

leekillough: Think of deleting a linked list entry, whose next  entry's iterator is computed from a `next` pointer stored in the current entry. By calling `erase(it++)` you are computing the `next` entry and setting `it` to point to it, before calling `erase(it)` on the old value of `it` and destroying the `next` pointer store in `*it`.

src/RevPrefetcher.cc
```diff
@@ -64,6 +74,18 @@ bool RevPrefetcher::IsAvail(uint64_t Addr){
   return false;
 }
 
+void RevPrefetcher::MarkInstructionLoadComplete(const MemReq& req){
+  auto it = OutstandingFetchQ.begin();
+  while((it != OutstandingFetchQ.end())){
+    if(it->Addr == req.Addr){
+      OutstandingFetchQ.erase(it++);
+      break;
```
leekillough, line 82: Actually, since you're `break`ing and `it` isn't used afterwards, it is not necessary to increment `it` here.

src/RevProc.cc
```diff
@@ -1638,15 +1638,31 @@ unsigned RevProc::GetNextHartToDecodeID() const {
...
-  if( it != LSQueue->end()){
-    DependencyClear(it->second.Hart, it->second.DestReg, (it->second.RegType == RevRegClass::RegFLOAT));
-    LSQueue->erase(it);
+  auto it = LSQueue->equal_range(req.LSQHash());      // Find all outstanding dependencies for this register
+  bool addrMatch = false;
+  if( it.first != LSQueue->end()){                  
+    for (auto i = it.first; i != it.second; ++i){    // Iterate over all outstanding loads for this reg (if any)
+        if(i->second.Addr == req.Addr){
+          if(LSQueue->count(req.LSQHash()) == 1){   // Only clear the dependency if this is the LAST outstanding load for this register
+            DependencyClear(i->second.Hart, i->second.DestReg, (i->second.RegType == RevRegClass::RegFLOAT));
```
leekillough, line 1647: I am a little concerned about this API which uses an all-or-nothing "isFloat" parameter, rather than a more general `RevRegClass`. Since it is for a specific register, and is correct for that register, I suppose that's okay.

The recent `fcvt` failure may indicate a problem with dependence analysis when both GP and FP registers are involved.

jleidel:
@leekillough are we good to merge this as is?  This PR is in the critical path for an immediate set of issues that need to get fixed

leekillough:
> @leekillough are we good to merge this as is? This PR is in the critical path for an immediate set of issues that need to get fixed

The redundant increment can be removed, or left as-is.  I am still unsure of the reason why `.empty()` check was needed. But simply going through the list until finding a match and then erasing an entry and _ending the loop_, should not require calling `empty()` or preincrementing iterators before `erase()`'ing them.

PR #231, opened 2023-12-19
adding test to context save/restore backing DRAM memory

jleidel:
This test has an example of how to utilize a binary file to context save/restore the contents of the memory.  The test creates a binary file using `dd`, executes an application and saves the file back off.  The file is mapped into the DRAM memory using `mmap`.  We ensure the file changes post-run using SHA1 hashes.


test/backingstore/backingstore.c
```diff
@@ -0,0 +1,40 @@
+/*
+ * ex2.c
```
leekillough, line 2: `backingstore.c`

test/backingstore/rev-backingstore.py
```diff
@@ -0,0 +1,100 @@
+#
+# Copyright (C) 2017-2023 Tactical Computing Laboratories, LLC
+# All Rights Reserved
+# contact@tactcomplabs.com
+#
+# See LICENSE in the top level directory for licensing details
+#
+# rev-test-cache2.py
```
leekillough, line 8: `rev-backingstore.py`

test/backingstore/rev-backingstore.py
```diff
@@ -0,0 +1,100 @@
+#
```
leekillough, line 1: Does it need a separate `.py` file considering the new, more universal `rev-model-options-config.py`?   It looks complicated and specific for this test, but that would be a question I'd ask for every new test.

test/backingstore/backingstore.c
jleidel: fixed in 8f113a9

test/backingstore/rev-backingstore.py
jleidel: fixed in 8f113a9

test/backingstore/rev-backingstore.py
jleidel: for this test, its fairly vital to maintain a nonstandard config file

leekillough (approved):

PR #233, opened 2023-12-21
Change functions to pass RevInst by reference instead of by value

leekillough:
As `RevInst` has grown in size, it has become more expensive to pass by value. Running `big_loop` after this change shows negligible improvement but no slowdown.


jleidel:
the upstream `devel` version is set to 100

leekillough:
> the upstream `devel` version is set to 100

I see set it as 30. Git has not merged it when I merge `devel`. The lastest upstream CI failure timed out at 30.

jleidel:
> > the upstream `devel` version is set to 100
> 
> I see set it as 30. Git has not merged it when I merge `devel`. The lastest upstream CI failure timed out at 30.

you're right.  i was looking at the wrong test

rkabrick (approved):
LGTM

PR #235, opened 2023-12-23
Simplify compressed instructions implementation

leekillough:
This simplifies the implementation functions of compressed instructions, in most cases allowing them to be aliases of uncompressed instructions' implementations.

More `RevInst` fields are filled in, so that in most cases, jumping straight to an uncompressed instruction's implementation is enough. There are 4 cases where we must call different execution functions based on parameters, but with some work we could separate those out into separate table entries, and then all of the compressed instructions' implementation functions could be regular instruction implementation functions.


rkabrick (approved):
Everything in here looks good. Makes much more sense to move logic to the decoder. 

@donofrio is the resident expert on compressed insns (aside from you) in Rev so if he is cool with it then merge baby merge!

jleidel (approved):
good to merge

PR #237, opened 2024-01-03
Happy New Year!

rkabrick:
**Updated the copyright year**

- For any other repos you want to do this to I used the following command... MacOS uses BSD version of `sed` which is why there's some weirdness going on with the command

`LC_CTYPE=C find . -type f -exec sed -i '' -E 's/(\/\/ Copyright.*-)(2023)( Tactical Computing Laboratories, LLC)/\12024\3/' {} +`

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -1,7 +1,7 @@
 //
 // _Rev_Common_h_
 //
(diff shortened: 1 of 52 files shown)


jleidel (approved):
good to merge

Issue #223, opened 2023-12-13
fcvt.w.s does not support rounding modes


ukasz:
**Describe the bug**
fcvt.w.s instruction supports rounding modes, so we can decide whether round up or down.
On REV this field seems to be ignored resulting in rounding down always.

**To Reproduce**
Compile and run the following code:
```
//clang bug.c -o bug -march=rv64imafd
#define assert(x) if (!(x)) { asm(".dword 0x00000000"); }
int main() {
  float x = 3.14f;
  int result;

  asm volatile("fcvt.w.s %0, %1" : "=r"(result): "f"(x));
  assert(result == 3);

  asm volatile("fcvt.w.s %0, %1, rup" : "=r"(result): "f"(x));
  assert(result == 4);
  return 0;
}
```

** Setup **
Tested based on the following REV commit:
```
Author: Kenneth Griesser <kgriesser@tclcomplabs.com>
Date:   Tue Nov 28 09:46:21 2023 -0700

    bug test for simple struct with constructor
```
**Expected behavior**
Attached program should not fail on invalid instruction assert.

**Additional info**
Please not that there are also other instructions with rm field encoded that may suffer from similar behavior.

leekillough:
Rounding modes and FP exceptions is a work currently in progress.

leekillough:
Just to update: Rounding modes for all FP instructions, not just `fcvt`, is being developed.

On an IEEE754-compliant host, it is assumed that setting rounding modes for FP32 and FP64 will have the same effect on the host machine as on the RISC-V hardware, with a few _exceptions_ (pun intended) which can be handled in software (such as RISC-V's requirement that `0*Inf + z` and `Inf*0 + z` must raise invalid exceptions in `fmadd`-type instructions regardless of `z`). The rounding mode will be set before every RISC-V instruction is executed.

If a RISC-V instruction can raise FP exceptions, then the FP exceptions are cleared and made non-trapping before the simulation of the RISC-V instruction on the host machine, and any exceptions raised will be detected after the instruction simulation has executed, and will be reflected in the FCSR architectural register.

RISC-V instructions which cannot raise FP exceptions, such as load and store, `FMV`, or `copysign`-like instructions, are not affected by FP rounding flags.

Related: #67

leekillough:
https://github.com/tactcomplabs/rev/pull/238

ukasz:
[...]

Issue #67, opened 2023-06-21
set/read floating-point exceptions does not work


marcinkoss:
Compile and run below code:

```
#define assert(x) if (!(x)) { asm(".dword 0x00000000"); }

#include <fenv.h>
#include <stdint.h>

void test_except(uint64_t e) {
	feclearexcept(FE_ALL_EXCEPT);
	int r = feraiseexcept(e);
	assert (r == 0);
	assert(fetestexcept(e) > 0);
}

int main() {
	test_except(FE_DIVBYZERO);
	test_except(FE_INEXACT);
	test_except(FE_INVALID);
        test_except(FE_OVERFLOW);
	test_except(FE_UNDERFLOW);
	return 0;
}
```
When calling fetestexcept() we should get value of previously set exception. In runtime it seems that no exception is raised.
See also: 
https://en.cppreference.com/w/c/numeric/fenv/FE_exceptions
https://en.cppreference.com/w/cpp/numeric/fenv/fetestexcept

leekillough:
Please verify it has been fixed

leekillough:
`feraiseexcept()` returns nonzero if its argument is nonzero, which seems to be a library bug (I compiled with ` -frounding-math -ftrapping-math -fsignaling-nans`):
```
00000000000101fa <feraiseexcept>:
   101fa:       897d                    and     a0,a0,31
   101fc:       00152073                csrs    fflags,a0
   10200:       00a03533                snez    a0,a0
   10204:       8082                    ret
```
If we ignore the return value of `feraiseexcept()`, the test passes.

leekillough:
The return value of `feraiseexcept()` should always be zero, which it is in [current glibc sources](https://github.com/bminor/glibc/blob/21c3f4b5368686ade28d90d8c7d79c4c95c72c1b/sysdeps/riscv/rvf/fraiseexcpt.c). It seems like [Newlib has this bug](https://github.com/bminor/newlib/blob/master/newlib/libm/machine/riscv/feraiseexcept.c).

Closing this issue as completed.

PR #238, opened 2024-01-04
Add support for CSR instructions, FP exception flags, FP rounding modes, Performance Counters

leekillough:
This adds support for:
- CSR instructions (under the `Zicsr` extension)
- Floating-point rounding modes
- Floating-point exception flags
- Performance counters

The floating-point rounding mode and exceptions are simulated on the host machine running Rev, assuming an IEEE 754-compliant host.

Before a RISC-V FP instruction which can raise exceptions is executed:
- the floating-point environment (rounding mode, exception status) on the host is saved
- exceptions are blocked from trapping
- exception flags are cleared
- the host's FP rounding mode is changed to the RISC-V's FP rounding mode

Then the RISC-V FP instruction is executed on the host. Then the FP exceptions raised on the host are set in the RISC-V FP exception flags, and the host's FP state is restored to what it was before simulating the RISC-V FP instruction.

Only RISC-V FP instructions which can raise FP exceptions are dependent on the RISC-V FP rounding mode, so each instruction is marked as to whether it can raise FP exceptions with a `raisefpe` flag set to `true`, and only instructions with `raisefpe` `true` save and restore the host's FP environment, and transfer raised exceptions to the RISC-V FP CSR register.

There are 3 virtual FP CSR registers: `1`, `2`, and `3`.  `CSR 1` consists of the accumulated FP exception flags in the least-significant 5 bits; `CSR 2` consists of the FP dynamic rounding mode in the least-significant 3 bits; and `CSR 3` consists of the rounding mode followed by the exception flags in the least-significant 8 bits. Any of these 3 FP CSR registers can be manipulated with CSR instructions, and there are pseudoinstructions for them which are displayed during disassembly.

The `rdcycle`, `rdtime` and `rdinstret` CSR registers are simulated using Rev/SST counters. On RV32, these 64-bit performance counters have `rdcycleh`, `rdtimeh`, and `rdinstreth` CSR registers which return the high 32 bits.


leekillough:
Fixes https://github.com/tactcomplabs/rev/issues/223 and https://github.com/tactcomplabs/rev/issues/67

include/RevCPU.h
```diff
@@ -212,17 +212,17 @@ class RevCPU : public SST::Component {
...
-  RevMem*               Mem{};           ///< RevCPU: RISC-V main memory object
-  RevLoader*            Loader{};        ///< RevCPU: RISC-V loader
-  std::vector<RevCore*> Procs{};         ///< RevCPU: RISC-V processor objects
-  bool*                 Enabled{};       ///< RevCPU: Completion structure
+  unsigned numCores{};     ///< RevCPU: number of RISC-V cores
+  unsigned numHarts{};     ///< RevCPU: number of RISC-V cores
+  unsigned msgPerCycle{};  ///< RevCPU: number of messages to send per cycle
+  //  unsigned          RDMAPerCycle{};  ///< RevCPU: number of RDMA messages per cycle to inject into PAN network
+  //  unsigned          testStage{};     ///< RevCPU: controls the PAN Test harness staging
+  //  unsigned          testIters{};     ///< RevCPU: the number of message iters for each PAN Test
```
leekillough, line 220: These members are currently unused, which caused a warning, and is why I commented them out.

include/RevCPU.h
```diff
@@ -266,16 +266,16 @@ class RevCPU : public SST::Component {
...
   int address{};  ///< RevCPU: local network address
 
   unsigned fault_width{};  ///< RevCPU: the width (in bits) for target faults
-  int64_t  fault_range{};  ///< RevCPU: the range of cycles to inject the fault
-  int64_t  FaultCntr{};    ///< RevCPU: the fault counter
+  // int64_t  fault_range{};  ///< RevCPU: the range of cycles to inject the fault
+  int64_t FaultCntr{};  ///< RevCPU: the fault counter
 
-  uint64_t PrevAddr{};  ///< RevCPU: previous address for handling PAN messages
+  // uint64_t PrevAddr{};  ///< RevCPU: previous address for handling PAN messages
```
leekillough, line 278: More unused members.

include/RevCore.h
```diff
@@ -54,20 +53,22 @@
...
-    RevMem*                   Mem,
-    RevLoader*                Loader,
+    unsigned                  id,
+    RevOpts*                  opts,
+    unsigned                  numHarts,
+    RevMem*                   mem,
+    RevLoader*                loader,
     std::function<uint32_t()> GetNewThreadID,
-    SST::Output*              Output
+    SST::Output*              output
```
leekillough, line 71: It is not necessary to use alternative case between constructor parameters and class members. During member initialization, the class member is distinct from constructor parameters, so `id( id ), opts( opts ), numHarts( numHarts )` etc. can be used and the member is chosen on the outside of the initializer, and the constructor parameter is chosen on the inside of the initializer.

include/RevCore.h
```diff
@@ -316,6 +323,8 @@ class RevCore {
 
   std::bitset<_MAX_HARTS_> CoProcStallReq{};
 
+  uint64_t cycles{};  ///< RevCore: The number of cycles executed
+
```
leekillough, line 327: The RISC-V spec says that `rdcycle` should return the number of cycles executed by the _core_ so far, not the hart, which is why this is in `RevCore` instead of `RevRegFile`.

include/RevExt.h
```diff
@@ -34,7 +35,13 @@ struct RevExt {
...
 
   /// RevExt: standard destructor. virtual so that Extensions[i] can be deleted
-  virtual ~RevExt() = default;
+  virtual ~RevExt()                  = default;
+
+  // We do not allow copying, moving or assigning
+  RevExt( const RevExt& )            = delete;
+  RevExt( RevExt&& )                 = delete;
+  RevExt& operator=( const RevExt& ) = delete;
+  RevExt& operator=( RevExt&& )      = delete;
```
leekillough, line 44: [Rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three) to avoid surprises caused by implicit definitions of constructors and assignment operators.

include/RevFenv.h
```diff
@@ -11,69 +11,103 @@
...
+    case FRMode::RUP:  // Round Up (towards +Inf)
+      ret = std::fesetround( FE_UPWARD );
+      break;
+
+    case FRMode::RMM:  // Round to Nearest, ties to Max Magnitude
+#ifdef FE_TONEARESTFROMZERO
+      ret = std::fesetround( FE_TONEARESTFROMZERO );
+#else
+      output->fatal( CALL_INFO, -1, "Error: Round to nearest Max Magnitude not implemented at PC = 0x%" PRIx64 "\n", R->GetPC() );
+#endif
```
leekillough, line 73: Round to nearest Max Magnitude, also called roundTiesToAway, is optional on binary radix floating-point systems, and is only mandatory on decimal radix floating-point systems, because of its use in financial calculations.

A C2X proposal added the `FE_TONEARESTFROMZERO` as an optional macro defining the rounding mode. If this macro is supported on the host environment, we can support RISC-V Round to Nearest Max Magnitude; otherwise it's not implemented.

include/RevRegFile.h
```diff
@@ -159,8 +160,10 @@ class RevRegFile {
 #endif
 
 public:
-  // Constructor which takes a RevFeature
-  explicit RevRegFile( const RevFeature* feature ) : IsRV32( feature->IsRV32() ), HasD( feature->HasD() ) {}
+  // Constructor which takes a RevCore to indicate its hart's parent core
+  // Template is to prevent circular dependencies by not requiring RevCore to be a complete type now
+  template<typename T, typename = std::enable_if_t<std::is_same_v<T, RevCore>>>
+  explicit RevRegFile( T* core ) : Core( core ), IsRV32( core->GetRevFeature()->IsRV32() ), HasD( core->GetRevFeature()->HasD() ) {}
```
leekillough, line 166: We store the parent `RevCore` of the Hart in `RevRegFile`, so that Core performance counters can be obtained by any Hart.

include/RevRegFile.h
```diff
@@ -335,6 +338,42 @@ class RevRegFile {
...
   }
 
+  /// Get a CSR register
+  template<typename T>
+  T GetCSR( size_t csr ) const {
+    // We store fcsr separately from the global CSR
+    switch( csr ) {
+    case 1: return static_cast<uint32_t>( fcsr ) >> 0 & 0b00011111u; break;
+    case 2: return static_cast<uint32_t>( fcsr ) >> 5 & 0b00000111u; break;
+    case 3: return static_cast<uint32_t>( fcsr ) >> 0 & 0b11111111u; break;
```
leekillough, line 348: There are 3 virtual FP CSR registers -- the exception flags, the rounding mode, and both combined.

include/RevThread.h
```diff
@@ -112,17 +112,17 @@ class RevThread {
...
-  std::unique_ptr<RevVirtRegState>    VirtRegState{};               // Register file
-  std::unordered_set<uint32_t>        ChildrenIDs{};                // Child thread IDs
-  std::unordered_set<int>             fildes{ 0, 1, 2 };            // Default file descriptors
+  uint32_t ID{};        // Thread ID
+  uint32_t ParentID{};  // Parent ID
+  //  uint64_t                        StackPtr{};                   // Initial stack pointer
+  //  uint64_t                        FirstPC{};                    // Initial PC
+  std::shared_ptr<RevMem::MemSegment> ThreadMem{};  // TLS and Stack memory
+  RevFeature*                         Feature{};    // Feature set for this thread
+  //  uint64_t                        ThreadPtr{};                  // Thread pointer
```
leekillough, line 121: More unused members.

include/insns/RV32D.h
```diff
@@ -85,11 +85,7 @@ class RV32D : public RevExt {
   }
 
   static bool fclassd( RevFeature* F, RevRegFile* R, RevMem* M, const RevInst& Inst ) {
-    double   fp64 = R->GetFP<double>( Inst.rs1 );
-    uint64_t i64;
-    memcpy( &i64, &fp64, sizeof( i64 ) );
-    bool quietNaN = ( i64 & uint64_t{ 1 } << 51 ) != 0;
-    R->SetX( Inst.rd, fclass( fp64, quietNaN ) );
+    R->SetX( Inst.rd, fclass( R->GetFP<double>( Inst.rs1 ) ) );
```
leekillough, line 88: `fclass()` is complete now, and doesn't need help about `quietNaN`.

include/insns/RV32D.h
```diff
@@ -144,22 +140,22 @@ class RV32D : public RevExt {
...
     { Rev32DInstDefaults().SetMnemonic("fcvt.w.d %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b1100001).SetImplFunc(fcvtwd ).SetrdClass (RevRegClass::RegGPR).SetfpcvtOp(0b00).Setrs2Class(RevRegClass::RegUNKNOWN) },
     { Rev32DInstDefaults().SetMnemonic("fcvt.wu.d %rd, %rs1"           ).SetFunct3(0b000).SetFunct2or7(0b1100001).SetImplFunc(fcvtwud).SetrdClass (RevRegClass::RegGPR).SetfpcvtOp(0b01).Setrs2Class(RevRegClass::RegUNKNOWN) },
-    { Rev32DInstDefaults().SetMnemonic("fcvt.d.w %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdw ).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b00).Setrs2Class(RevRegClass::RegUNKNOWN) },
-    { Rev32DInstDefaults().SetMnemonic("fcvt.d.wu %rd, %rs1"           ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdwu).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b01).Setrs2Class(RevRegClass::RegUNKNOWN) },
+    { Rev32DInstDefaults().SetMnemonic("fcvt.d.w %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdw ).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b00).Setrs2Class(RevRegClass::RegUNKNOWN).SetRaiseFPE( false ) },
```
leekillough, line 157: A conversion from `int32_t` to `double` is lossless, and is unaffected by rounding, and can never raise exceptions.

include/insns/RV32D.h
```diff
@@ -144,22 +140,22 @@ class RV32D : public RevExt {
...
+    { Rev32DInstDefaults().SetMnemonic("fsgnjn.d %rd, %rs1, %rs2"      ).SetFunct3(0b001).SetFunct2or7(0b0010001).SetImplFunc(fsgnjnd).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) != DECODE_RS2( Inst ); } ).SetRaiseFPE( false ) },
+    { Rev32DInstDefaults().SetMnemonic("fneg.d %rd, %rs"               ).SetFunct3(0b001).SetFunct2or7(0b0010001).SetImplFunc(fsgnjnd).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == DECODE_RS2( Inst ); } ).SetRaiseFPE( false ) },
+    { Rev32DInstDefaults().SetMnemonic("fsgnjx.d %rd, %rs1, %rs2"      ).SetFunct3(0b010).SetFunct2or7(0b0010001).SetImplFunc(fsgnjxd).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) != DECODE_RS2( Inst ); } ).SetRaiseFPE( false ) },
+    { Rev32DInstDefaults().SetMnemonic("fabs.d %rd, %rs"               ).SetFunct3(0b010).SetFunct2or7(0b0010001).SetImplFunc(fsgnjxd).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == DECODE_RS2( Inst ); } ).SetRaiseFPE( false ) },
```
leekillough, line 148: The `copysign()`-like instructions cannot raise FP exceptions, and are not affected by rounding mode.

include/insns/RV32D.h
```diff
@@ -144,22 +140,22 @@ class RV32D : public RevExt {
...
     { Rev32DInstDefaults().SetMnemonic("fcvt.d.s %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b0100001).SetImplFunc(fcvtds ).Setrs2Class(RevRegClass::RegUNKNOWN).SetfpcvtOp(0b00) },
     { Rev32DInstDefaults().SetMnemonic("feq.d %rd, %rs1, %rs2"         ).SetFunct3(0b010).SetFunct2or7(0b1010001).SetImplFunc(feqd   ).SetrdClass (RevRegClass::RegGPR) },
     { Rev32DInstDefaults().SetMnemonic("flt.d %rd, %rs1, %rs2"         ).SetFunct3(0b001).SetFunct2or7(0b1010001).SetImplFunc(fltd   ).SetrdClass (RevRegClass::RegGPR) },
     { Rev32DInstDefaults().SetMnemonic("fle.d %rd, %rs1, %rs2"         ).SetFunct3(0b000).SetFunct2or7(0b1010001).SetImplFunc(fled   ).SetrdClass (RevRegClass::RegGPR) },
-    { Rev32DInstDefaults().SetMnemonic("fclass.d %rd, %rs1"            ).SetFunct3(0b001).SetFunct2or7(0b1110001).SetImplFunc(fclassd).SetrdClass (RevRegClass::RegGPR).Setrs2Class(RevRegClass::RegUNKNOWN) },
+    { Rev32DInstDefaults().SetMnemonic("fclass.d %rd, %rs1"            ).SetFunct3(0b001).SetFunct2or7(0b1110001).SetImplFunc(fclassd).SetrdClass (RevRegClass::RegGPR).Setrs2Class(RevRegClass::RegUNKNOWN).SetRaiseFPE( false ) },
```
leekillough, line 154: The `fclass` instruction cannot raise exceptions, and is not affected by rounding mode.

include/insns/RV32D.h
```diff
@@ -144,22 +140,22 @@ class RV32D : public RevExt {
...
     { Rev32DInstDefaults().SetMnemonic("fcvt.wu.d %rd, %rs1"           ).SetFunct3(0b000).SetFunct2or7(0b1100001).SetImplFunc(fcvtwud).SetrdClass (RevRegClass::RegGPR).SetfpcvtOp(0b01).Setrs2Class(RevRegClass::RegUNKNOWN) },
-    { Rev32DInstDefaults().SetMnemonic("fcvt.d.w %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdw ).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b00).Setrs2Class(RevRegClass::RegUNKNOWN) },
-    { Rev32DInstDefaults().SetMnemonic("fcvt.d.wu %rd, %rs1"           ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdwu).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b01).Setrs2Class(RevRegClass::RegUNKNOWN) },
+    { Rev32DInstDefaults().SetMnemonic("fcvt.d.w %rd, %rs1"            ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdw ).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b00).Setrs2Class(RevRegClass::RegUNKNOWN).SetRaiseFPE( false ) },
+    { Rev32DInstDefaults().SetMnemonic("fcvt.d.wu %rd, %rs1"           ).SetFunct3(0b000).SetFunct2or7(0b1101001).SetImplFunc(fcvtdwu).Setrs1Class(RevRegClass::RegGPR).SetfpcvtOp(0b01).Setrs2Class(RevRegClass::RegUNKNOWN).SetRaiseFPE( false ) },
```
leekillough, line 158: A conversion from `uint32_t` to `double` is lossless, and is unaffected by rounding, and can never raise exceptions.

include/insns/RV32I.h
```diff
@@ -129,14 +129,6 @@ class RV32I : public RevExt {
...
   static constexpr auto& ebreak    = nop;
 
-  // Unimplemented Zicsr extension instructions
-  static constexpr auto& csrrw     = nop;
-  static constexpr auto& csrrs     = nop;
-  static constexpr auto& csrrc     = nop;
-  static constexpr auto& csrrwi    = nop;
-  static constexpr auto& csrrsi    = nop;
-  static constexpr auto& csrrci    = nop;
-
```
leekillough, line 139: These implementation functions have been moved to `Zicsr.h`.

include/insns/Zicsr.h
```diff
@@ -0,0 +1,184 @@
...
+      SetFormat( RVTypeI );
+    }
+  };
+
+  // clang-format off
+  std::vector<RevInstEntry> ZicsrTable = {
+    { RevZicsrInstDefaults().SetMnemonic( "csrw %csr, %rs1"        ).SetFunct3( 0b001 ).SetImplFunc( csrrw      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RD( Inst ) == 0; } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "csrrw %csr, %rd, %rs1"  ).SetFunct3( 0b001 ).SetImplFunc( csrrw      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RD( Inst ) != 0 && DECODE_IMM12( Inst ) != 0x1 && DECODE_IMM12( Inst ) != 0x2; } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "fsflags %rs"            ).SetFunct3( 0b001 ).SetImplFunc( csrrw      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RD( Inst ) != 0 && DECODE_IMM12( Inst ) == 0x1; } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "fsrm %rs"               ).SetFunct3( 0b001 ).SetImplFunc( csrrw      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RD( Inst ) != 0 && DECODE_IMM12( Inst ) == 0x2; } ) },
```
leekillough, line 162: `fsflags` and `fsrm` are pseudoinstructions which are `csrrw` instructions underneath.

include/insns/Zicsr.h
```diff
@@ -0,0 +1,184 @@
...
+
+    { RevZicsrInstDefaults().SetMnemonic( "csrrs %rd, %csr, %rs1"  ).SetFunct3( 0b010 ).SetImplFunc( csrrs      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) != 0; } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "csrr %rd, %csr"         ).SetFunct3( 0b010 ).SetImplFunc( csrrs      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == 0 && DECODE_IMM12( Inst ) != 0x1 && DECODE_IMM12( Inst ) != 0x2 && [](auto imm){ return imm < 0xc80 || imm > 0xc82; }( DECODE_IMM12( Inst ) | 0x80 ); } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "frflags %rd"            ).SetFunct3( 0b010 ).SetImplFunc( csrrs      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == 0 && DECODE_IMM12( Inst ) == 0x1; } ) },
+    { RevZicsrInstDefaults().SetMnemonic( "frrm %rd"               ).SetFunct3( 0b010 ).SetImplFunc( csrrs      ).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == 0 && DECODE_IMM12( Inst ) == 0x2; } ) },
+
```
leekillough, line 168: `frflags` and `frrm` are pseudoinstructions which are `csrrs` instructions underneath.

src/RevCPU.cc
```diff
@@ -861,7 +861,7 @@ void RevCPU::HandleThreadStateChangesForProc( uint32_t ProcID ) {
 
 void RevCPU::InitMainThread( uint32_t MainThreadID, const uint64_t StartAddr ) {
   // @Lee: Is there a better way to get the feature info?
-  std::unique_ptr<RevRegFile> MainThreadRegState = std::make_unique<RevRegFile>( Procs[0]->GetRevFeature() );
+  std::unique_ptr<RevRegFile> MainThreadRegState = std::make_unique<RevRegFile>( Procs[0] );
```
leekillough, line 864: The `RevRegFile` constructor has been changed to take the parent `RevCore`, from which it can get the `RevFeature` and other fields.

src/RevCore.cc
```diff
@@ -17,27 +17,26 @@ namespace SST::RevCPU {
...
+  RevLoader*                loader,
   std::function<uint32_t()> GetNewTID,
-  SST::Output*              Output
+  SST::Output*              output
 )
-  : Halted( false ), Stalled( false ), SingleStep( false ), CrackFault( false ), ALUFault( false ), fault_width( 0 ), id( Id ),
-    HartToDecodeID( 0 ), HartToExecID( 0 ), numHarts( NumHarts ), opts( Opts ), mem( Mem ), coProc( nullptr ), loader( Loader ),
-    GetNewThreadID( std::move( GetNewTID ) ), output( Output ), feature( nullptr ), sfetch( nullptr ), Tracer( nullptr ) {
+  : id( id ), numHarts( numHarts ), opts( opts ), mem( mem ), loader( loader ), GetNewThreadID( std::move( GetNewTID ) ),
+    output( output ) {
```
leekillough, line 29: Most of the members are initialized to `0` with default initializers now, so there is no need to crowd the constructor initialization list.

leekillough: See previous comment about capitalization of constructor parameters being unnecessary to distinguish them from members.

src/RevCore.cc
```diff
@@ -994,6 +998,9 @@ RevInst RevCore::DecodeRInst( uint32_t Inst, unsigned Entry ) const {
   DInst.funct3    = InstTable[Entry].funct3;
   DInst.funct2or7 = InstTable[Entry].funct2or7;
 
+  // Whether the instruction raises floating-point exceptions
+  DInst.raisefpe  = InstTable[Entry].raisefpe;
+
```
leekillough, line 1003: Only `R`- and `R4`-format instructions need to copy the `raisefpe` field.

src/RevCore.cc
```diff
@@ -1633,7 +1642,9 @@ void RevCore::MarkLoadComplete( const MemReq& req ) {
 bool RevCore::ClockTick( SST::Cycle_t currentCycle ) {
   RevInst Inst;
   bool    rtn = false;
-  Stats.totalCycles++;
+  ++Stats.totalCycles;
+  ++cycles;
+  currentSimCycle = currentCycle;
```
leekillough, line 1647: For `rdcycle` and `rdtime`.

src/RevCore.cc
```diff
@@ -1803,12 +1813,11 @@ bool RevCore::ClockTick( SST::Cycle_t currentCycle ) {
         ExecPC
       );
 #endif
-      Stats.retired++;
+      ++Stats.retired;
+      ++RegFile->InstRet;
```
leekillough, line 1817: For `rdinstret`.

src/RevExt.cc
```diff
@@ -59,18 +29,27 @@ bool RevExt::Execute( unsigned Inst, const RevInst& payload, uint16_t HartID, Re
...
+    // exceptions cleared and the rounding mode set according to the encoding
+    // and frm register. The FP exceptions which occur on the host are stored
+    // in FCSR when saved_fenv is destroyed.
+    RevFenv saved_fenv( regFile, payload.rm, output );
+
+    // Execute the instruction
+    ret = func( feature, regFile, mem, payload );
+
+    // Fall-through destroys saved_fenv, setting the FCSR's fflags and
+    // restoring the host's FP environment
```
leekillough, line 52: **This is the main mechanism for emulating FP rounding modes and exceptions on the host machine.**

test/CMakeLists.txt
```diff
@@ -161,7 +161,8 @@ macro(add_rev_test test_name test_dir timeout labels)
     # Custom target for running the specified script
     add_custom_target(run_${test_name_lower}_script
       COMMAND ${CMAKE_COMMAND} -E echo "Executing script ${optional_script}"
-      COMMAND cd ${CMAKE_CURRENT_SOURCE_DIR}/${test_name_lower} && bash ${optional_script}
+      COMMAND bash ${optional_script}
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${test_dir_lower}
```
leekillough, line 186: `WORKING_DIRECTORY` should be set instead of using `cd` to change the directory, and the directory should be `${test_dir_lower}` instead of `${test_name_lower}`.

test/CMakeLists.txt
```diff
@@ -203,7 +204,7 @@ add_rev_test(DIVW divw 30 "all;memh;rv64")
 add_rev_test(DIVW2 divw2 30 "all;memh;rv64")
 add_rev_test(X0 x0 30 "all;memh;rv64")
 add_rev_test(ARGC argc 30 "all;memh;rv64;loader;" SCRIPT "run_argc.sh")
-add_rev_test(ARGC_SHORT argc 30 "all;memh;rv64;loader;" SCRIPT "run_argc_short.sh")
+add_rev_test(ARGC_SHORT argc_short 30 "all;memh;rv64;loader;" SCRIPT "run_argc_short.sh")
```
leekillough: `argc_short` used the wrong `test_dir` of `argc` instead of `argc_short`.

test/CMakeLists.txt
```diff
@@ -212,6 +213,9 @@ add_rev_test(ZICBOM zicbom 45 "all;memh;rv64" SCRIPT "run_zicbom.sh")
 add_rev_test(BACKINGSTORE backingstore 100 "all;rv64" SCRIPT "run_backingstore.sh")
 add_rev_test(MEM_DUMP mem_dump 10 "all;rv64" SCRIPT "run_mem_dump.sh")
 add_rev_test(LWSP lwsp 30 "all;memh;rv64")
+add_rev_test(RDCYCLE rdcycle 5 "all;rv64")
+add_rev_test(RDTIME rdtime 5 "all;rv64")
+add_rev_test(RDINSTRET rdinstret 5 "all;memh;rv64")
```
leekillough: `rdcycle` and `rdtime` differences measure an indeterminately long time to execute a fixed number of instructions because of stalls, so MemH is disabled for them. But `rdinstret` counts retired instructions, not the time it took, so it works for both RevMem and MemH.

test/cxx/stl/unordered_map/Makefile
```diff
@@ -14,9 +14,6 @@
 
 EXAMPLE=unordered_map
 CXX=${RVCXX}
-CXX=riscv64-unknown-elf-g++  # TODO: Remove this line when "frflags" instruction is implemented (for floor function)
-
-#ARCH=rv64g
```
leekillough, line 19: We no longer need to hard-code the compiler, since `floor()` works now.

test/fenv/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	CXX_OPTS += -ffp-model=strict
+else
+	CXX_OPTS += -fsignaling-nans
+endif
+
+CXX_OPTS += -frounding-math -ftrapping-math -Werror=double-promotion -Werror=conversion -ffp-contract=on
+
+buildtests: fenv_test.makefile
+	make -j4 -f fenv_test.makefile CXX="$(CXX)" CXX_OPTS="$(CXX_OPTS)" ARCH="$(ARCH)" ABI="$(ABI)" TONEAREST UPWARD DOWNWARD TOWARDZERO
+
```
leekillough, line 29: The default target builds the tests without running them.

test/fenv/Makefile
```diff
@@ -0,0 +1,50 @@
...
+runtests: fenv_test.makefile
+	make -j4 -f fenv_test.makefile CXX="$(CXX)" CXX_OPTS="$(CXX_OPTS)" ARCH="$(ARCH)" ABI="$(ABI)" RUN_TONEAREST RUN_UPWARD RUN_DOWNWARD RUN_TOWARDZERO
+	echo "Simulation is complete"
```
leekillough, line 32: The `runtests` target, which is built by `run_fenv.sh`, runs the tests, and, only if they succeed, prints the success message `"Simulation is complete"`.

test/fenv/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+fenv_test.makefile: fenv_gentests
+	printf '%%.exe: %%.cc\n\t$$(CXX) $$(CXX_OPTS) -static -mcmodel=medany -march=$$(ARCH) -mabi=$$(ABI) -lm $$< -o $$@\n' > fenv_test.makefile
+	./fenv_gentests TONEAREST  >> fenv_test.makefile
+	./fenv_gentests UPWARD     >> fenv_test.makefile
+	./fenv_gentests DOWNWARD   >> fenv_test.makefile
+	./fenv_gentests TOWARDZERO >> fenv_test.makefile
```
leekillough, line 39: The tests and their Makefile are automatically generated.

test/fenv/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+# build test generator on local host
+fenv_gentests: fenv_test.h fenv_gentests.cc
+	g++ $(CXX_OPTS) fenv_gentests.cc -o $@
```
leekillough, line 43: An `.exe` extension is not used for `fenv_gentests`, since it runs on the host machine and is not usually a RISC-V executable. This prevents errors during disassembly.

test/fenv/fenv_gentests.cc
```diff
@@ -0,0 +1,280 @@
...
+ * All Rights Reserved
+ * contact@tactcomplabs.com
+ *
+ * See LICENSE in the top level directory for licensing details
+ *
+ */
+
+#include "fenv_test.h"
+
+constexpr unsigned maxtests_per_file = 500;
```
leekillough, line 16: A certain maximum number of tests per source file is generated. All 4000+ tests could be in one file, but it would be slower to compile and run. I tried different numbers of tests per file, and `500` was around optimal in terms of test runtime.

test/fenv/fenv_gentests.cc
```diff
@@ -0,0 +1,280 @@
...
+  -1.0f,
+  1.5f,
+  -1.5f,
+  std::numeric_limits<FP>::quiet_NaN(),
+  std::numeric_limits<FP>::signaling_NaN(),
+  std::numeric_limits<FP>::infinity(),
+  -std::numeric_limits<FP>::infinity(),
+  std::numeric_limits<FP>::lowest(),
+  std::numeric_limits<FP>::max(),
+};
```
leekillough, line 148: These are selected floating-point values which, when combined with other FP operands, handle most special return values and FP exceptions. It was originally around 5 values larger, but I reduced it to reduce test time.

test/fenv/fenv_gentests.cc
```diff
@@ -0,0 +1,280 @@
...
+  FP( -0.1 ),
+  FP( 0.5 ),
+  FP( -0.5 ),
+  FP( 0.9 ),
+  FP( -0.9 ),
+  std::numeric_limits<FP>::quiet_NaN(),
+  std::numeric_limits<FP>::signaling_NaN(),
+  std::numeric_limits<FP>::infinity(),
+  -std::numeric_limits<FP>::infinity(),
+};
```
leekillough, line 186: These are special FP values for when converting from FP to integer, with controlled rounding.

test/fenv/run_fenv_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sst --add-lib-path=../../build/src/ ../rev-model-options-config.py -- --program="$1" --verbose=0 | awk 'BEGIN {pass=0} /Simulation is complete/ {pass=1;next} {print} END {exit !pass}'
```
leekillough, line 3: The AWK script detects and filters out `Simulation is complete`, and returns a nonzero exit code if `Simulation is complete` was not seen, which becomes the exit code of the script. If it fails, the parallel `make` fails, and `Simulation is complete` is never printed by `make`.

test/rdcycle/rdcycle.c
```diff
@@ -0,0 +1,52 @@
...
+#define NOP512  NOP256 NOP256
+#define NOP1024 NOP512 NOP512
+
+int main( int argc, char** argv ) {
+  size_t cycle1, cycle2;
+
+  asm volatile( " rdcycle %0" : "=r"( cycle1 ) );
+
+  // 1024 nops produces around 1024 cycles
+  NOP1024;
```
leekillough: `1024` `nop` instructions are placed between two performance counter samplings, and the difference should be between `1024` and `1026`. For `rdcycle` and `rdtime`, MemH is too slow and unpredictable. But for `rdinstret`, MemH is just as safe as RevMem.

test/csr/csr.cc
```diff
@@ -0,0 +1,259 @@
...
+      if constexpr( ARG == 0 ) {
+        asm volatile( "csrrc %0, %1, zero" : "=r"( old ) : "I"( CSR ) );
+      } else {
+        size_t arg = ARG;
+        asm volatile( "csrrc %0, %1, %2" : "=r"( old ) : "I"( CSR ), "r"( arg ) );
+      }
+    }
+    return old;
+  }
+};
```
leekillough, line 99: When the operand to the CSR instruction is a register, we store the value in a register and pass that register to the CSR instruction. If the value is `0`, we use the `x0` register instead of another register which happens to be `0`. When `rd == x0`, we explicitly code `rd` as `x0`.

test/csr/csr.cc
```diff
@@ -0,0 +1,259 @@
...
+  size_t operator()() {
+    size_t old = 0;
+    if constexpr( RD0 ) {
+      asm volatile( "csrrci zero, %0, %1" ::"I"( CSR ), "K"( ARG ) );
+    } else {
+      asm volatile( "csrrci %0, %1, %2" : "=r"( old ) : "I"( CSR ), "K"( ARG ) );
+    }
+    return old;
+  }
+};
```
leekillough, line 138: When the operand to the CSR instruction is an immediate, we pass it using the RISC-V inline assembly modifier `K`, which indicates a 5-bit zero-extended immediate instead of a register. Like the register operand CSR instructions, when `rd == x0`  we explicitly code `x0` instead of using a local register.

test/csr/csr.cc
```diff
@@ -0,0 +1,259 @@
...
+    CSRTest3<KIND, CSR, ~size_t{ 4 }, RD0>();
+    CSRTest3<KIND, CSR, ~size_t{ 8 }, RD0>();
+    CSRTest3<KIND, CSR, ~size_t{ 16 }, RD0>();
+    CSRTest3<KIND, CSR, ~size_t{ 31 }, RD0>();
+    CSRTest3<KIND, CSR, +size_t{ 256 }, RD0>();
+    CSRTest3<KIND, CSR, ~size_t{ 256 }, RD0>();
+    CSRTest3<KIND, CSR, +size_t{ 1 } << sizeof( size_t ) * 8 - 1, RD0>();
+    CSRTest3<KIND, CSR, ~size_t{ 1 } << sizeof( size_t ) * 8 - 1, RD0>();
+  }
+}
```
leekillough, line 242: When the operand is a register, we use values which do not fit in a zero-extended 5-bit immediate value. When the operand is an immediate, we can only use values `0` - `31`.

test/fenv/fenv_gentests.cc
```diff
@@ -0,0 +1,315 @@
...
+  strcat( test_src, R"( %0, %1" : "=r"(res) : "f"(x) ); return res; })" );
+
+  using INT_FUNC1 = std::pair<INT ( * )( FP ), const char*>[];
+  for( auto oper_pair : INT_FUNC1{
+         {[]( volatile auto x ) { return to_int<INT>( x ); }, test_src},
+  } ) {
+    for( auto x : special_fcvt_values<FP, INT> ) {
+      generate_test( oper_pair, x );
+    }
+  }
```
leekillough, line 230: For FP -> Integer conversions, we use the `fcvt` instruction corresponding to the source and destination types, and call it in inline assembly.

leekillough:
The changes are complete and ready for review. Any additional commits are purely cosmetic, in response to review comments, or improve test coverage / test accuracy.

This PR represents over 6 months of work, and substantially improves the Rev ecosystem. It fixes two outstanding issues.

include/RevThread.h
```diff
@@ -112,17 +112,17 @@ class RevThread {
...
-  std::shared_ptr<RevMem::MemSegment> ThreadMem{};                  // TLS and Stack memory
-  RevFeature*                         Feature{};                    // Feature set for this thread
-  uint64_t                            ThreadPtr{};                  // Thread pointer
-  ThreadState                         State{ ThreadState::START };  // Thread state
-  std::unique_ptr<RevVirtRegState>    VirtRegState{};               // Register file
-  std::unordered_set<uint32_t>        ChildrenIDs{};                // Child thread IDs
-  std::unordered_set<int>             fildes{ 0, 1, 2 };            // Default file descriptors
+  uint32_t ID{};        // Thread ID
+  uint32_t ParentID{};  // Parent ID
+  //  uint64_t                        StackPtr{};                   // Initial stack pointer
```
rkabrick, line 117: Should these be deleted?

rkabrick (approved):
This all looks great. I don't think I saw the option here but I would say the only thing I'd recommend adding is a CMake option for something like `-DENABLE_FP_TESTING` or `-DBUILD_FP_TESTING` something like that.

leekillough:
> This all looks great. I don't think I saw the option here but I would say the only thing I'd recommend adding is a CMake option for something like `-DENABLE_FP_TESTING` or `-DBUILD_FP_TESTING` something like that.

The testing time is not significantly increased by this PR, and it would be a lot of work and would add unnecessary complications to testing: "Oh, you have to define this macro if you want the floating-point to be tested."

rkabrick:
What would be a lot of work? I would at the very least include these tests only when `-DBUILD_ASM_TESTING` is passed if you don't want to include a separate cmake arg

leekillough:
> What would be a lot of work?

Adding a macro to enable floating-point tests. Then it would need to be added to the CI and anywhere else testing is done.

Right now the CI takes 20 minutes to run the full Rev suite, including the floating-point tests. That should be considered acceptable.

MemH is disabled for the floating-point tests, since it is much slower than RevMem, and the floating-point tests are not concerned with memory.

The floating-point tests also have their own form of 4-way parallelism with parallel `make`.

rkabrick:
> Adding a macro to enable floating-point tests.

That can be done in ~2 lines of cmake.

I would say two options that are better than what it is now: 
1. Create a separate `ENABLE_FP_TESTING` that defaults to ON (eliminates any changes required to the jenkins builds
2. Only enable the tests when `BUILD_ASM_TESTING` is set to on. This macro already exists and these tests are assembly based tests I believe.

leekillough:
I don't even know why there is a `BUILD_ASM_TESTING` macro. It should always be turned on. This is after all, a simulator. Don't you want it to exercise as many instructions as possible (even in a pseudo-RISC architecture)?

Same with the floating-point tests. They should always be turned on.

If someone wants to focus on only a certain test, then they can `cd` to that test directory and build/run it manually. That's what I did when developing the test.

leekillough:
On my workstation, it takes much less than 20 minutes to run the full suite, even with the floating-point tests. The 20 minute time is on `pennywise`.

If someone has a workstation/server comparable to mine, I doubt they are going to insist on disabling the floating-point tests.

We should not be starting from the full suite and subtracting from it; we should be starting from individual tests and adding them to the full suite.
Why don't we have an option to disable `big_loop`? Or a half-dozen other long tests?

leekillough:
I strongly object to disabling the floating-point tests. They are necessary for validating the Rev floating-point ecosystem, and they validate 6 months of work which cannot simply be ignored because someone wants the tests to run faster. I spent 2 days making them run as fast as they do now.

Even with the floating-point tests enabled, the full Rev suite takes 20 minutes or less. They should be always included.

If someone wants to run an individual test manually, they can. There is no need to add macros to disable tests.

kpgriesser:
Suggestion: as we grow the test base I would recommend adding the infrastructure for levels of testing that all tests to be easily tagged for 'short', 'long', or 'really really long' regression suites (level 0, 1 , 2... ). This is always a point of contention on projects when the test base and testing time starts grow so it may be wise to agree and put the infrastructure in up front.

To not derail this PR, I'd be glad to start a separate one to discuss/resolve.

include/RevThread.h
leekillough: > Should these be deleted?

If you're sure these unused variables won't be used in the future (e.g. @donofrio 's Pipeline), we can delete them. I left them around as comments in case they were going to be used later.

donofrio (approved):
Changes look good - this is a major improvement to rev's correctness and enhances the usefulness of the simulator.  Agree that we need to create levels of regression testing. A short suite that can be run frequently while actively developing as a spot-check and then a longer suite that can be run prior to submitting a PR and as the regular CI suite

leekillough:
`TEST_LEVEL` can be set to `1`, `2` or `3`.

`1` is a very fast test which runs in a minute or less
`2` adds the assembly `isa` tests and all tests before this PR.
`3` runs all tests, and is the default. This includes floating-point tests in this PR which exercise all rounding modes and thousands of combinations of arguments, including special values like `0.0`, `-0.0`, `quietNaN`, `signalingNaN`, and `+/-Inf`, making sure the returned answers and the exceptions raised are correct.

include/RevCPU.h
kpgriesser: good - I thought I had pruned all this out. The PAN model is in a different branch

test/fenv/run_fenv_test.sh
kpgriesser: I like this. For future thought, if we had a way to control an exit code from SST itself we could avoid all this post-processing of files and strings which gets really expensive if we're generating large numbers of log files. I may bring this up again in the sst users group.

leekillough: I tried to make the exit code of the script be the sole determinant of whether the test passed or failed, but it introduced too many problems, so I had to fall back on this expected regex in output. The AWK script detects it in tests and turns it into an exit code for a build; then the parallel make passes or fails if all of its parallel targets pass. If all pass, the parallel make prints the message which `ctest` expects, and the test passes.

kpgriesser (approved):
Looking forward to enabling more CSRs with this.  If we have interrupt support and a timer CSR that could trigger an interrupt we would be able to enable programmatic control of sleeping/waking the CPU during polling loops which has been an unfortunate performance  issue.

For future thought:  A floating implementation which allows modeling pipelined delays of different floating point instructions could be very useful for modeling. This would be analogous to our simpleMem and MemH options.

PR #239, opened 2024-01-06
Simplify initialization of options at startup

leekillough:
This is a low-priority PR, which can wait until after releases. It:
1. Makes the handling of options more data-driven with less code duplication
2. Passes option vectors by reference instead of by value, to reduce function call overhead
3. Uses `std::nothrow` as indicated by a previous comment of mine


rkabrick (approved):
This is awesome! Merge away

PR #240, opened 2024-01-06
Make Instruction Table Building Simpler

leekillough:
### This is a low-priority PR, which can wait until after releases.

It simplifies the building of the instruction tables, and reduces the source code size:
1. It merges the `RevInstDefaults` and `RevInstEntry` structs, which were largely duplicated definitions, but could have different member initializations.
2. It replaces the `RevInstEntryBuilder` class -- which had an `RevInstEntry InstEntry` member -- with simply the `RevInstEntry` base class, by moving the `.Set()` functions to the `RevInstEntry` class, so that anything derived from the `RevInstEntry` base class can call `.Set()`.
3. Instead of shadowing base class data members and their default definitions in a derived class, it calls `Set()` functions in the derived class's constructor. The previous method could have problems unless accessed through the very most derived class and could cause strange bugs (`static` vs non-`static` data; `constexpr` vs non-`constexpr`, etc.); now all data lives in the base class and accessing data members is unambiguous. There is no slicing or shadowing of data members.
4. The old idiom of `{ RevInstEntryBuilder<RevInstDefaults>().Set...().Set...().InstEntry }` for generating table entries is replaced with the simpler but not too different `{ RevInstDefaults().Set...().Set...() }` (without an `.InstEntry` at the end).
> The reason this works is that the tables store a `RevInstEntry`, which is the base class. `RevInstDefaults` is the same as `RevInstEntry` now. `RevInstDefaults().Set...().Set...()`  or `RevInstSomeDerived().Set...().Set...()` creates a `RevInstEntry` or a derived class of it, and calls several `Set()` modifiers, and uses the last returned value, which can be stored in a `RevInstEntry` since it is necessarily a derived class of `RevInstEntry`. No slicing occurs, because all data members are stored in `RevInstEntry`.

In the large text diffs, it is mostly the change of:
 `{ RevInstEntryBuilder<RevInstDefaults>() ... .InstEntry }`
to:
`{ RevInstDefaults() ... }`.

All CI tests pass with this change.


leekillough:
The tables have been minimized in size and made easier to read, by:
- Setting the most common settings in a `RevInstEntry` derived class' constructor, so that those common settings for a class of instructions do not need to be repeated for each instruction
- Removing all `.Set()` calls which do not change a default value, so that lines are much smaller
- Changing the order of the `.Set()` calls so that as many rows of instructions as possible have columns which line up. A canonical ordering of { Mnemonic, Funct3, Funct2or7, ImplFunc, ... [regClass settings] ..., Format, Opcode, ... [other settings] } is used to make the columns line up as much as possible without a lot of extra spacing.

The new tables are more compact and highlight the crucial differences between instructions in a given table, moving the common settings out as much as possible.

jleidel:
Should this also require updates to `ExtensionDevelopment.md`?

leekillough:
I'll update the docs.

leekillough:
This keeps `RevRegClass::RegIMM` set exactly the same as in `devel` -- no changes there.

donofrio (approved):
Looks good - before we merge we should give a couple downstream devs a heads up this is coming in case it breaks custom instruction extensions they have created.

leekillough:
Cc: @Artlands

Artlands:
> Cc: @Artlands

Thanks for the heads-up! I will update the xBGAS instruction table accordingly.

leekillough:
> > Cc: @Artlands
> 
> Thanks for the heads-up! I will update the xBGAS instruction table accordingly.

If you don't mind, I will merge this now, so that you can see the new format and work with it locally. I don't expect too much disruption -- in fact, it makes things simpler.

PR #230, opened 2023-12-19
Fix incorrect register classes causing dependency errors; add new MemH test for dependency checks; remove obsolete MemH tests

leekillough:
This adds a new test `ex7` which detects https://github.com/tactcomplabs/rev/issues/228 on MemH but shows it passing on RevMem.

This also removes the obsolete `memh_1` and `memh_2` test directories, since they aren't used anymore.

Issue #228, opened 2023-12-19
A problem with FPU register dependency tracking


ukasz:
**Describe the bug**
We found an issue which is reproducible with config using memory hierarchy (tests/memh_2), but not on basic one (tests/ex2).
Failing example is shown below.
```
// clang acc.c -o acc -march=rv64imafd
#include <assert.h>
#define N 3
 
int main() {
        float a[2];
        float res = 0;
        for(int i = 0; i< 2; i++)
                a[i] = i;
        for(int i = 0 ;i < 2 ;i++)
                res+=a[i];
        assert(res != 3.0f); //just to keep optimizer away
        return 0;
}
```
Alternatively you may be able to reproduce it with C++ std::accumulate:
```
template <typename T> T test_accumulate() {
        T tab[N];
        for (int i = 0; i < N; i++)
                tab[i] = i;
        T sum = std::accumulate(tab, tab + N, 0);
        return sum;
}
template <typename T> T expected(T x) {
        return (x*(x-1))/2;
}

int main()
{
        float ret_float = test_accumulate<float>();
        assert(expected<float>(N) == ret_float);
}
```

What the program does is a sequence of stores followed by a sequence of loads and floating point addition.
On basic config calculated sum is equal to the expected value. On memh config it is not.
In dissassembly we have:
```
00000000000120ac <main>:
   120ac: 13 01 01 fd   addi    sp, sp, -48
   120b0: 23 34 11 02   sd      ra, 40(sp)
   120b4: 23 30 81 02   sd      s0, 32(sp)
   120b8: 13 04 01 03   addi    s0, sp, 48
   120bc: 13 05 00 00   li      a0, 0
   120c0: 23 26 a4 fe   sw      a0, -20(s0)
   120c4: 23 20 a4 fe   sw      a0, -32(s0)
   120c8: 23 2e a4 fc   sw      a0, -36(s0)
   120cc: 6f 00 40 00   j       0x120d0 <main+0x24>
   120d0: 83 25 c4 fd   lw      a1, -36(s0)
   120d4: 13 05 10 00   li      a0, 1
   120d8: 63 54 b5 00   bge     a0, a1, 0x120e0 <main+0x34>
   120dc: 6f 00 40 03   j       0x12110 <main+0x64>
   120e0: 6f 00 40 00   j       0x120e4 <main+0x38>
   120e4: 03 25 c4 fd   lw      a0, -36(s0)
[...]

PR #242, opened 2024-01-13
Cleanup dependency check to avoid using isFloat bool to describe entire instruction, and make code more readable

leekillough:
This ends the use of the `bool isFloat()` function and associated `bool isFloat` parameter, which was previously used to categorize RV instructions as either "totally GP-register", or "totally FP-register". This change instead uses `RevRegClass` instead of `isFloat` everywhere.

In instructions which read/modify both GP and FP registers, this was a problem, as illustrated by https://github.com/tactcomplabs/rev/issues/228 and https://github.com/tactcomplabs/rev/pull/230 , although that was caused by putting the incorrect `RevRegClass` fields in an instruction during table building, and not because of using `isFloat`. However, it looks like the table had been generated by hand by assuming that `isFloat` completely describes the instructions.

The `ExternalDep*` functions used the old interface with `isFloat`, and the old functions are kept around but marked deprecated. The new functions use a `RevRegClass` parameter, which is already defined in `common/include/RevCommon.h`.

The dependency checking functions have been modified to use `switch` instead of `if` on `RevRegClass`, so that how each `RevRegClass` is handled differently is more clear, and so that additional ones can be added easily.

Complex Boolean expressions are replaced by `switch` or `if`-`else` statements for readability. For example the compact but cryptic:
```c++
 return (reg != 0 || regClass != RevRegClass::RegGPR) &&
      regFile->GetLSQueue()->count(LSQHash(reg, regClass, HartID)) > 0;
```
is replaced with:
```c++
    if(reg == 0 && regClass == RevRegClass::RegGPR){
      return false;   // GPR x0 is not considered
    }else{
      return regFile->GetLSQueue()->count(LSQHash(reg, regClass, HartID)) > 0;
    }
```

The FP rounding mode (**rm**) used to be decoded iff `isFloat()` was true, but that is not necessary. Only the `R` format FP instructions with opcode `0b1010011` and the `R4` format FMA instructions, need to decode the FP rounding mode. No `I`, `S`, `U`, `B`, etc. formats encode a FP rounding mode, so that unnecessary code which called `isFloat()` has been removed from the decoder.

`MarkLoadComplete()` was simplified, by using C++17 syntax and by combining two error messages into one.

For example, the C++17 syntax for iterating through all elements in a `std::unordered_multimap` which equal a given key:
```c++
 for(auto [i, end] = LSQueue->equal_range(req.LSQHash()); i != end; ++i){
```


include/RevProc.h
```diff
@@ -730,45 +759,45 @@ class RevProc{
...
-    return RegFile->FP_Scoreboard.any() || RegFile->RV_Scoreboard.any();
+    switch(regClass){
+      case RevRegClass::RegGPR:
+        return regFile->RV_Scoreboard.any();
+      case RevRegClass::RegFLOAT:
+        return regFile->FP_Scoreboard.any();
+      case RevRegClass::RegUNKNOWN:
+        return regFile->RV_Scoreboard.any() || regFile->FP_Scoreboard.any();
+      default:
+        return false;
```
donofrio, line 777: This should be an error condition - an invalid RegClass was passed

leekillough: I thought that if an "invalid" or "unknown" `RevRegClass` was passed, the dependency checker should return `false` as the default. If a register class is unknown or invalid, it does not have any dependencies as far as this check goes.

There is a `RegIMM` class which I have thought about removing (replacing instances of it with `RegUNKNOWN`), but which I wanted to discuss with you first.

donofrio (approved):
Looks good - use of the RegClass in lieu of the isFloat is overall clearer. good to merge

PR #243, opened 2024-01-16
Removing superfluous  ReadRequest when hitting page boundary

donofrio:
When operating Rev with MemH enabled if we hit a page boundary (like we are operating with `RevMem`) we generate two ReadRequests. This will cause two read responses to the same address / register causing an exception to be thrown in the `MarkLoadComplete() `function upon return of the 2nd read response.  This change so regardless of where we are in terms pf page boundaries we generate a single read request. 

There is still some dead code in there... wanted to make it _very_ clear what changed.


leekillough (approved):
LGTM

jleidel (approved):
good to merge

PR #244, opened 2024-01-16
Overhaul ECALL / Syscall handling

### LINKED ISSUE #253: Out of Memory error when using rev_mmap (with the memH memory backend)
**Describe the bug**

Running the following program will give the out of memory error.

```
int main () {
    size_t total_size = 1024;
    void *ptr = (void *)rev_mmap(0, total_size,  PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
}
```

**To Reproduce**
- SST-Core Version (13.1.0), gcc-10.1.0
- The devel branch
- The last commit hash: 9b7ef56ebd52f8960b8aa2c6098ac6f7f1c5acc3
- The compiler utilized to build Rev: gcc-10.1.0
- The compiler utilized to build the test: riscv64-unknown-elf-gcc 12.2.0
- The Python simulation input:
```
import os
import sst

VERBOSE = 1
MEM_SIZE = 1024*1024*1024-1

comp_cpu = sst.Component("cpu", "revcpu.RevCPU")
comp_cpu.addParams({
        "verbose" : VERBOSE,                                # Verbosity
        "numCores" : 1,                               # Number of cores
        "clock" : "2.0GHz",                           # Clock
        "memSize" : MEM_SIZE,                         # Memory size in bytes
        "machine" : "[0:RV64GC]",                      # Core:Config; RV64G for core 0
        "startAddr" : "[0:0x00000000]",               # Starting address for core 0
        "memCost" : "[0:1:10]",                       # Memory loads required 1-10 cycles
        "program" : os.getenv("REV_EXE", "malloc.exe"),  # Target executable
        "enable_memH" : 1,                            # Enable memHierarchy support
        "splash" : 1                                  # Display the splash message
})

comp_lsq = comp_cpu.setSubComponent("memory", "revcpu.RevBasicMemCtrl")
comp_lsq.addParams({
      "clock"           : "2.0Ghz",
      "max_loads"       : 64,
      "max_stores"      : 64,
      "max_flush"       : 64,
      "max_llsc"        : 64,
      "max_readlock"    : 64,
      "max_writeunlock" : 64,
      "max_custom"      : 64,
      "ops_per_cycle"   : 64
})

iface = comp_lsq.setSubComponent("memIface", "memHierarchy.standardInterface")

l1cache = sst.Component("l1cache", "memHierarchy.Cache")
l1cache.addParams({
    "access_latency_cycles" : "4",
    "cache_frequency" : "2 Ghz",
    "replacement_policy" : "lru",
    "coherence_protocol" : "MESI",
    "associativity" : "4",
    "cache_line_size" : "64",
    "L1" : "1",
    "cache_size" : "16KiB"
})

memctrl = sst.Component("memory", "memHierarchy.MemController")
memctrl.addParams({
    "clock" : "2GHz",
    "addr_range_start" : 0,
    "addr_range_end" : MEM_SIZE,
    "backing" : "mmap",
})

memory = memctrl.setSubComponent("backend", "memHierarchy.simpleMem")
memory.addParams({
    "access_time" : "100ns",
    "mem_size" : "8GB"
})

link1 = sst.Link("link1")
link1.connect( (iface, "port", "1ns"), (l1cache, "high_network_0", "1ns") )
link2 = sst.Link("link2")
link2.connect( (l1cache, "low_network_0", "1ns"), (memctrl, "direct_link", "1ns") )

```
- The test source is similar to `/test/syscalls/munmap` but using the aforementioned code.

**Expected behavior**
The simulation should complete without any issues.

**Trace**
```
RevCPU[cpu:RevCPU:0]: Initialization of RevCPUs complete.
Expanding heap by 1073740635 Bytes
Expanding heap from 0x159b1 to 0x4001550c
MaxHeapSize = 0xfffffff
FATAL: RevCPU[cpu:ExpandHeap:648500]: Out Of Memory --- Attempted to expand heap to 0x4001550c which goes beyond the maxHeapSize = 0xfffffff set in the python configuration. If unset, this value will be equal to 1/4 of memSize.
```

From the trace, we can see that the `total_size` is not passed to rev_mmap correctly. 

**Additional context**
The value of `total_size` is not loaded to `a1` before executing rev_mmap. This is because RevProc::ECALL_mmap does not check the scoreboard. The scoreboard should be checked for all registers used in the system calls to avoid this issue. I add the following code before reading the registers in RevProc::ECALL_mmap to fix the bug:

```
if (ScoreboardCheck(RegFile, static_cast<uint16_t>(RevReg::a0), RevRegClass::RegGPR) || 
      ScoreboardCheck(RegFile, static_cast<uint16_t>(RevReg::a1), RevRegClass::RegGPR)) {
    return EcallStatus::CONTINUE;
  }
```

leekillough:
1. It uses `__attribute__((naked))` to prevent the compiler from generating any prologue/epilogue code in the Syscalls wrappers, or from making any assumptions about the uses of the Syscall wrapper's arguments, or of how the return value register is set. This allows raw inline assembly, and we set `a7` and then do the `ECALL` directly, and then `RET`, since the `a0` value returned from `ECALL` is the return register for the wrapper. <br> <br>**Previously, at `-O3` the entire wrapper call could be eliminated (not just inlined) by the compiler, because there was no indication that it used its arguments, or that it generated its return value from inline assembly.**<br><br>It would be very messy to do this the conventional way for each Syscall, telling the compiler how its function parameters are passed into the inline assembly as inputs, making sure they are placed in the correct registers for the `ECALL`, and telling the compiler how the `ECALL` return register of `a0` needs to become the wrapper function's return register `a0` so that it doesn't think it can just throw away a call to the wrapper. With `__attribute__((naked))`, this complexity is avoided, and the wrappers can become one-liners.
2. It uses `static` for the wrapper functions, since this is in a header file and may be `#include`d more than once. `inline` is incompatible with `__attribute__((naked))`, so `inline` was left out. <br><br>**Previously, multiply defined symbols could occur if the same Syscall wrapper function was called from two different `.c` files.**
3. It uses a macro to simplify the generation of the wrappers, and gets rid of a redundant "table of comments". The macro takes the Syscall number and the prototype of the wrapper as its two arguments.
4. It fixes the incorrect names of some Syscalls in `include/RevProc.h`'s comments.<br>
5. It adds parameter names to C function _definitions_, because **parameter names were not optional in function _definitions_ in C until C20, and they produce a lot of warnings.** In particular, the `socket` Syscalls did not have parameter names.<br>
6. It fixes the return types of many of the Syscalls, since simply using `int` will cause C/C++ code to expect a 32-bit result, and it will be sign-extended on RV64. `long`, `size_t`, `ssize_t`, `void*`, and other types should be used instead of always using `int`. **An automated program going through Syscall prototypes in headers/manpages should be eventually used to determine the correct return types.**
7. It waits for `a0`-`a7` dependencies to be satisfied before executing `ECALL`. Although different Syscalls take different arguments, waiting for `a0`-`a7` to be ready is general enough for any Syscall. Fixes #253.
8. ECALLs are implemented by setting SCAUSE during the `ECALL` instruction, and then at each `ClockTick`, checking whether SCAUSE is set, and if so, executing the Ecall handler and stalling the current instruction. SCAUSE is cleared when the Ecall handler indicates that it has completed. There is no backtracking of `PC` anymore, and the `ECALL` instruction only gets executed once.
9. `SCAUSE` is implemented as a simple `enum class` member of `RevRegFile`. There are no longer two versions of it, one for RV32 and one for RV64.
10. The `ECALL_*` Syscall handler functions no longer take a `RevInst` parameter, since it was never used.
11. The `Ecalls` table mapping Syscall numbers to handler functions, maps them to `RevProc` member function pointers, rather than the much more expensive `std::function`.
12. The `Ecalls` table is `static` now, and is initialized in `src/RevSysCalls.cc`.
13. `EXCEPTION_CAUSE` in `RevInstTable.h` is renamed to `RevExceptionCause` in `RevRegFile.h` and is made an `enum class`. This makes it more type-safe and abides by the CamelCase naming convention.
14. `EcallStatus`'s values `SUCCESS`, `CONTINUE`, `ERROR` are no longer defined in terms of `EXCEPTION_CAUSE`, because the exception cause is unrelated to the status of an `ECALL` handler function.


rkabrick (approved):
Looks much better than it used to! Nice work.

jleidel (approved):
good to merge

donofrio (approved):
Looks good - this should address a bunch of bugs

PR #245, opened 2024-01-16
Fix syscall wrappers

leekillough:
This in an emergency, abbreviated version of https://github.com/tactcomplabs/rev/pull/244 .
1. Use `__attribute__((naked))` to avoid bad optimization of syscall wrappers such as `rev_mmap()`.
2. Use `static` to avoid multiple definitions of syscall wrappers in header files.

Syscalls to `rev_mmap()` were optimized out of one of the partner's tests, causing weird memory allocation failures, as explained in https://github.com/tactcomplabs/rev/pull/244 .

There's also a good chance that multiply defined symbols will occur.

This is a stopgap measure. The more thorough fix is in https://github.com/tactcomplabs/rev/pull/244 .


jleidel (approved):
good to merge

PR #247, opened 2024-01-29
Clear Ecall state every time we successfully execute an ecall

rkabrick:
This clears the ecall state object every time we successfully return from an Ecall instruction. We weren't doing this before and it would lead to some odd behavior for system calls that relied on information contained in this object (ie. `if(EcallState.bytesRead .... `

### DIFF
--- a/include/RevSysCalls.h
+++ b/include/RevSysCalls.h
@@ -41,7 +41,7 @@ struct EcallState {
     string.clear();
     path_string.clear();
     bytesRead = 0;
-    buf[0] = '\0';
+    buf.fill('\0');
   }
   EcallState() {
     buf[0] = '\0';
--- a/src/RevProc.cc
+++ b/src/RevProc.cc
@@ -2372,8 +2372,10 @@ void RevProc::ExecEcall(RevInst& inst){
 
     // For now, rewind the PC and keep executing the ECALL until we
     // have completed
-    if(EcallStatus::SUCCESS != status){
+    if( status != EcallStatus::SUCCESS ){
       RegFile->SetPC( RegFile->GetPC() - inst.instSize );
+    } else {
+      Harts[HartToDecodeID]->GetEcallState().clear();
     }
   } else {
     output->fatal(CALL_INFO, -1, "Ecall Code = %" PRIu64 " not found", EcallCode);


jleidel (approved):
good to merge

include/RevSysCalls.h
```diff
@@ -41,7 +41,7 @@ struct EcallState {
     string.clear();
     path_string.clear();
     bytesRead = 0;
-    buf[0] = '\0';
+    buf.fill('\0');
```
leekillough, line 44: I don't think we have to `fill` the entire `buf` -- or even the first byte of it, with `\0`. In `rev_write` we use:
```c++
    EcallState.string += std::string_view(EcallState.buf.data(), EcallState.bytesRead);
```
The `buf.data()` returns a pointer to the underlying array data, and `bytesRead` tells how many bytes we've read so far, so clearing `bytesRead` should be sufficient.

The `string_view` creates a view of a not-necessarily-null-terminated string which is a raw number of bytes. That way we can read data containing `\0` and not stop on `\0`. Although `std::string` is guaranteed to be null-terminated, it may contain `\0` bytes embedded within its first `size()` bytes.

src/RevProc.cc
```diff
@@ -2372,8 +2372,10 @@ void RevProc::ExecEcall(RevInst& inst){
 
     // For now, rewind the PC and keep executing the ECALL until we
     // have completed
-    if(EcallStatus::SUCCESS != status){
+    if( status != EcallStatus::SUCCESS ){
       RegFile->SetPC( RegFile->GetPC() - inst.instSize );
+    } else {
+      Harts[HartToDecodeID]->GetEcallState().clear();
```
leekillough, line 2378: This is probably a good general solution for all `ECALL`s, but we already call `EcallState.clear()` on success in `rev_write`:

```c++
  if(nleft == 0 && LSQueue->count(lsq_hash) == 0){
    int rc = write(fd, EcallState.string.data(), EcallState.string.size());
    RegFile->SetX(RevReg::a0, rc);
    EcallState.clear();
    DependencyClear(HartToExecID, RevReg::a0, false);
    return EcallStatus::SUCCESS;
}
```

leekillough:
Closing because of https://github.com/tactcomplabs/rev/pull/249

PR #248, opened 2024-01-29
Lee write fix

leekillough:
My WIP fix for the write issue. Need to compare.


rkabrick (commented):
Not sure about the slurm changes (idk anything about what is right there) but for the rest of the changes it looks good. We talked about how some of the changes contained are moot now that we have the dependency tracking for ecalls taken care of. Maybe rename the branch before doing a PR so it doesn't look like these changes are what fixed the write issue? Not necessary but may be an idea

jleidel:
> Not sure about the slurm changes (idk anything about what is right there) but for the rest of the changes it looks good. We talked about how some of the changes contained are moot now that we have the dependency tracking for ecalls taken care of. Maybe rename the branch before doing a PR so it doesn't look like these changes are what fixed the write issue? Not necessary but may be an idea

I think those are fine.  These are simple BASH changes.

leekillough:
The Bash script changes were made in my working directory, and are also in #244

PR #249, opened 2024-02-01
Refine how EcallState is cleared

leekillough:
This is a refinement of https://github.com/tactcomplabs/rev/pull/247

- Make Ecalls clear `EcallState` when `EcallStatus::SUCCESS` is returned
- Remove redundant clears of `EcallState` when `EcallStatus::SUCCESS` is returned
- Fix potential bug when `EcallState.string` is moved to `EcallState.path_string` (guarantee `string` is empty after it is moved from, which `std::move` does not guarantee)
- Do not zero out `EcallState.buf` because `EcallState.bytesRead` indicates the number of bytes in `buf` which are valid, and `buf` does not need to be zeroed, just `bytesRead`

### DIFF
--- a/include/RevSysCalls.h
+++ b/include/RevSysCalls.h
@@ -41,11 +41,9 @@ struct EcallState {
     string.clear();
     path_string.clear();
     bytesRead = 0;
-    buf[0] = '\0';
   }
-  EcallState() {
-    buf[0] = '\0';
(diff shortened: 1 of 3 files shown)


rkabrick:
LGTM... If @jleidel and @donofrio are cool with it merge

PR #251, opened 2024-02-19
Added the same environment check for RVOBJDUMP that we do for RVCXX & RVCC

rkabrick:
We don't check to see if `RVOBJDUMP` is set in the environment first but we do for RVCC and RVCXX... The logic to assign it automatically based on `RISCV` makes some assumptions that won't hold up for all environments. 

This just adds a check to see if there is already a value set for RVOBJDUMP set as a environment variable.


leekillough (approved):

jleidel (approved):
approve to merge

PR #252, opened 2024-02-22
Simplify RevMem atomic operations

leekillough:
When arbitrarily reviewing the atomic memory operation code on RevMem, it was confusing. This change cleans it up.
- Move variables unused in RevMem to the MemH branch of the `if`-statement
- Use local variables instead of allocating on heap
- Use anonymous unions to avoid casts
- Avoid unnecessary `memset()` and `memcpy()` calls
- Add explanatory comments, since the original steps were unclear


jleidel (approved):
Good to merge.  This may need special care when merging to downstream forks, but this looks good.

PR #254, opened 2024-03-11
Add clang-format support

leekillough:
This adds a `clang-format` `pre-commit` hook to reformat files before commit.

Lines are kept less than 80 characters long.

Spaces are inside of parentheses (after `(` and before `)`), but not outside of them. `clang-format`  version 15 options were carefully tweaked to keep the formatting as close to the existing code as possible. I don't know which version of `clang-format` others use, so it needs to be tested on a variety of platforms. If `clang-format` cannot be found, it won't be run.

`// clang-format off` and `// clang-format on` directives were inserted in a few files to prevent unwanted reformatting (mainly in large tables).

The `pre-commit` script automatically changes the copyright date at the top of the file. It is not necessary to run a script every new year on all files. Usually copyrights are dated based on the last modification date of a file, and this script labels copyrights using that convention.

This PR only adds the `clang-format` and `pre-commit` hooks -- it does not reformat the entire code base. The code base will be incrementally updated, and `clang-format` will be run on each file as it is committed in future commits. This reduces the chance of conflicts with other changes.

The entire code base can be reformatted by running `pre-commit --reformat`, but that will produce a huge commit of mostly-whitespace changes as lines are wrapped to be less than 80 characters. Doing it incrementally at each file's commit will reduce the disruption.

Non-ASCII characters in source files are converted to ASCII equivalents.

Right now only `.h` and `.cc` files are passed to `clang-format` -- other file extensions can be added later if necessary.


kpgriesser (approved):
no issues

jleidel (approved):
good to merge

leekillough:
I'll wait for @donofrio and @rkabrick's review first -- I found many `clang-format` options were dependent on `clang-format` version, so on MacOS it might be harder to get this version, or there may be other differences such as Linux tools not as easy to get on MacOS. I can walk through an example of using it on a test branch.

leekillough:
To test the `pre-commit` and `clang-format`:
```
git checkout devel
git checkout -b test
# make a change to a large file like RevProc.cc
git add <files>
git commit -m "test clang-format"
git status
```
You might see errors during the `git commit`. If not, then `git status` should be clean. I may need to double-check the `pre-commit` script to make sure every error case is handled properly, and/or to add tests to make sure that the current version of `clang-format` supports the `.clang-format` file's options.

jleidel:
should we also add this to our standard Jenkins harness?  
Which version of Clang were you using to test?

leekillough:
As for Jenkins:
1. The existing CI tests should be sufficient to catch any errors introduced by `clang-format` since they are performed `pre-commit`.
2. Extra tests can be added to enforce `clang-format` does not change source code, producing failures if the CI tests are run on code which needs modification by `clang-format`, but this might make failures harder to understand, especially for outside partners, and it might slow down progress of merging changes which otherwise pass all other CI tests.

I am using `clang-format` 15 on Ubuntu.

leekillough:
In other words, `clang-format` is run locally before a commit, and is not run during builds or tests. It simply ensures that the _committed_ code abides by the style guidelines such as very few lines > 80 characters, brace style, etc.

After it is committed, the CI build and testing, as well as any manual builds or scripts, work exactly the same as before this PR. The only change in workflow is a `pre-commit` script depending on `clang-format` is run on each file before it's committed.

The only dependency is that the user has a locally executable version of `clang-format` which is a recent enough version that it is able to handle the `.clang-format` configuration file's directives. Older versions of `clang-format` may report unrecognized options and abort with errors.

The `clang-format` run is not architecture-specific, and is not dependent on a RISC-V LLVM toolchain. A local desktop, laptop or workstation `clang-format` executable is sufficient if it is a recent enough version.

leekillough:
I want to get an idea of probable cases where the `pre-commit` script can fail, whether due to `clang-format` or something else, and work around them, or add better error-handling, with diagnostic messages telling the user what they need to do to fix the problem.

leekillough:
This would depend on the OS/desktop/build environments of partners, how easily available `clang-format` is on those platforms, the `pre-commit` modes of failure and how to handle them, etc.

This PR introduces a feature which is externally visible to all contributors, and which cannot simply be tucked under the CI.

(We could do it ourselves, but then every external PR has to be re-added and re-committed by ourselves on a local machine; I don't any have experience running `pre-commit` on the GitHub server side, although I suppose it's possible with Actions.)

leekillough:
A Git hook must be installed by the user, either with `git config core.hooksPath` or copying/symlinking hooks into `.git/hooks`.

This PR enforces the `clang-format` changes by running `scripts/test_git_hooks.sh` at the beginning of every build. If `core.hooksPath` is not set to `.githooks`, the build aborts with instructions on how to configure Git to use `.githooks`. Unfortunately there's not a way for Git hooks to be installed automatically as a part of tracked (committed) files, unless we nefariously modify the user's Git configuration in a build script.

The CI tests are not negatively affected by this change, so they either already use `.githooks`, or they call `cmake` differently which avoids running `scripts/test_git_hooks.sh`.

The `pre-commit` hook edits 4 properties of text files:
1. Copyright year (or range) at the top of the file, to match the current year of the commit
   * Bye bye annual New Year's mass copyright date changes
2. No trailing whitespace at the ends of lines, and always LF at end of file
3. Non-ASCII characters (e.g. "smart quotes") replaced with ASCII equivalents
4. `clang-format` formatting according to `.clang-format`
   - Few lines > 80 characters (rare exceptions, like large formatted tables between `// clang-format off` and `// clang-format on` or trailing Doxygen comments)
   - Brace, parenthesis, declaration, namespace, function, macro formatting similar to existing code
   - `#include`s sorted alphabetically (makes duplicate `#include`s easier to spot, and a canonical ordering makes insertion points uniform across files)
   - If a `clang-format` command cannot be found in `PATH`, platform-specific (MacOS, Debian/Ubuntu, RedHat/Fedora/CentOS, etc.) instructions for installing it are printed, and the commit is aborted

If any of these 4 properties modifies a file, a yellow warning message describing the change and affected file is printed during `git commit`. The modifications are only performed on files staged for commit unless the `--reformat` option is passed to `pre-commit` to reformat the entire tree, which will likely cause disruption.

This has been tested on Linux and MacOS interactively, and scripts modified to work on both. Others' feedback is desired, by running this in your environment (also try builds, follow error messages, etc.):
```
git clone -b clang-format git@github.com:tactcomplabs/rev.git
cd rev
git config core.hooksPath .githooks
git checkout -b test
# make a change to a large file like RevProc.cc
git add <files>
git commit -m "test clang-format"
git status
```

rkabrick:
Things seem to be working well. I've attached what happened after following your instructions. 

I didn't have `clang-format` installed so I was prompted to install it via `brew install clang-format` after trying to do a `git commit -m "test"`. So I did `brew install clang-format` then repeated my git commit command and saw the following. 

This is on an M1 Mac

image

leekillough:
Thanks; that's expected behavior, although the wording of the yellow clang-format change warning could be improved. I've been scratching my head on how best to phrase it.

kpgriesser:
confirmed line reformatting works on Ubuntu 22.04 and macOS Sonoma 14.2.1

leekillough:
I have changed the wording to be like:
```
src/RevCPU.cc: Reformatting with clang-format according to coding guidelines
```

donofrio (approved):

PR #255, opened 2024-03-11
Use do{ }while(0) idiom for function-style macros

leekillough:
When macros can be called like functions, use `do{ }while(0)` [around the body of complex macros](https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for/257425#257425) (more than single expressions) so that the macro can be called like a function and used anywhere inside an `if`-statement or elsewhere with normal syntax (ending statements with semicolons, etc.).

### DIFF
--- a/include/RevTracer.h
+++ b/include/RevTracer.h
@@ -40,13 +40,13 @@
 
 // Tracing macros
 #ifndef NO_REV_TRACER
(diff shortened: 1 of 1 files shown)


kpgriesser (approved):
Interesting, I hadn't used this construct for macros before. No issues with the PR.

Issue #260, opened 2024-03-19
c.addi16sp instruction is incorrectly printed as c.lui by Rev Tracer


leekillough:
**Describe the bug**

When `rd == x2` (stack pointer), the `c.lui` instruction should be printed as a `c.addi16sp` instruction. When `rd == 0`, it is a reserved hint instruction.

**To Reproduce**

Latest `devel`

**Expected behavior**

Rev Tracer should print `c.addi16sp` when `rd == x2`.

**Trace**

```
RevCPU[cpu:GetNextHartToDecodeID:41463000]: Core 0; Hart switch from 0 to 0
RevCPU[cpu:FetchAndDecodeInst:41463000]: Core 0; Hart 0; Thread 1; PC:InstPayload = 0x1e130:0x80826125
RevCPU[cpu:Render:41463000]: Core 0; Hart 0; Thread 1; *I 0x1e130:80826125    c.lui %rd, $imm    0x3fffea7c<-x2 x2<-0x3fffeadc
```
**Additional context**

Since loading the upper part of the stack pointer with a compressed instruction is uncommon, `c.lui` with `rd == x2` was reserved as an instruction to add to `x2` instead of loading an upper immediate.

kpgriesser:
see comments in #261. We should agree on how to handle the default internal mnemonic.   Building with spike give the expected mnemonic.

// REV_USE_SPIKE==OFF
RevCPU[cpu0:Render:74729000]: Core 0; Hart 0; Thread 1; *I 0x1014a:80826105    c.lui %rd, $imm	 0x3fffffffbdc<-x2 x2<-0x3fffffffbfc 
// REV_USE_SPIKE==ON
RevCPU[cpu0:Render:74729000]: Core 0; Hart 0; Thread 1; *I 0x1014a:80826105    c.addi16sp sp, 32	 0x3fffffffbdc<-sp sp<-0x3fffffffbfc

PR #262, opened 2024-03-22
Add instruction table predicate option, to disambiguate compressed instructions

### LINKED ISSUE #261: Tracer is printing compressed register move instructions as c.jr instructions
**Describe the bug**

When a compressed register <- register copy ( `c.mv` ) is encountered, Rev Tracer prints it as a `c.jr` instruction.

**To Reproduce**

Latest `devel` 

**Expected behavior**

It should print `c.mv` when `rs2 != 0`

**Trace**

```
RevCPU[cpu:GetNextHartToDecodeID:41465000]: Core 0; Hart switch from 0 to 0
RevCPU[cpu:FetchAndDecodeInst:41465000]: Core 0; Hart 0; Thread 1; PC:InstPayload = 0x19fce:0xe90187aa
RevCPU[cpu:Render:41465000]: Core 0; Hart 0; Thread 1; *I 0x19fce:e90187aa    c.jr %rs1  0x0<-x0 0x0<-x10 x15<-0x0
```
```
RevCPU[cpu:GetNextHartToDecodeID:41467000]: Core 0; Hart switch from 0 to 0
RevCPU[cpu:FetchAndDecodeInst:41467000]: Core 0; Hart 0; Thread 1; PC:InstPayload = 0x19fd2:0xf0ef8522
RevCPU[cpu:Render:41467000]: Core 0; Hart 0; Thread 1; *I 0x19fd2:f0ef8522    c.jr %rs1  0x0<-x0 0x90<-x8 x10<-0x90
```

**Additional context**

`C.MV` copies the value in register `rs2` into register `rd`. `C.MV` expands into `add rd, x0, rs2`. `C.MV`
is only valid when `rs2 != x0`; the code points with `rs2 = x0` correspond to the `C.JR` instruction. The
code points with `rs2 != x0` and `rd = x0` are HINTs.

`C.MV` expands to a different instruction than the canonical `MV` pseudoinstruction, which instead
uses `ADDI`. Implementations that handle `MV` specially, e.g. using register-renaming hardware,
may find it more convenient to expand `C.MV` to `MV` instead of `ADD`, at slight additional hardware
cost.

leekillough:
This adds an optional predicate to the instruction table entries, which can be set by calling `.SetPredicate()`.

This fixes #261 and #260 by having separate table entries for the different mnemonics, and using the `rd/rs1` or `rs2` register fields to disambiguate them.

A predicate can be added to an instruction table entry as simply as `.SetPredicate( []( uint32_t Inst ){ return DECODE_LOWER_CRS2(Inst) != 0; } )`, which in this example makes sure that the compressed-encoded `rs2` register is not `x0` for the `c.mv %rd, %rs2` mnemonic to apply. If `rs2 == x0`, then the `c.jr %rs1` mnemonic is used instead, based on another instruction table entry with the opposite predicate condition.

Previously, after our overhaul of the compressed instruction decoding, there were still 4 compressed instruction `Execute` functions which had to disambiguate the instructions at execution time and dispatch them to the correct uncompressed instruction `Execute` function. Now that we can fully disambiguate them in the instruction tables by using predicates, the execute functions can be totally separated and we don't need to disambiguate them at `Execute` time.

For the mappings from uncompressed and compressed encodings to Instruction Table Entries, the `std::map` has been changed to a `std::unordered_multimap` to allow multiple entries with the same encoding key, and a new function `matchInst()` has been added to replace the previous calls to `EncToEntry.find(Enc)` and `CEncToEntry.find(Enc)`. `matchInst()` returns the first table entry returned by `std::unordered_multimap::equal_range(Enc)` whose predicate is satisfied. This way even if multiple table entries match the same encoding, only the ones satisfying their predicate will be found during lookup. The predicate function, which defaults to always returning `true`, is passed the 16/32-bit raw instruction code, and almost all of the predicates use the already-existing `DECODE_RD()` and `DECODE_LOWER_CRS2()` macros to decide if an entry matches an instruction mnemonic.

```c++
// Find the first matching encoding which satisfies a predicate, if any
auto RevCore::matchInst(
  const std::unordered_multimap< uint32_t, unsigned >& map,
  uint32_t                                             encoding,
  const std::vector< RevInstEntry >&                   InstTable,
  uint32_t                                             Inst ) const {
  // Iterate through all entries which match the encoding
  for( auto [it, end] = map.equal_range( encoding ); it != end; ++it ) {
    unsigned Entry = it->second;
    // If an entry is valid and has a satisfied predicate, return it
    if( Entry < InstTable.size() && InstTable[Entry].predicate( Inst ) )
      return it;
  }

  // No match
  return map.end();
}
```
I have exhaustively gone through all of the compressed instructions, and have added predicates to handle special cases like `x0` or `x2` being registers, or an immediate value being `0`. If the RISC-V spec says that an instruction is only valid if a condition is true, and is otherwise "reserved", I add that condition to the predicate. If the RISC-V spec says that an instruction is a hint if a condition is true (such as `rd == x0`), and if the default instruction behavior now acts as a `nop` (such as by adding `0`), I don't add that condition to the predicate. Later when hints are defined by us or the RISC-V spec, we can add predicates to handle the hint and non-hint cases separately.

~~I have not exhaustively gone through all of the uncompressed instructions, some of which might be marked as "reserved" but not a "hint" by the spec if certain conditions are met, like `x0` or an immediate value of `0`. To be more true to the spec, we can add those predicates to uncompressed instructions later.~~

**Edit:** 
- I went through all of the uncompressed instructions, looking for constraints on their parameters in the spec, and adding any as predicates
- I fixed the `addiw` predicate
- I added predicates for  `slliw`, `srliw` and `sraiw` that `imm[5] == 0`
- I added pseudoinstructions, including `nop`
- I removed unsupported extension names which have gone away or are not close to ratification
- I'm investigating the timeout error in the `memset_2` MemH test, which only happens with GCC and SST 13.1.0; none of my changes should have slowed it down so as to cause a timeout. (**Edit:** The timeout has been increased from 60 to 75 seconds, because SST 13.1.0 is slower and it sometimes went over 60 seconds.)


rkabrick (approved):
I don't see anything wrong with this... Strange that 13.1.0 runs slower than 12 but alas... out of our control 

LGTM

jleidel:
There were "fixes" put in for 13.1.0 for several event handling inconsistencies.  Ultimately, this likely results in a slowdown of the SST core when propagating large numbers of messages.  We're hoping to see speedups in SST 14.0.0

include/insns/RV32I.h
```diff
@@ -142,67 +142,52 @@ class RV32I : public RevExt {
...
   // Compressed instructions
 
-  // c.addi4spn %rd, $imm == addi %rd, x2, $imm
-  // if Inst.imm == 0; this is a HINT instruction and is effectively a NOP
-  static bool
-    caddi4spn( RevFeature* F, RevRegFile* R, RevMem* M, const RevInst& Inst ) {
-    return ( Inst.imm == 0 ? nop : addi )( F, R, M, Inst );
-  }
+  // c.addi4spn %rd, $imm == addi %rd, x2, $imm, $imm != 0
+  static constexpr auto& caddi4spn = addi;
```
donofrio, line 158: adding 0 and 0 and storing to x0 (which is always zero) is, of course, a harmless operation, however, are we losing the representation of `nop` here (and a few other places)?

include/insns/RV32I.h
```diff
@@ -249,7 +239,8 @@ class RV32I : public RevExt {
...
     { RevInstDefaults().SetMnemonic("slt %rd, %rs1, %rs2"  ).SetFunct3(   0b010).SetImplFunc(slt   ).SetFunct2or7(0b0000000)                                                  .SetFormat(RVTypeR).SetOpcode(0b0110011) },
-    { RevInstDefaults().SetMnemonic("sltu %rd, %rs1, %rs2" ).SetFunct3(   0b011).SetImplFunc(sltu  ).SetFunct2or7(0b0000000)                                                  .SetFormat(RVTypeR).SetOpcode(0b0110011) },
+    { RevInstDefaults().SetMnemonic("sltu %rd, %rs1, %rs2" ).SetFunct3(   0b011).SetImplFunc(sltu  ).SetFunct2or7(0b0000000)                                                  .SetFormat(RVTypeR).SetOpcode(0b0110011).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) != 0; } ) },
+    { RevInstDefaults().SetMnemonic("snez %rd, %rs"        ).SetFunct3(   0b011).SetImplFunc(sltu  ).SetFunct2or7(0b0000000)                                                  .SetFormat(RVTypeR).SetOpcode(0b0110011).SetPredicate( []( uint32_t Inst ){ return DECODE_RS1( Inst ) == 0; } ) },
```
donofrio, line 243: Do we need an implementation of snez or does it fall out of the sltu implementation?

kpgriesser:
Going through the code my main comment would be how to audit and/or test that the decodings all match the spec assuming everything is hardcoded. Towards that end I thought a nice cross check would be to run tests with and without spike and compare the disassembly. Food for thought, here is the result of an initial experiment that shows some subtle but IMHO benign differences.  Overall, I see no reason not to merge this PR.

$ cd test/tracer
$ make -j -s
$ awk -F'*' '{print $2}' tracer.mem.rv64g.log | awk '{printf("%s %s\n",$2,$3);}' > insns

# no spike vs using spike on the disassembly branch

Here are represented differences with the instruction mnemonic side by side. The left side is using the internal disassembler and the right side is spike.
0x11d00:00300e13 addi   |       0x11d00:00300e13 li
0x11d14:fe0e18e3 bne    |       0x11d14:fe0e18e3 bnez
0x11d14:fe0e18e3 bne    |       0x11d14:fe0e18e3 bnez
0x11d14:fe0e18e3 bne    |       0x11d14:fe0e18e3 bnez
0x11d18:00000793 mv     |       0x11d18:00000793 li
0x11d2c:00008067 jalr   |       0x11d2c:00008067 ret

As you mentioned, the registers are not printed internally which would also be nice to add.  One major improvement over spike would also be using the program counter and relative offset in branches to determine the actual branch target and provide the actual symbol. With those changes we could avoid spike altogether but that is another PR.

kpgriesser (approved):

include/insns/RV32I.h
leekillough: That particular diff is only a textual difference in the source code (not a semantic one), because in the new code, we do not need to disambiguate functionality at execution time -- the execute function can be fixed in the table entry.

`nop` is separated out as a distinct mnemonic here:
```c++
    { RevInstDefaults().SetMnemonic("addi %rd, %rs1, $imm" ).SetFunct3(   0b000).SetImplFunc(addi  ).Setrs2Class(RevRegClass::RegUNKNOWN).Setimm(FImm)                        .SetFormat(RVTypeI).SetOpcode(0b0010011).SetPredicate( []( uint32_t Inst ){ return DECODE_IMM12( Inst ) != 0; } ) },
    { RevInstDefaults().SetMnemonic("mv %rd, %rs1"         ).SetFunct3(   0b000).SetImplFunc(addi  ).Setrs2Class(RevRegClass::RegUNKNOWN).Setimm(FImm)                        .SetFormat(RVTypeI).SetOpcode(0b0010011).SetPredicate( []( uint32_t Inst ){ return DECODE_IMM12( Inst ) == 0 && ( DECODE_RS1( Inst ) != 0 || DECODE_RD( Inst ) != 0 ); } ) },
    { RevInstDefaults().SetMnemonic("nop"                  ).SetFunct3(   0b000).SetImplFunc(addi  ).Setrs2Class(RevRegClass::RegUNKNOWN).Setimm(FImm)                        .SetFormat(RVTypeI).SetOpcode(0b0010011).SetPredicate( []( uint32_t Inst ){ return DECODE_IMM12( Inst ) == 0 && ( DECODE_RS1( Inst ) == 0 && DECODE_RD( Inst ) == 0 ); } ) },

```
There are three separate cases handled here with predicates:
1. `addi` where the immediate != 0
2. `addi` where the immediate == 0 and one of the two registers is not `x0` -- this corresponds to a `mv` pseudoinstruction
3. `addi` where the immediate == 0 and both registers are `x0` -- this corresponds to the `nop` pseudoinstruction

All 3 cases are handled here, and the correct instruction is displayed.

include/insns/RV32I.h
leekillough: > Do we need an implementation of snez or does it fall out of the sltu implementation?

`snez` is a pseudoinstruction which is a `sltu` instruction.

All pseudoinstructions listed in the RISC-V spec were implemented by me, by using a shorter name for the same instruction where applicable.

We do not have "an implementation of snez". We have a textual disassembly of `sltu` as `snez` when it applies as the pseudoinstruction mnemonic. The functionality is the same, provided by the `sltu` execute function.

leekillough: > We do not have "an implementation of snez". We have a textual disassembly of `sltu` as `snez` when it applies as the pseudoinstruction mnemonic. The functionality is the same, provided by the `sltu` execute function.

I think the source of confusion may be that in the past, there was one _and only one_ instruction table entry for each instruction, and that instruction necessarily had its own, and only, implementation.

The new table is not a one-to-one mapping anymore, hence the use of `std::unordered_multimap` instead of `std::map`. Instead, multiple table entries can exist for the same encoding, and only the one whose predicate is true is chosen, such as the pseudoinstruction mnemonic for an otherwise general instruction.

`addi` has 3 table entries, one for the general `addi` instruction, one for the `mv` class of pseudoinstructions, and one for the very specific pseudoinstruction `nop`. They all share the same `addi` execute function -- we do not need to implement `mv` or `nop` as separate execute functions, since they are all part of the same `addi` instruction. But we can distinguish them by their displayed mnemonic when we disassemble/trace them.

leekillough:
@kpgriesser : I'm trying to understand your `diff`. It might be easier to understand if you did a side-by-side `diff -y`.

One thing different from Spike is that we still use symbolic operand names like `%rd` and `%rs1` instead of the actual operands.

include/insns/RV32I.h
donofrio: makes sense - thanks Lee

donofrio (approved):
Looks good to me

PR #263, opened 2024-04-03
Allow files to reformat to be specified on the command line

leekillough:
This allows specifying filenames on the `pre-commit` script command-line, to reformat files which might exist outside of a Git repository.

Note that the `pre-commit` script should still exist in a Rev Git repository clone, because the script uses its own path to set the directory where the `.clang-format` file should be found, which is at the top of the Rev source tree.

`clang-format` [does not have a very good way](https://bugs.llvm.org/show_bug.cgi?id=20753) of explicitly specifying a `.clang-format` file. You can only specify it indirectly by using `--assume-filename`, where that refers to some file in a source tree where `.clang-format` exists up the directory tree. In this case, `pre-commit` uses `--assume-filename="$0"` (specifying itself) to point to somewhere in the tree where `.clang-format` can be found up the directory hierarchy.

So for example,
```
cd ~/mydir
~/rev/.githooks/pre-commit mysourcefile1.cpp mysourcefile2.cpp
```
Right now it does not support  recursively reformatting directories, unless it's called from inside a Git repo without explicit filenames (such as using the `--reformat` option to reformat an entire repo), because that would be too dangerous.

I have tested this on Linux but not on MacOS.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -31,28 +31,43 @@ set_colors() {
 
 usage() {
     cat<<EOF
-Usage: pre-commit [ --reformat] [ --nostage ]
+Usage: pre-commit [ --reformat] [ --nostage ] [ file1 ] [ file2 ] [ file3 ] ...
 
 Format files prior to commit so that they match style guidelines
 
+Normally called before commit, but if files are explicitly specified, reformats
+them in-place without respect to commits.
+
 Options:
-    --reformat   Reformat all files in the source tree, rather than staged files
-    --nostage    Do not add files modified by pre-commit to the staging area
+   --reformat   Reformat all files in the source tree, rather than staged files
+   --nostage    Do not add files modified by pre-commit to the staging area
 EOF
     exit 1
 }
 
 # Parse command-line options
 parse_options() {
+    FILES=()
     REFORMAT=
     NOSTAGE=
+    EXPLICIT_FILES=
+    local endopts=
     for opt in "$@"; do
-        case $opt in
-            --reformat) REFORMAT=1 ;;
-            --nostage)  NOSTAGE=1  ;;
-            *) usage
-        esac
+        if [[ -n $endopts ]]; then
+            FILES+=("$opt")
+        else
+            case $opt in
+                --)         endopts=1  ;;
+                --reformat) REFORMAT=1 ;;
+                --nostage)  NOSTAGE=1  ;;
+                -*)         usage      ;;
+                *)          FILES+=("$opt")
+            esac
+        fi
     done
+    if [[ ${#FILES[@]} -ne 0 ]]; then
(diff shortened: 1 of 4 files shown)


kpgriesser:
I ran this on MacOs. It runs but doesn't do the full conversion. Example:
Original code snippet:
           MachineConfig.MapMemBase, MachineConfig.GMapMemBase,
           MachineConfig.UDbase, MachineConfig.SPMemBase,
           MachineConfig.ControlBase, MachineConfig.ProgBase,
           MachineConfig.EventQueueOffset,

Conversion using pre-commit script: rev/.githooks/pre-commit updown.h
 MachineConfig.MapMemBase, MachineConfig.GMapMemBase, MachineConfig.UDbase, Ma\
chineConfig.SPMemBase, MachineConfig.ControlBase,
           MachineConfig.ProgBase, MachineConfig.EventQueueOffset, (more variables)

Conversion inside rev git repo: clang-format -I updown.h
           MachineConfig.MapMemBase,
            MachineConfig.GMapMemBase,
            MachineConfig.UDbase,
            MachineConfig.SPMemBase,
            MachineConfig.ControlBase,
            MachineConfig.ProgBase,
            MachineConfig.EventQueueOffset,

Is there another step to get the final output I missed?

leekillough:
Were you running it by explicitly specifying filenames, or were you using `git commit` or `pre-commit --reformat`?  The latter two were fixed by my latest push, but explicitly specifying filenames should not have changed.

It goes through each file, whether explicitly specified on the command-line, or determined with `git` commands, and then performs the same steps on each file either way.

If you could edit the `pre-commit` script to say:
```
    set -xeEu -o pipefail
```
(add `x` option) and send me the script's trace output, I could better tell what went wrong, if anything.

leekillough:
If it modified your files, it should have shown (in yellow if on a terminal) what changes it made. It edits the files in-place.

leekillough:
If you're getting different formatting in different directories, it means something is wrong with the `--assume-filename=` handling or something else, such as if the current directory changed in the middle of the script between the time you invoked it and the time it calls `clang-format`, leading to a relative path not working anymore. In that case, I have to change it to figure out absolute paths.

`clang-format` does not produce errors if it cannot find a `.clang-format` file for an `--assume-filename=` argument -- it simply silently falls back to a default style. That `clang-format` bug listed above is valid -- I'm surprised a tool in Clang does not have an option to fine-tune its main purpose, such as explicitly specifying a configuration file.

kpgriesser (commented):
I ran the following checks on the same file: 
- MacOs on a file outside of the rev repo
- Ubuntu 22.04 on a file outside of the rev repo AND a file inside a repo with a conflicting .clang_format file.

All 3 cases  produced the same result which also matched the version generated in the rev repo itself.

This PR is good to go from my perspective.

Issue #25, opened 2022-08-19
SEXT is not used correctly in some instructions


Artlands:

Artlands:
Sorry, my mistake.

PR #264, opened 2024-04-07
Regression suite for stable C++ support methods on REV target

kpgriesser:
In order to port more complex workloads to REV I'm finding it extremely challenging to get functionally correct code using C++ constructs. Seeming trivial code changes can lead to wildly different behaviors.

This PR is intended to nail down a baseline minimum set of C++ code supported on rev and document the associated methods and workarounds (if necessary).

A test script is provided that runs the suite using these compiler flags:
-O0 (expected to pass)
-O2 (expected to fail)
-O2 -DO2_WORKAROUND (expected to pass)

The test suite passes except for a single test using the -O2 compiler optimizations. The failure mode suggests a stack pointer corruption issue.

To run:
```
cd test/udruntime
./run-test.bash
```

Result:
```
Testing -O0 ... Passed
See O0.log
Testing -O2 with O2_WORKAROUND
Passed
See O2_WORKAROUND.log
Testing -O2 without O2_WORKAROUND
Failed
See O2.log
Done with testing
```

The O2_WORKAROUND simply saves a function argument in a static variable which appears to affect the compilation. Without the workaround the core gets a bad pointer from the stack and crashes.

Obviously, this workaround is empirically derived an unreliable so this runtime library has limited use beyond limited prototype code.

### DIFF
--- a/test/udruntime/include/debug.h
+++ b/test/udruntime/include/debug.h
@@ -1,0 +1,202 @@
+/*
+ * Copyright (c) 2021 University of Chicago and Argonne National Laboratory
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Author - Jose M Monsalve Diaz
+ *
+ * Helper functions to add debug comments, error messages and others
+ *
+ */
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+//TODO set using compiler flags only
+#ifndef ASST
+#define ASST
+#endif
+
+#ifdef ASST
+#include "rev-macros.h"
+#include "syscalls.h"
+
+#undef assert
+#define assert TRACE_ASSERT
+#endif
+
+#ifndef DEBUG_MODE
+#define rev_udrt_print( ... )
+#endif
+
+#define __FILENAME__ \
+  ( strrchr( __FILE__, '/' ) ? strrchr( __FILE__, '/' ) + 1 : __FILE__ )
+
+// Macro for output of information, warning and error messages
+#ifdef DEBUG_MODE
+
+#ifndef ASST
+#define UPDOWN_ASSERT( assertion, message, ... )              \
+  {                                                           \
+    for( ; !( assertion );                                    \
+         printf( "[UPDOWN_ASSERT_FAIL: %s:%i] " message "\n", \
+                 __FILENAME__,                                \
+                 __LINE__,                                    \
+                 ##__VA_ARGS__ ),                             \
+         fflush( stderr ),                                    \
+         fflush( stdout ),                                    \
+         assert( 0 ) ) {                                      \
+    }                                                         \
+  }
+#else
+#define UPDOWN_ASSERT( assertion, message, ... ) \
+  { TRACE_ASSERT( assertion ); }
+#endif
+
+#ifndef ASST
+#define UPDOWN_WARNING( message, ... )               \
+  {                                                  \
+    printf( "[UPDOWN_WARNING: %s:%i] " message "\n", \
+            __FILENAME__,                            \
+            __LINE__,                                \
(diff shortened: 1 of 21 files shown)


jleidel:
@kpgriesser should this PR target the UpDown fork?

test/udruntime/include/debug.h
```diff
@@ -0,0 +1,202 @@
...
+
+#define __FILENAME__ \
+  ( strrchr( __FILE__, '/' ) ? strrchr( __FILE__, '/' ) + 1 : __FILE__ )
+
+// Macro for output of information, warning and error messages
+#ifdef DEBUG_MODE
+
+#ifndef ASST
+#define UPDOWN_ASSERT( assertion, message, ... )              \
+  {                                                           \
```
leekillough, line 64: Suggest a `do {} while(0)` construct around the macro.

leekillough: Same with all macros. [See here](https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros)

test/udruntime/include/networkid.h
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  networkid_t( networkid_t&& ) = default;
+
+  networkid_t& operator=( networkid_t& rhs ) {
+    NetworkId = rhs.get_NetworkId();
+    return *this;
+  }
+
+  networkid_t& operator=( networkid_t&& ) = default;
```
leekillough, line 147: Good. You're using the [Rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three).

test/udruntime/include/operands.h
```diff
@@ -0,0 +1,228 @@
...
+
+//END A_SST
+
+namespace UpDown {
+
+template< typename T >
+class RevAllocPolicy {
+public:
+  RevAllocPolicy< T >()  = default;
+  ~RevAllocPolicy< T >() = default;
```
leekillough, line 24: Within the template, template arguments `< T >` are redundant.

test/udruntime/include/operands.h
```diff
@@ -0,0 +1,228 @@
...
+    std::size_t addr = reinterpret_cast< std::size_t >( p );
+    rev_munmap( addr, sizeof( T ) * l );
+  }
+};
+
+template< typename T >
+class StdAllocPolicy {
+public:
+  StdAllocPolicy< T >()  = default;
+  ~StdAllocPolicy< T >() = default;
```
leekillough, line 46: Same here.

test/udruntime/include/revalloc.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+  inline explicit StandardAllocPolicy( StandardAllocPolicy const& ) {
+  }
+
+  template< typename U >
+  inline explicit StandardAllocPolicy( StandardAllocPolicy< U > const& ) {
+  }
+
+  //    memory allocation
+  inline pointer
```
leekillough, line 65: Just FYI: `inline` is implicit for functions defined inside of classes. But maybe you wanted to be verbose.

kpgriesser:
> @kpgriesser should this PR target the UpDown fork?

Yes and No. I'll create a branch in the UpDown fork so that the referenced source code resides there.  I put this in rev to ensure there was nothing in that branch that affected the bugs.  What I would like to have is more targeted testing of the problematic constructs in this code and some kind of guidelines for handling C++ test porting to Rev ( issue 214 is a good example which resulted in special handling requirements for command line argument passing ). 

When I have time I'll repackage smaller test cases and would be glad to own documenting the guidelines. The problems that are sensitive to the compiler and small code changes are the most difficult and time consuming to deal with ( and remain unresolved ).

kpgriesser:
I replicated this branch in updown-accel repository.  This branch should identically match the latest rev and contain the target code of interest.

test/udruntime/include/updown.h
```diff
@@ -0,0 +1,1784 @@
...
+     *
+     * This is equivalent to malloc, it finds a free region that
+     * can allocate the current size, and, if there is free space
+     * left, it creates a new region with the remaining free space.
+     *
+     * @param size size to be allocated in bytes.
+     * @return void* pointer to the allocated memory
+     */
+  void* get_region( uint64_t size, bool global ) {
+    UPDOWN_INFOMSG( "Allocating new region of size %lu bytes", size );
```
leekillough, line 322: It might be better to use `"%" PRIu64` instead of `%lu`. `PRIu64` and similar macros are defined in `<cinttypes>` (C++) or `<inttypes.h>` (C). We have warnings as errors when `printf()` arguments don't match their format's size.

test/udruntime/include/updown.h
```diff
@@ -0,0 +1,1784 @@
...
+    // Iterate over the regions finding one that fits
+
+    //std::map<void *, mem_region_t> *regions_to_use;
+    std::map< void*,
```
leekillough, line 326: Unless you need to traverse all map elements in sorted order, `std::unordered_map` is generally faster than `std::map` (it uses a hash table instead of a balanced binary tree).

kpgriesser:
Closing  this PR and replacing it with updown-accel repo PR #25

test/udruntime/include/updown.h
```diff
@@ -0,0 +1,1784 @@
...
+        void* new_region_addr = static_cast< char* >( used_reg->first ) + size;
+        ( *regions_to_use )[new_region_addr] = { new_size, true };
+        UPDOWN_INFOMSG(
+          "Creating a new region 0x%lX with the remaining size %lu",
+          reinterpret_cast< uint64_t >( new_region_addr ),
+          new_size );
+      }
+    }
+
+    UPDOWN_INFOMSG( "Returning region 0x%lX = {%lu, Used}",
```
leekillough, line 411: See above about `PRIu64`. `PRI` means `printf`, `u` means `unsigned` format, and `64` means 64-bit size. `"%" PRIu64` is the portable way to print `uint64_t`.

test/udruntime/include/updown.h
```diff
@@ -0,0 +1,1784 @@
...
+
+// moved from above to avoid inner classes
+template< class AllocationPolicy >
+class UDRuntime_t : public AllocationPolicy {
+private:
+  struct base_addr_t {
+    volatile ptr_t mmaddr;
+    volatile ptr_t spaddr;
+    volatile ptr_t ctrlAddr;
+    volatile ptr_t progAddr;
```
leekillough, line 598: Just FYI: `volatile` does not guarantee atomicity (`std::atomic` does), and `volatile`'s use is discouraged. I've even seen proposals to deprecate `volatile`.

test/udruntime/include/updown.h
```diff
@@ -0,0 +1,1784 @@
...
+  void     dumpBaseAddrs() {
+#ifndef ASST
+    printf(
+#else
+    rev_udrt_print(
+#endif
+      "  mmaddr    = 0x%lX\n"
+      "  spaddr    = 0x%lX\n"
+      "  ctrlAddr  = 0x%lX\n"
+      "  progAddr  = 0x%lX\n",
```
leekillough, line 925: `"0x%" PRIX64` should be used for maximum portability.

test/udruntime/include/operands.h
kpgriesser: fixed and pushed

test/udruntime/include/operands.h
kpgriesser: fixed/pushed

test/udruntime/include/revalloc.hpp
kpgriesser: fixed/pushed

test/udruntime/include/updown.h
kpgriesser: fix/pushed

test/udruntime/include/updown.h
kpgriesser: will leave as-is for now as performance isn't a concern at this point

PR #266, opened 2024-04-11
Adding memory dumping

rkabrick:
This PR is a draft of a memory dumping function for Rev. There are a few caveats that I'd like everyone's feedback on before thinking about merging it:

Here is an example with the command line args: 
image

Here is an example of what the static mem section of a program prints like:

image

1) Should this eventually work with mem hierarchy? It seems like that may be abysmal from a performance perspective but I can extend it to work if necessary
2) Do yall like the format of the printing? Any critiques are welcome 
3) Are there any 'helper' function version of this you think would be useful? (ie. `PrintPage`, `PrintStack`, etc.? 
4) This is a question for @leekillough but I've overridden the `ostream` operator before for other data structures... Does it make sense / is it possible to override the ostream operator for `RevMem` such that you can also pass a range to print or is there a rule that you don't pass additional args when doing that? This is necessary because if you just override the ostream and someone sets memSize to something like 4GB then you're looking at a 4+GB output getting printed.


rkabrick:
For reference... I've been testing this by changing MemSize to 10MB and then  dumping the entire memory space to a file. Then in vim I use the following search regex to find non-zero regions of memory:

`/0x[0-9a-fA-F]\+:\( [0-9a-fA-F][0-9a-fA-F]\)\{15\} \(00\)\@!/`

skuntz:
I assume this is intended to help with debugging? Not sure how feasible, but the two that come immediately to my mind are print stack and print 8/16/32/64b at a specific address.

jleidel:
> I assume this is intended to help with debugging? Not sure how feasible, but the two that come immediately to my mind are print stack and print 8/16/32/64b at a specific address.

Yep!  Debugging some of the more complex simulations can get pretty difficult.  Its sometimes nice to print the register state at a given clock tick.  This will allow us to differentiate between what we *believe* is in memory versus what is actually there.  

That being said, I agree that it would be really useful to provide the ability to print specific addresses or address ranges.

rkabrick:
> That being said, I agree that it would be really useful to provide the ability to print specific addresses or address ranges.

Yeah the DumpMem function has the following signature: 

`void RevMem::DumpMem( const uint64_t startAddr, const uint64_t numBytes, const uint64_t bytesPerRow )`

So the helper functions I'm asking would basically be functions that implement this under the hood 

> print stack and print 8/16/32/64b at a specific address.
Should be possible with the current signature! You would just call `DumpMem` with the base address of the current executing thread's stack pointer and then pass `_STACK_SIZE_` as the number of bytes. 

If I understand your second point correctly I *think* it should work with the current implementation? If not would you mind showing what you mean with an example

jleidel:
> > That being said, I agree that it would be really useful to provide the ability to print specific addresses or address ranges.
> 
> Yeah the DumpMem function has the following signature:
> 
> `void RevMem::DumpMem( const uint64_t startAddr, const uint64_t numBytes, const uint64_t bytesPerRow )`
> 
> So the helper functions I'm asking would basically be functions that implement this under the hood
> 
> > print stack and print 8/16/32/64b at a specific address.
> > Should be possible with the current signature! You would just call `DumpMem` with the base address of the current executing thread's stack pointer and then pass `_STACK_SIZE_` as the number of bytes.
> 
> If I understand your second point correctly I _think_ it should work with the current implementation? If not would you mind showing what you mean with an example

@rkabrick I think you are correct that the current interface supports the request from @skuntz.  

Would it also be useful to provide a function to print the symbol table?

rkabrick:
> Would it also be useful to provide a function to print the symbol table?

What do you mean? More than the existing printing capabilities inside of RevLoader?

jleidel:
> > Would it also be useful to provide a function to print the symbol table?
> 
> What do you mean? More than the existing printing capabilities inside of RevLoader?

The loader does it using verbosity (which also triggers a lot of other statements being printed).  Possibly a parameter?

donofrio:
Getting this to work with RevMem is quite useful - agreed on printing specific memory addresses. I like the format... that's the standard hexdump format, no need to mess with success. Having this work with memH would be _incredibly_ useful. We should approach that as a separate project as that begs a few fundamental questions like - do you get the cached value when you print or... ?

leekillough:
> This is a question for @leekillough but I've overridden the ostream operator before for other data structures... Does it make sense / is it possible to override the ostream operator for RevMem such that you can also pass a range to print or is there a rule that you don't pass additional args when doing that? This is necessary because if you just override the ostream and someone sets memSize to something like 4GB then you're looking at a 4+GB output getting printed.

You can overload `operator<<` to use your own data structure to output to an `ostream`. We're not really overriding `ostream` -- we are adding additional overloads for `operator<<`. As long as it has at least one of our own user-defined types, it can't conflict with the standard library.

A range can be expresssed as a `std::pair` of iterators or some other single-argument data structure, like `RevMemRange`. `equal_range` and other standard functions return ranges as pairs of iterators (like `begin()` and `end()`).

If you want to globally, within Rev, change the range to be printed, then that can be done by setting up another data structure, which when outputted to `ostream`, takes a range or other Rev-global setting into account. The range could be a RevMem attribute which it refers to when printing.

We're not changing `ostream`. We're adding overloads so that our own data structures can interact with `ostream` in ways we define.

kpgriesser:
The use case I'd like to have, which this should no doubt support, is to be able to dump the initial memory image after the .elf is loaded and the final memory image. By image, I mean picking a certain range and this is a specific REV's view of memory. I use memhierchy pretty much exclusively so, yes, please have it support that.  For future thought consider. an option for a binary format and let hexdump extract and format the data.

leekillough:
> > This is a question for @leekillough but I've overridden the ostream operator before for other data structures... Does it make sense / is it possible to override the ostream operator for RevMem such that you can also pass a range to print or is there a rule that you don't pass additional args when doing that? This is necessary because if you just override the ostream and someone sets memSize to something like 4GB then you're looking at a 4+GB output getting printed.

You can also create IO modifiers if you create your own class derived from `ostream`, say `class rev_ostream : public ostream`. It would be a derived class of `ostream` and it could contain extra data such as memory address ranges. You can define modifiers similar to `std::hex` to change the address range before printing it. Then when you output a `RevMem` to a `rev_ostream`, it can use the range stored in the `rev_ostream`. Something like:

```c++
rev_ostream << RevSetRange(lo, hi) << RevMem << std::endl;
```
But it doesn't necessarily have to be too complicated. We can define our own operator overloads which have `ostream` or a class derived thereof as one of its arguments. This is not overriding `ostream` -- it is defining overloads which connect our data structures with an `ostream`.

rkabrick:
Added the following ecalls: 

```c
REV_SYSCALL( 9000, void dump_mem_range( uint64_t addr, uint64_t size ) );
REV_SYSCALL( 9001, void dump_mem_range_to_file( const char* outputFile, uint64_t addr, uint64_t size ) );
REV_SYSCALL( 9002, void dump_stack( ) );
REV_SYSCALL( 9003, void dump_stack_to_file( const char* outputFile ) );
```

jleidel:
Code changes look very reasonable

rkabrick:
Alright so I've added some more stuff: 

1. Ability to specify ranges in the python configuration (I'm not *in love* with this because it's not terribly explicit in saying "Specifying ranges here will result in memory dumps at the initialization phase & then once again just before the program finishes. Any suggestions are welcome but nonetheless this is what it looks like: 

image

2. Specifying a range here will not go through the `CalcPhysAddr` function which idk how we feel about. If we do go through that function then we risk the range dumps triggering a segfault. 

3. The ranges get dumped to files that match the name they are specified with (with the details of the segment outlined at top:

image

The goal being the ability to do the following post program run:

image

I am currently dumping the values here at two points in `RevCPU.cc`: Right after initialization & then right before the call to `primaryComponentOkToEndSim` 

I'd really appreciate feedback specifically from @kpgriesser and @jleidel if that's possible. Ken, I put your name in a few comments with some of the above points outlined that I would like your feedback on.

include/RevMem.h
```diff
@@ -109,17 +109,31 @@ class RevMem {
...
+    }
+
+    /// MemSegment: Override the less than operator
+    bool operator<( const MemSegment& other ) const {
+      return BaseAddr < other.BaseAddr;
+    }
+
+    /// MemSegment: Override the greater than operator
+    bool operator>( const MemSegment& other ) const {
+      return BaseAddr > other.BaseAddr;
```
leekillough: For completeness, you might want to define `operator>=`, `operator<=`, `operator==`, `operator!=`.

include/RevMem.h
```diff
@@ -361,6 +375,11 @@ class RevMem {
     return FreeMemSegs;
   }
 
+  ///< RevMem: Get DumpRanges vector
+  std::map< std::string, std::shared_ptr< MemSegment > >& GetDumpRanges() {
+    return DumpRanges;
+  }
```
leekillough: You might want to overload a `const` version too:
```c++
const std::map< std::string, std::shared_ptr< MemSegment > >& GetDumpRanges() const {
    return DumpRanges;
}
```
(`const` added to the return reference type and the `*this` implicit object.)

include/RevMem.h
```diff
@@ -445,6 +469,22 @@ class RevMem {
...
+  /// RevMem: Dump the memory contents
+  void DumpMem( const uint64_t startAddr,
+                const uint64_t numBytes,
+                const uint64_t bytesPerRow  = 16,
+                std::ostream&  outputStream = std::cout );
+
+  void DumpValidMem( const uint64_t bytesPerRow  = 16,
+                     std::ostream&  outputStream = std::cout );
+
+  void DumpMemSeg( std::shared_ptr< MemSegment > MemSeg,
```
leekillough: Unless a multithreaded Rev can destroy a `std::shared_ptr<MemSegment>` in another thread and you need to "borrow" `MemSeg` in a thread-safe way, you can use `const std::shared_ptr<MemSegment>& MemSeg` to be faster (it avoids atomic increment/decrement of reference count).

common/syscalls/syscalls.h
kpgriesser, line 1: side note: I think I'll start a PR to coordinate how we assign syscalls. I'm coincidentally using the same ranges which I'll fix in my branch.

src/RevCPU.cc
kpgriesser, line 1: Regarding line 246.  Can you clarify the context for this?
      // FIXME: @Ken I don't think we need these
       Mem->AddDumpRange( segName, startAddr, size )

test/mem_dump/mem_dump.py
kpgriesser, line 1: a nit: change line 8 to have the actual file name mem_dump.py instead of rev-basic-config.py

kpgriesser: similarly for line 4

src/RevCPU.cc
kpgriesser: regarding line 1151+
// @KEN Confirm this is okay... If you would prefer to have the address come after the result of CalcPhysAddr
   // you risk getting a segfault if the address has not been allocated yet
   // FIXME:
: uint64_t translatedStartAddr = startAddr; //CalcPhysAddr( 0, startAddr )

So basically, we get the virtual (aka logical) address in the dump.  I definitely would prefer this over the physical address.  I'm making a mental note for a future enhancement that the trace could provide the virtual and physical address and that we should consider an SST PR to support physical memory dumps ( I think they may already be on this ).

test/syscalls/mem_dump/mem_dump.c
kpgriesser, line 1: Nice and concise!

kpgriesser:
can you add mem_dump test to test/CMakeLists.txt? If we don't want these tests in the regression then can there be a script or makefile in the test directory itself so it's easy to run?

kpgriesser:
If no other issues I would be fine having this merged. I would like to merge it into my updown-accel repo and see if it can help crack the runtime bug.

rkabrick:
Alright finally revisited this. All of your feedback should have been addressed and this *should* be ready or merging

kpgriesser (approved):

jleidel (approved):
good to merge

leekillough (approved):

PR #267, opened 2024-04-24
Fix and add C++ tests

leekillough:
Rename C++ test source files from `.c` to `.cc`
Rename `revalloc.hpp` to `revalloc.h` in line with other headers in Rev
Make a single `revalloc.h` allocator header common to all C++ stl tests
Add the C++ tests to the test suite (they were not being tested)


test/cxx/stl/CMakeLists.txt
rkabrick, line 1: I would add `cxx` tag to these

test/cxx/stl/map/revalloc.hpp
rkabrick, line 1: Do you have any thoughts on moving revalloc.hpp to the `common` directory instead? Seems like we're asking for trouble if we have so many copies of this file in the test directory that then could potentially diverge if we don't keep a good enough eye on it?

rkabrick (approved):
Everything looks good... just a few nits

test/cxx/stl/map/revalloc.hpp
leekillough: It depends on whether we want to make `revalloc.h` internal-only or externally visible.

test/cxx/stl/CMakeLists.txt
leekillough: Added, although the `cxx` tag handling was broken (it was looking for `cpp` instead of `cxx`), and if it was done correctly, there would be a `sbrk` memory error at the start of the program, which you say you have another fix for.

I've changed the tag name to `c++` because it's clearer; I've commented out the code which changes to `_start` for C++, until your other fix is in.

test/cxx/stl/map/revalloc.hpp
donofrio: moving to common works for me - no issues making it externally visible

leekillough: Note that with this change, there is only one `revalloc.h` in the `test/cxx/stl` directory. Since it provides "allocator" functionality, it seems like its use would be mostly limited to C++ STL tests, so one copy in the `test/cxx/stl` directory seems fine.

leekillough: I'd rather not pollute `common` with it -- do we anticipate other Rev clients in need of our special allocator? If not, keeping one in the `test/cxx/stl` directory seems better to me.

donofrio: as long as there is only one, then that works. It is probably better there than common

leekillough: I suppose if any other tests use C++ STL containers, it might be necessary.

leekillough: I moved it to common, on the anticipation that other tests which aren't specifically STL tests but nonetheless use STL containers, might need to use it or something like it. Having it tucked down in `test/cxx/stl` may make it harder to find.

leekillough: I can revert and go back to `test/cxx/stl`

leekillough: Poll?

rkabrick: I prefer common because it'll be next to the rev specific implementations of system calls. Seems like those two things should be near one another so that when people see the syscalls implementation they also realize we have a custom allocator.

leekillough: Unless there's any objections, I'll merge this (with the `common/include/revalloc.h`) tomorrow.

I nudged up the `big_loop` timeout because the CI was failing at just over 100 seconds.

PR #268, opened 2024-04-26
Add struct declarations to remove warnings

leekillough:
Removes  warnings such as:
```
syscalls.h:420:98: warning: 'struct old_timespec32' declared inside parameter list will not be visible outside of this definition or declaration
````


jleidel (approved):
good to merge

PR #269, opened 2024-04-26
change clang-format to not align declarations across blank lines

leekillough:
This changes the `clang-format` to not align declarations across blank lines.

Previously, something like this could occur:
```
extern std::vector< void ( * )() > fenv_tests;
unsigned                           failures;

int                                main() {
}
```
`main()` should not be indented like that. This change changes the `.clang-format` so that declarations are not lined up if there is a blank line between them. A few files are affected with only whitespace changes.


rkabrick (approved):
Looks good

PR #271, opened 2024-04-30
Fix copyright date handling

leekillough:
A 1-character bug in `.githooks/pre-commit` prevented copyright dates from being modified correctly.

This fixes `.githooks/pre-commit` and runs it on the entire tree, fixing some copyright dates.


rkabrick (approved):
No problems I see... Should probably go in before the other PR you have open

PR #272, opened 2024-04-30
clang-format changes to decrease breaks

leekillough:
Increase `ColumnLimit` from 80 to 132 characters

80 and 132 are arbitrary numbers derived from Fortran 77 and Fortran 90 line length limits; any finite `ColumnLimit` may be specified, or `0` for no limit whatsoever

```
Make function calls which cannot fit on one line(
    be
    broken
    into
    block
    style
  );
```
Remove spaces after `<` and before `>` in templates

Only align declarations / functions / variables across one blank line

Align trailing comments if they are not too far apart (one blank line max)

Allow function definitions on a single line only if they're inline inside of a class

**Reviewers: you had some concerns about certain C++ constructs/files. Please see if the new formatting is acceptable. If not, I can try to tweak it some more.**


donofrio:
This is a definite improvement. I like the formatting - the line length works well.

donofrio (approved):

rkabrick (approved):
I agree with Dave, this looks much better

jleidel:
> I agree with Dave, this looks much better

i agree too.  good to merge

Issue #273, opened 2024-05-01
[FEATURE REQUEST] Documentation 


rkabrick:
**Rev Feature Request**
I tried building the documentation for Rev for the first time and was unable to do so. I know we have the Doxygen skeleton kind of throughout the repo but `-DBUILD_DOCUMENTATION` ends in the following error. 

<img width="725" alt="image" src="https://github.com/tactcomplabs/rev/assets/31963805/d146aca4-03d0-42bc-8165-27a45c47e6c0">

**Version of SST**
N/A

**Additional context**
See above

jleidel:
In progress

jleidel:
Fix in 0d69918

jleidel:
confirmed fix in CI

PR #274, opened 2024-05-03
fixing doxygen builds; update for Issue273

jleidel:
this fixes the documentation bug (Issue #273) by adding the appropriate scripts, fixing the name of the doxygen config file and updating the README documentation

### DIFF
--- a/scripts/cleanmanpages.sh
+++ b/scripts/cleanmanpages.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+# -- cleans the unnecessary directory reference manpages _*.3
+
+BASE=$1
+
+rm -Rf $BASE/_*.3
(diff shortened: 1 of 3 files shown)


scripts/cleanmanpages.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# -- cleans the unnecessary directory reference manpages _*.3
+
+BASE=$1
+
+rm -Rf $BASE/_*.3
```
leekillough, line 6: If this is under `build/`, we already `.gitignore` it; if `.3` files are placed somewhere else in the tree, we might have to add `*.3` to `.gitignore`.

Otherwise looks good.

leekillough (approved):

scripts/cleanmanpages.sh
jleidel: Its falls under `doxygen`.  Maybe we add a `.gitignore` with `.3` in that directory as well

PR #278, opened 2024-05-08
Clean up revalloc.h

leekillough:
The main reason for this PR is that the unused function `mynew()`, which is neither `inline` nor `template`d, caused multiply-defined symbol errors.

Other cleanup:
* Remove some redundant comments
* Remove redundant `inline` keyword
* Remove redundant `public:` specifiers
* Simplify automatically-generated constructors/destructors with `= default` syntax
* Add some needed `#include`s

### DIFF
--- a/common/include/revalloc.h
+++ b/common/include/revalloc.h
@@ -1,30 +1,16 @@
-#include "syscalls.h"
-#include "unistd.h"
-#include <limits>
-#include <memory>
-#include <stdlib.h>
-
 #ifndef __REV_REVALLOC__
 #define __REV_REVALLOC__
 
-/*void* operator new(std::size_t t){
-     void* p = reinterpret_cast<void*>(rev_mmap(0,
-              t,
-              PROT_READ | PROT_WRITE | PROT_EXEC,
-              MAP_PRIVATE | MAP_ANONYMOUS,
-              -1,
-              0));
-    return p;
-}*/
-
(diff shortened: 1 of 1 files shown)


common/include/revalloc.h
```diff
@@ -109,55 +87,48 @@ class Allocator : public Policy {
...
-  template<typename U>
-  inline Allocator( Allocator<U> const& ) {}
+  explicit Allocator()          = default;
+  ~Allocator()                  = default;
+  Allocator( Allocator const& ) = default;
 
-  template<typename U, typename P>
-  inline Allocator( Allocator<U, P> const& rhs ) : Policy( rhs ) {}
+  template<typename... U>
+  Allocator( Allocator<U...> const& rhs ) : Policy( rhs ) {}
```
donofrio, line 100: Making this variadic is interesting - why just this constructor?

leekillough: > Making this variadic is interesting - why just this constructor?

The other constructors like copy constructor cannot be templated and still be considered "the" copy constructor. Some constructors need a very specific signature and cannot be templated.

leekillough: Also note that non-templated functions are always rated higher in priority than templated functions. If a function/constructor call can exactly match a non-templated function overload, then it takes priority over all templated overloads. C++ legalism.

donofrio (approved):
Looks good to me

PR #280, opened 2024-05-15
Fix EBREAK Decoding

leekillough:
Right now the decoder shifts the instruction right 19 bits to get the 12-bit immediate value, when it should be 20 bits. This causes `EBREAK` to not decode correctly. We already have a macro for decoding the 12-bit immediate.


donofrio (approved):
Good catch - better to use the macro consistently

PR #281, opened 2024-05-21
Clean up C++ classes

leekillough:
- For classes which have dynamically-allocated memory or pointer members, use the [Rule of 3](https://en.cppreference.com/w/cpp/language/rule_of_three) and explicitly delete the implicit copy constructor and implicit copy assignment operator. This prevents the allocated memory from being leaked, and will get a compilation error if you attempt to copy or assign to the class. Rev's classes with pointer members almost never need to be implicitly copied or assigned to.

- Add explicit `{}` default initializers to class members, or add `const` to the class member if it only needs to be initialized once, so that it's an error if it's not initialized. This achieves 2 things:

1. Avoids uninitialized data from being used if the member is not initialized in a constructor: `0`, `false`, `nullptr` or a default value will be used if a constructor does not initialize the member; if the constructor initializes the member, the `{}` default initializer is ignored and the constructor takes precedence -- no double initialization will occur. If the member is a class which has its own default initializer, then this calls the default initializer just as before, just with explicit `{}` syntax.

   (One note: In `CompletionRec_t`, the `data` pointer value is initialized in the body of the constructor with `memcpy()` in case its size is different from the `void*` pointer argument. `data` is now initialized to `0` before the constructor body is run, and then the `memcpy()` is run.)

2. Removes compiler warnings about not initializing members in constructors. These warnings are enabled by `-Weffc++` but I didn't add it, since [it is somewhat strict and based on a book](https://stackoverflow.com/a/11529328/788330). I turned it on only to find these changes.

- Reformat the spacing, such as by joining long lines into one long line and then re-running `clang-format`. For lines which were already split to 80 characters or less, clang-format usually won't rejoin them to 132 characters unless you explicit join them yourself and re-run `clang-format`.

### DIFF
--- a/common/syscalls/syscalls.h
+++ b/common/syscalls/syscalls.h
@@ -661,6 +661,16 @@ typedef unsigned long int rev_pthread_t;
 // void *restrict arg);
 REV_SYSCALL( 1000, int rev_pthread_create( rev_pthread_t* thread, void* attr, void* fn, void* arg ) );
 REV_SYSCALL( 1001, int rev_pthread_join( rev_pthread_t thread ) );
+
+// ==================== REV MEMDUMP
+REV_SYSCALL( 9000, void dump_mem_range( uint64_t addr, uint64_t size ) );
+REV_SYSCALL( 9001, void dump_mem_range_to_file( const char* outputFile, uint64_t addr, uint64_t size ) );
+REV_SYSCALL( 9002, void dump_stack( ) );
+REV_SYSCALL( 9003, void dump_stack_to_file( const char* outputFile ) );
+REV_SYSCALL( 9004, void dump_valid_mem( ) );
+REV_SYSCALL( 9005, void dump_valid_mem_to_file( const char* outputFile ) );
+REV_SYSCALL( 9006, void dump_thread_mem( ) );
+REV_SYSCALL( 9007, void dump_thread_mem_to_file( const char* outputFile ) );
 // clang-format on
 
 #endif  //SYSCALL_TYPES_ONLY
--- a/include/RevCPU.h
+++ b/include/RevCPU.h
@@ -203,28 +203,29 @@ public:
     {"TLBMisses",           "TLB misses",                                           "count",  1},
     )
 
+  // clang-format on
+
   // Passed as a function pointer to each RevCore for when they encounter a function that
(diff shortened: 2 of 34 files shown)


kpgriesser:
I'm learning new things with every PR but if I'm writing new code, I'm still a still little confused regarding when to use {}, literal, or nothing for simple types. It does seem to make the code a little more confusing to read as well. As an example,

   uint32_t                            ParentID;                     // Parent ID
   uint64_t                            StackPtr{};

If StackPtr were I complex struct I would not look twice but since it is a simple type it begs the question why StackPtr and not ParentID. If all I'm reviewing is the header file I have no more information.

leekillough:
`ParentID` must be initialized in the constructor; a lot of these adding `{}` was driven by `-Weffc++` which did not report `ParentID` being uninitialized in a constructor. But if consistency is desired, `{}` can be added at no performance loss because if the constructor initializes the member, it takes precedence over the default initializer.

The `{...}` is called "uniform initialization syntax" and exists since C++11. Its differences with `()` (direct) and `=` (copy) initialization:
1. Narrowing conversions (e.g. `double` -> `float`) are not allowed.
2. If a class type has a constructor which takes a `std::initializer_list`, it will be called rather than a constructor with the number and type of arguments between `{}`. That is how `std::vector` can be initialized as `v(n, x)` (`n` copies of `x`) and `v{x, y}` (a vector with elements `x` and `y`).
3. It solves the "Most Vexing Parse" problem, where something like `int x();` is taken as a function declaration instead of an `int` variable with an empty initializer. `int x{};` is not a function declaration and there's no ambiguity.
4. `variable{}` provides value initialization (filled with zero bytes) for aggregates and fundamental scalar types.

For more info, see: https://herbsutter.com/2013/05/09/gotw-1-solution/

rkabrick:
Once you merge in the most recent mem dump changes this LGTM

kpgriesser:
I don't have any particular concerns about this PR other than it's touching a lot of files so I'd like to do one big merge in my other repos.  

Will -Weffc++ be part of our build flow so we all follow the same coding style and don't slide backwards?

leekillough:
`-Weffc++` is unlikely to become a permanent option.

This PR is not necessary.

kpgriesser (approved):

PR #283, opened 2024-05-23
Add two more syscall structs

leekillough:
Two more missing `struct` declarations were discovered in C++20 mode.


leekillough:
This PR causes build errors in Rev. Another fix will need to be considered for C++20 compatibility.

PR #284, opened 2024-05-28
Updates for sst 14.0.0 support

kpgriesser:
This PR is to track changes related to updating REV to support the latest SST release v14.0.0.  The current release notes identify deprecated functions that may affect REV.

Removed previously deprecated functions: Link::sendInitData(), Link::recvInitData(), SimpleNetwork::sendInitData(), SimpleNetwork::recvInitData() and BaseComponent::getSimulation() have been removed.

The SST_ELI_REGISTER_SUBCOMPONENT_DERIVED and SST_ELI_REGISTER_MODULE_DERIVED macros are removed

The SimpleMem interface is removed. Use the StandardMem interface instead.

The Event::Handler and Clock::Handler handler types are deprecated in favor of Event::Handler2 and Clock::Handler2. The new types support checkpoint/restart. 

The release notes are located here:
https://sst-simulator.org/SSTPages/SSTmicroRelease_V14dot0dot0/ )

### DIFF
--- a/src/RevNIC.cc
+++ b/src/RevNIC.cc
@@ -63,11 +63,13 @@ void RevNIC::init( unsigned int phase ) {
       req->dest                                    = SST::Interfaces::SimpleNetwork::INIT_BROADCAST_ADDR;
       req->src                                     = iFace->getEndpointID();
       req->givePayload( ev );
-      iFace->sendInitData( req );
+
+      //iFace->sendInitData( req );  // removed for SST 14.0.0
+      iFace->sendUntimedData( req );
     }
   }
-
-  while( SST::Interfaces::SimpleNetwork::Request* req = iFace->recvInitData() ) {
+  //while( SST::Interfaces::SimpleNetwork::Request* req = iFace->recvInitData() ) {
+  while( SST::Interfaces::SimpleNetwork::Request* req = iFace->recvUntimedData() ) {  // SST 14.0.0
     nicEvent* ev = static_cast<nicEvent*>( req->takePayload() );
     numDest++;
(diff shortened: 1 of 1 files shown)


kpgriesser:
In RevNIC.cc there were single instances of  Link::sendInitData() and Link::recvInitData() which were changed to SimpleNetwork::sendInitData() and SimpleNetwork::recvInitData(), respectively. With this change REV appears to compile and run correctly.

For due diligence I searched for the other deprecated functions and macros and found the following:

  getSimulation() - not used

  SST_ELI_REGISTER_SUBCOMPONENT_DERIVED and SST_ELI_REGISTER_MODULE_DERIVED - not used

  SimpleMem interface - not used ( not to be confused with memHierarhcy.simpleMem used in our configuration files ).
 
  Event::Handler and Clock::ClockHandler - present but left unchanged for now. These are deprecated and not removed so we can swap these out later.

  I've tested this on MacOs only. Will run on other OS's as well but IMHO this is a low risk PR

kpgriesser:
Finished comparability testing across MacOs, Gizmo(Redhat), and Ubuntu using sst 13.1.0 and sst.14.0.0. All combinations compiled and ran as expected ( so this PR should be safe for existing users on 13.1.0 )

leekillough (approved):

jleidel:
can you retarget to the `devel` branch?

kpgriesser:
yes, done - 1 file change make much more sense than over 400!

rkabrick:
I'd definitely prefer the 400 file changes but I suppose this will do! 

LGTM

PR #286, opened 2024-06-03
Add .gitignore entries to ignore memory dump test files

leekillough:
Some temporary files are generated by the memory dump tests. This ignores them.


rkabrick:
Whoops I definitely should’ve done this. Looks good

PR #287, opened 2024-06-05
Handle load flags (e.g., sign/zero extension, NaN boxing) at the end of ReadMem() regardless of whether RevMem or MemH is used

leekillough:
Handle load flags (e.g., sign/zero extension, NaN boxing) at the end of `ReadMem()` regardless of whether `RevMem` or `MemH` is used.

Previously, the integer/floating-point load _execute_ functions performed the operation on the destination register regardless of whether `RevMem` or `MemH` was used, causing a garbage temporary result in the destination register if `MemH` was used.

Use `memcpy()` and separate 1/2/4-byte variables to make the code C++-compliant and independent of host machine's endianness. Previously, the code assumed a little-endian host and performed undefined behavior with pointer casting, which violates the C/C++ strict aliasing rule (it is undefined behavior to cast a pointer from one type to another and then dereference it, unless the second type is a `char`).

Add another pseudoinstruction: `ra` as destination register is implicit in `jal` instruction -- `objdump` and other disassemblers do not list `ra` as a destination register but make it implicit by leaving it out of the disassembly -- only non-`ra` destination registers are explicitly listed in disassembly.


include/RevRegFile.h
```diff
@@ -32,11 +32,12 @@ namespace SST::RevCPU {
 struct RevInst;
 
 /// BoxNaN: Store a boxed float inside a double
-inline void BoxNaN( double* dest, const void* value ) {
+inline void BoxNaN( double* dest, const float* value ) {
   uint32_t i32;
-  memcpy( &i32, value, sizeof( float ) );                 // The FP32 value
+  memcpy( &i32, value, sizeof( i32 ) );                   // The FP32 value
```
rkabrick, line 37: What's the difference here? Wouldn't it be the same?

rkabrick (approved):
Looks good to me... If you wouldn't mind educating me on what I asked about in my comment that would be neat : )

include/RevRegFile.h
leekillough: Just a good programming practice; when copying into a variable with `memcpy()`, it is good to make the size of the copy equal to the size of the variable being copied into.

leekillough: As I recall, there is a `static_assert` making sure that the sizes of `float` and `i32` are the same.

include/RevMemCtrl.h
kpgriesser, line 1: I -really- like seeing immediate values replaced by enums (e.g. b010 replaced by RevFlag::F_NONCACHEABLE )

leekillough: Yes, this was another change I neglected to mention.

kpgriesser (approved):

Issue #285, opened 2024-06-01
Stack buffer initialization corrupts memory.


ldalessa:
**Describe the bug**

When I allocate and initialize a small `char` buffer on the stack it seems to be corrupting memory, in particular the `argv` state. My understanding is that `_STACK_SIZE_` is `1MB` so I don't think that's the problem.

**To Reproduce**

SST stuff

```shell
$ sst --version
SST-Core Version (-dev, git branch : devel, SHA: cb2e3206a8af52702b23cd15c81a6ea0a6b3cfa4)
$ sst-config --CXX
mpic++ -std=c++17
$ sst-config --ELEMENT_CXXFLAGS 
 -std=c++17 -fPIC -DHAVE_CONFIG_H -I/home/ldalessa/local/include
```

REV stuff (configured with `-DCMAKE_CXX_COMPILER=mpic++ -DCMAKE_BUILD_TYPE=Release`).

```shell
$ mpic++ --version
g++ (Debian 13.2.0-25) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ git status
On branch devel
Your branch is up to date with 'origin/devel'.

nothing to commit, working tree clean
$ git show HEAD
commit 8292c73900119412eb63b1d1a8f059e7c0efa2d0 (HEAD -> devel, origin/devel)
Merge: 27249d0 5ffaac7
Author: Ryan Kabrick <31963805+rkabrick@users.noreply.github.com>
Date:   Thu May 30 16:31:03 2024 -0400

    Merge pull request #266 from tactcomplabs/mem-dump
    
    Adding memory dumping
```

Test program. `volatile` to prevent the loop from being removed, `noipa` to prevent the call to `foo` from being removed, I split it into a function `foo` to make it easier for me to look at the asm, but that doesn't impact the bug as long as the loop isn't removed by the dead-code passes.

```c++
extern "C"
[[gnu::noipa]]                                                                                                                                                                                                                                  
void foo() {                                                                                                                                                                                                                                  
    volatile char data[256];                                                                                                                                                                                                                  
    for (int i = 0; i < 256; ++i) {                                                                                                                                                                                                           
        data[i] = '\0';                                                                                                                                                                                                                       
    }                                                                                                                                                                                                                                         
}                                                                                                                                                                                                                                             
[...]

Issue #214, opened 2023-11-29
Argument passing / atoi conversion  sensitivity to compiler output


kpgriesser:
**Describe the bug**
 The test incorrectly attempts to read from location 0 during command line argument parsing. This test is sensitive to how the code is compiled so a passing and failing case are provided.

**To Reproduce**

cd test/argc_bug
make clean; make
Comment out line 20 of argc_bug.c to provided a passing case to compare against.
argc_bug.c:20 #define INDUCE_ARGC_BUG

All issues must be accompanied by: 
- The version of SST utilized (`sst --version`; `sst-config --CXX`; `sst-config --ELEMENT_CXXFLAGS`)
SST-Core Version (13.0.0)

- The source code branch being used
- The last commit hash
2f2bda7 (HEAD -> devel, origin/devel) argc bug test

- The compiler utilized to build Rev
Apple clang version 15.0.0 (clang-1500.0.40.1)

- The compiler utilized to build the test
riscv64-unknown-elf-g++ (g2ee5e430018-dirty) 12.2.0

- The python or JSON simulation input
- C/CXX test source (or mention which test in the current source tree failed)

cd test/argc_bug
make clean; make

Comment out line 20 of argc_bug.c to provided a passing case to compare against.
argc_bug.c:20 #define INDUCE_ARGC_BUG

**Expected behavior**

Test executes with no exceptions

**Trace**
Failing trace summary:
-----------------------
The stack location at 0x3ffffbac is written with a zero and later used as a pointer into memory:

<_strtol_l.constprop.0>
[28000] *I 0x103c8:ec061101    c.addi  sp, -32           0x3ffffbcc<-sp sp<-0x3ffffbac
[29000] *I 0x103ca:e822ec06    c.sdsp  ra, 24(sp)        0x3ffffbac<-sp 0x10298<-ra [0x3ffffc6c,8]<-0x0000000000010298
[30000] *I 0x103cc:e426e822    c.sdsp  s0, 16(sp)        0x3ffffbac<-sp 0x3ffffb94<-s0 [0x3ffffc2c,8]<-0x000000003ffffb94
[31000] *I 0x103ce:e04ae426    c.sdsp  s1, 8(sp)         0x3ffffbac<-sp 0x4<-s1 [0x3ffffbec,8]<-0x0000000000000004
[...]

PR #288, opened 2024-06-11
Complete overhaul of argv handling

leekillough:
This fixes https://github.com/tactcomplabs/rev/issues/285.

- The wrong value of `argv` was passed to `main()` in a lot of cases, causing `argv[]` to be corrupted when other variables used the same memory it did.
- The code previously used hard-coded values like `1024` for the `argv` space. Now the space for `argv` is computed and allocated dynamically.
- The code previously used `MemTop` and `StackTop` and played around with their values in unsafe ways. Now the code computes the space it needs for `argv`, and moves the `StackTop` down that many bytes, like pushing `argv[]` and all its pointed-to strings on the stack.
- It used a `SetNextThreadMemAddr()` function to have a delayed effect to setting the thread's top memory address, when `argv` is shared across all threads and the `StackTop` only needs to be computed once and then shared among threads (each new thread creates a new segment).
- Previously the `Args` and `exe` parameters were stored in a lot of different classes. Now, they are simply passed to the `RevLoader` constructor, which passes them to `LoadElf()`, which passes them to `LoadProgramArgs()`. There is no need to store them as class variables, nor to have functions to query them.
- On RV32, `argv[]` needs to be an array of contiguous 4-byte pointers. Previously, it was always 8-byte pointers. The `LoadProgramArgs()` has been templatized with an `XLEN` template parameter which is determined based on ELF32 or ELF64.
- `WriteCacheLine()`'s data pointer was not marked `const` even though it is read-only.
- `RevCore`, `RevOpts`, and `RevLoader` had multiple copies of `splitStr()`, which is used to tokenize strings. Since `RevOpts.h` is commonly known to `RevCore` and `RevLoader`, and since tokenizing strings belongs more properly in "Options" handling, I replaced the 3 `splitStr()`'s with one in `RevOpts` which is much simpler and less buggy, since it uses `strtok_r()`. It was also generalized to accept a string of delimiter characters such as whitespace characters, rather than a single character.
- The Processor Objects were initialized identically in both sides of an `if` depending on whether a coprocessor is present. The processor objects can be unconditionally initialized before the coprocessor objects are optionally initialized.
- The `args` parameter can now be passed to Rev as a true array of values, arrays being written in Python's syntax. If the `--args` parameter does not, after skipping whitespace, start with a `[`, then it is assumed to be a string of whitespace-separated arguments, and the old Rev behavior applies. But if it starts with a `[`, then it is interpreted as a true array, and each argument may contain spaces in it.
- In `RevLoader::LoadElf`, `eh64` was a pointer which was cast from `membuf` and then dereferenced. Besides violating "strict aliasing", it made `eh64` a dangling pointer when `membuf` was unmapped, and to tell whether to use RV32 or RV64 `argv`, I wanted to test it near the end of `LoadElf`. So I made `eh64` a scalar variable, used `memcpy()` to set it safely, and then it was good until the end of `LoadElf` regardless if `membuf` had been unmapped.
- I added two test cases which I verified fail on `devel` currently, but pass with this PR:
 1. https://github.com/tactcomplabs/rev/issues/285
 2. A test which verifies `argv`'s layout in memory on RV32.


include/RevLoader.h
```diff
@@ -355,11 +348,8 @@ class RevLoader {
   /// Loads a 64bit Elf binary
   bool LoadElf64( char* MemBuf, size_t Size );
 
-  ///< Splits a string into tokens
-  void splitStr( const std::string& s, char c, std::vector<std::string>& v );
-
   ///< Breaks bulk writes into cache lines
-  bool WriteCacheLine( uint64_t Addr, size_t Len, void* Data );
+  bool WriteCacheLine( uint64_t Addr, size_t Len, const void* Data );
```
rkabrick, line 352: Probably a stupid question but how does `const void*` actually work? No different than something like a const char*?

include/RevMem.h
```diff
@@ -94,7 +94,7 @@ class RevMem {
     bool contains( const uint64_t& vBaseAddr, const uint64_t& Size ) {
       // exclusive top address
       uint64_t vTopAddr = vBaseAddr + Size - 1;
-      return ( this->contains( vBaseAddr ) && this->contains( vTopAddr ) );
+      return contains( vBaseAddr ) && contains( vTopAddr );
```
rkabrick, line 97: Is the version with `this` incorrect? Or is the new version more correct

include/RevMem.h
```diff
@@ -408,8 +405,7 @@ class RevMem {
   uint64_t TLSBaseAddr       = 0;                   ///< RevMem: TLS Base Address
   uint64_t TLSSize           = sizeof( uint32_t );  ///< RevMem: TLS Size (minimum size is enough to write the TID)
   uint64_t ThreadMemSize     = _STACK_SIZE_;        ///< RevMem: Size of a thread's memory segment (StackSize + TLSSize)
-  uint64_t NextThreadMemAddr = memSize - 1024;      ///< RevMem: Next top address for a new thread's memory
-                                                    //           (starts at the point the 1024 bytes for argc/argv ends)
+  uint64_t NextThreadMemAddr = memSize;             ///< RevMem: Next top address for a new thread's memory
```
rkabrick, line 408: This is much better

src/RevCPU.cc
```diff
@@ -158,22 +158,19 @@ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
   Mem->SetMaxHeapSize( maxHeapSize );
 
   // Load the binary into memory
-  // TODO: Use std::nothrow to return null instead of throwing std::bad_alloc
-  Loader = new RevLoader( Exe, Args, Mem, &output );
+  Loader = new( std::nothrow ) RevLoader( Exe, Opts->GetArgv(), Mem, &output );
```
rkabrick, line 161: What does the nothrow do?

src/RevCPU.cc
```diff
@@ -387,6 +378,22 @@ RevCPU::~RevCPU() {
...
+  std::string args     = params.find<std::string>( "args" );
+  auto        nonspace = args.find_first_not_of( delim );
+  if( nonspace == args.npos || args[nonspace] != '[' ) {
+    RevOpts::splitStr( args, delim, arglist );
+  } else {
+    params.find_array( "args", arglist );
+  }
+  return arglist;
+}
+
```
rkabrick: Does this accomplish what we were talking about or does it require changes to the tests that currently use CL Args?

src/RevLoader.cc
```diff
@@ -445,128 +437,103 @@ bool RevLoader::LoadElf64( char* membuf, size_t sz ) {
...
   //
-  // Note that the addresses are written contiguously
+  // Note that the argv[i] addresses are written contiguously.
   //
-  // Next, we setup the prologue header just above the base stack pointer
-  // that contains the layout of this array.  This includes our ARGC value
-  // We start this block of memory at SP+48 bytes as follows:
+  // The size of each argv[i] pointer is equal to XLEN.
+  // It is necessary that argv[i] be contiguous in memory across i
+  // and that sizeof(argv[i]) == sizeof(char*) == XLEN.
```
rkabrick, line 477: Probably another stupid question but this is just saying the entries inside of argv are all character pointers correct?

rkabrick (approved):
1. Really great work... This looks infinitely better than what was in the code originally
2. I know @kpgriesser had an open issue with some command line arg idiosyncrasy... I'd be curious to see if this remedies his issue
3. I don't think there were any logical bugs I saw so feel free to merge without answering my little questions along the way

include/RevLoader.h
leekillough: Any pointer (to non-`volatile`) can be implicitly converted to `const void*`.

A pointer to `const T` cannot be implicitly converted to `void*` since it drops `const`, hence the reason for this change.

I was trying to write `s.c_str()` to write out a string, and got an error that its `const char*` could not be converted to `void*`.

include/RevMem.h
leekillough: Both are correct and the same, but the `this->` is redundant and makes it harder to read. `this->` is implicit  when accessing members of a class or its base classes from within a class's own member functions.

About the only time you need to use `this->` is when you have a `template` class trying to access a base class's members when the base class is dependent on its template parameter(s). Or if you have a local variable shadowing a class variable/function and you want to access the class variable/function instead of the local variable.

src/RevCPU.cc
leekillough: Passing `std::nothrow` to `new` in placement-new syntax causes it to return `nullptr` if the allocation fails, similar to how `malloc()` behaves, instead of throwing a `std::bad_alloc`, which is how `new` usually behaves.

The original code tested the result of `new` against `nullptr` without specifying `std::nothrow`, so it was never going to catch a failed allocation --  `new` would have just raised a `std::bad_alloc` exception and unwound the stack until a `catch` caught it. If nothing caught it, the program would abort.

src/RevCPU.cc
leekillough: It accomplishes what we talked about -- it first treats the `args` parameter as a string, and if it does not (after leading whitespace) start with a `[`, treats it the way that the old Rev and Tests handled it. If it starts with a `[`, the `args` parameter is treated as an array.

This is better than basing it on the number of arguments in the array, since sometimes you'll want only one argument, even if it has spaces in it.

The existing `argc` test uses the space-delimited format `"one two three"` and this change maintains backwards compatibility.

src/RevLoader.cc
leekillough: Yes, but the main point of that comment was that the entries inside `argv` (top-level) have to be contiguous (no gaps) and each one needs to be `XLEN` in size, or the size of a pointer. `argv[]` cannot contain 8-byte entries on RV32, or 4-byte entries on RV64. Hence the template.

kpgriesser (approved):
This also fixes my issue #214

donofrio (approved):
Huge improvement! Thank you Lee - the arg passing/parsing has been the source of some vexing bugs

include/RevMem.h
leekillough: (I know this PR is closed, but just some education:)

```c++
template<typename T>
class base {
public:
  bool mymember(){
    return true;
  }
};

template<typename T>
class derived : public base<T> {
public:
  bool something(){
    return mymember();         // Does not work: base class is dependent on template parameter T and won't be found
    return this->mymember();   // Works. "this" establishes the template parameter dependency
  }
};

```

rkabrick: Makes a lot of sense!

Issue #289, opened 2024-06-18
lwsp instruction execution generating bad address outside of stack


kpgriesser:
**Describe the bug**
This test case is failing on a c.lwsp instruction which appears to be making an incorrect address calculation. The code was isolated  from a more complex C++ program using the -O1 option. This hopefully isolates an ongoing stack corruption problem.

**To Reproduce**

  git checkout lwsp
  cd test/lwsp
  make clean && make

- The version of SST utilized (`sst --version`; `sst-config --CXX`; `sst-config --ELEMENT_CXXFLAGS`)
   SST-Core Version (14.0.0)
- The source code branch being used
   lwsp
- The last commit hash
   8ad3667 (HEAD -> lwsp, origin/lwsp) bug test for lwsp
- The compiler utilized to build Rev
   Apple clang version 15.0.0 (clang-1500.3.9.4)
- The compiler utilized to build the test
   riscv64-unknown-elf-g++ (g2ee5e430018-dirty) 12.2.0

**Expected behavior**
Test self-checking should pass.

**Trace**
RevCPU[cpu0:Render:2000]: Core 0; Hart 0; Thread 1; *I 0x11206:7179      + c.addi16sp sp, -48	 0x3ffffffc<-sp sp<-0x3fffffcc 
RevCPU[cpu0:Render:3000]: Core 0; Hart 0; Thread 1; *I 0x11208:f422        c.sdsp  s0, 40(sp)	 0x3fffffcc<-sp 0x14028<-s0 [0x3ffffff4,8]<-0x0000000000014028 
RevCPU[cpu0:Render:4000]: Core 0; Hart 0; Thread 1; *I 0x1120a:1800        c.addi4spn s0, sp, 48	 0x3fffffcc<-sp s0<-0x3ffffffc 
RevCPU[cpu0:Render:5000]: Core 0; Hart 0; Thread 1; *I 0x1120c:67c5        c.lui   a5, 0x11	 a5<-0x11000 
RevCPU[cpu0:Render:6000]: Core 0; Hart 0; Thread 1; *I 0x1120e:5387b783    ld      a5, 1336(a5)	 0x11000<-a5 0x80000000000aced1<-[0x11538,8] 
[...]

PR #290, opened 2024-06-18
Fix immediate decoding for lwsp and flwsp (#289)

kpgriesser:
This fixes bug #289 in which an lwsp instruction execution had an incorrect immediate field calculation. The root cause was due to missing the '0b' prefix on a binary constant.  I also fixed flwsp which had the same problem ( but does not have a test ).

Note: Only lines 446 and 460 in RevCore.cc changed. The rest are auto-formatting changes.


leekillough (approved):
LGTM

donofrio (approved):
Great find! Looks good to merge in

rkabrick (approved):
LGTM!

test/lwsp/Makefile
```diff
@@ -0,0 +1,65 @@
+# Python configuration file
+# REVCFG ?= ./rev-memh.py
+REVCFG ?= ../rev-model-options-config.py
+
+# Rev toolchain
+CC=riscv64-unknown-elf-g++
```
leekillough, line 6: This should be `CXX=${RVCXX}` once my [PR](https://github.com/tactcomplabs/rev/pull/291) is merged.

`CC` should only be used for C compilers; `CXX` for C++ compilers.

You may not need to create your own `Makefile` -- `test/CMakeLists.txt` has a general way of handling tests.

kpgriesser: thanks - I'll move to the general build method.

leekillough: It is already in my [PR](https://github.com/tactcomplabs/rev/pull/291)

PR #291, opened 2024-06-18
Pass RVCXX to tests; use ${RVCXX} instead of hard-coded compiler; fix copy-and-paste comments

leekillough:
- Pass `RVCXX` to Rev tests -- previously, only `RVCC` was being passed
- Use `${RVCXX}` instead of a hard-coded compiler to compile C++ tests
- Fix copy-and-paste comments about test names
- Right now the `unordered_map` and `unordered_map_obj` tests need to use the old hard-coded compiler <code>riscv64&#8209;unknown&#8209;elf&#8209;g++</code>  instead of `${RVCXX}` because if the correct C++ compiler is used, the `unordered_map` calls `floor()` which uses the `frflags` instruction which is pending on [#238](https://github.com/tactcomplabs/rev/pull/238) being finished.


leekillough:
@kpgriesser : We use `.cc` for C++ files, so I renamed `*.cpp` to `*.cc`. This way we only need to handle one suffix.

kpgriesser:
All the changes look fine. Can you update the README.md so users know how to set these?  The only mention in the document is  this example:  cmake -DRVCC=/path/to/riscv/compiler/exe.  It would be good to clarify that these are associated with gcc and g++.

PR #294, opened 2024-06-24
Add TEST_LEVEL variable to control which tests are run

### LINKED ISSUE #293: [FEATURE REQUEST] Different testing levels
A testing level should be added to control the level and timing of testing. Such as Low, Medium, High or simply Short and Long.

This is preferred to explicitly naming tests, such as `BUILD_ASM_TESTING` or `ENABLE_FP_TESTING`.

The default should enable all tests.

leekillough:
`-DTEST_LEVEL=3` runs all tests, including the FP tests on the `fenv` branch. It is the default level.

`-DTEST_LEVEL=2` runs all tests except the `fenv` tests. It runs the `ISA` tests, which are no longer controlled by `BUILD_ASM_TESTING`.

`-DTEST_LEVEL=1` runs a small number of tests, and takes less than a minute to run.

This fixes #293 .

The `all` label has been removed from tests.


rkabrick (approved):
Okay this looks good... Just for clarity's sake: If not otherwise specified, a test added using the `add_rev_test` cmake macro will always be run (ie. level_1)

leekillough:
> Okay this looks good... Just for clarity's sake: If not otherwise specified, a test added using the `add_rev_test` cmake macro will always be run (ie. level_1)

Yes, `test_level=1` is the default if it's not specified. There were only a few places where `test_level=2` needed to be added, and `test_level=3` is only the FP tests in my `fenv` branch (not shown here but will be merged with this when this is merged into `devel`)

PR #298, opened 2024-06-25
Fix stack and thread local storage allocation

leekillough:
The current code computes the stack pointer `sp` incorrectly. It does not guarantee it's on a 16-byte boundary, and it does not calculate the size correctly, using `Mem->GetThreadMemSegs().front()->getTopAddr()` instead of `Mem->GetStackTop()`.

The stack pointer `sp` grows downward. The thread local storage is accessed with nonnegative offsets from `tp`.

We compute the `StackTop - TLSSize` and round it down to a multiple of 16 bytes. `sp` and `tp` are both set to this address.

### DIFF
--- a/include/RevCPU.h
+++ b/include/RevCPU.h
@@ -235,9 +235,6 @@ private:
   // - Handles updating LSQueue & MarkLoadComplete function pointers
   void AssignThread( std::unique_ptr<RevThread>&& ThreadToAssign, unsigned ProcID );
 
-  // Sets up arguments for a thread with a given ID and feature set.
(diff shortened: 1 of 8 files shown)


rkabrick (approved):
Assuming this passes all the CI this looks good to me! 

I'm guessing this keeps the stack pointer aligned the whole time?

leekillough:
> I'm guessing this keeps the stack pointer aligned the whole time?

Assuming that all code adds/subtracts from `sp` in multiples of 16, and as long as we don't set `sp` to an unaligned address somewhere else, yes.

leekillough:
Fixes https://github.com/tactcomplabs/rev/issues/285

PR #300, opened 2024-06-26
Add support for extension versions

leekillough:
In [V2.0 of the RISC-V Specification](https://github.com/riscv/riscv-isa-manual/releases/tag/20240411), optional version strings after extension names were added (see **Chapter 39**).

The optional version strings are of the form: `nnn` [ `p` `nnn` ] with major and minor version numbers. Two versions of an extension ISA are considered compatible if their major version numbers are the same.

If the minor version is omitted, it is assumed to be `0`.

If the version string is omitted, it is assumed to be  the version of the extension when this version string feature was added.

If the `P` extension must follow another extension with a major but not minor version number, an underscore character must separate the version number and `P`.

This PR improves the `RevFeature` parser to support optional version strings in the architecture string, and to specify a range of versions which Rev supports.

Right now for supported extensions, the range is `[n,n]` where `n` is the default major version number. For unsupported extensions, the range is `[-1,0]`. There is a test case which validates that `V0p7` is not supported, so that if Vector support is ever added, Rev will display an error message if the `0.7` version of the Vector spec is requested (because once supported, the lower bound for `V` would be `1`).

Right now version numbers are only parsed and checked for support; they are not stored anywhere so that we could support multiple versions of an extension, but that could be added later.

### DIFF
--- a/include/RevFeature.h
+++ b/include/RevFeature.h
@@ -35,10 +35,10 @@ enum RevFeatureType : uint32_t {
   RV_P        = 1 << 9,   ///< RevFeatureType: P-Extension
   RV_V        = 1 << 10,  ///< RevFeatureType: V-extension
   RV_H        = 1 << 11,  ///< RevFeatureType: H-extension
-  RV_ZICSR    = 1 << 12,  ///< RevFEatureType: Zicsr-extension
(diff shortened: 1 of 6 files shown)


rkabrick (approved):
LGTM

jleidel (approved):
good to merge

PR #301, opened 2024-06-27
Zfa implementation

leekillough:
This implements the `Zfa` extension, adding  more floating-point instructions.

The `fcvt` instructions use the lower 2 bits of `rs2` to encode the type of conversion. The `Zfa` instructions use more than 2 bits of `rs2` for their FP instructions. The encoding used for table lookup is currently maxed out at 32 bits. It needs to be changed from `uint32_t` to `uint64_t` to be able to encode all 5 bits of `rs2`. Compressed instructions still only need `uint32_t` for their encoding, but they will automatically be upconverted to `uint64_t` when searched for in the tables.

The `rs2` field on floating-point instructions means more than simply the `fcvt` mode on `Zfa`, so the variable name in the encoding has been renamed from `fpcvtOp` to `rs2fcvtOp`.

The "optional compressed instructions" was a misnomer, and we do not need a separate table for them in `RevExt`. Those instructions were _RV32C-only_ instructions. The "O" means "Only RV32C", not "Optional". There is no need to tell in `RevCore::SeedInstTable()` whether a RV32C-only table needs to be added, and pass that flag to `RevCore::EnableExt()`. 

Instead, the `RV32I` constructor simply needs to add `c.jal` to its list of compressed instructions only on RV32 (not RV64), and the `RV32F` constructor only needs to call `SetCTable()` for its compressed instructions on RV32F (not RV64F or RV32D). Then there are only two lists of instructions, uncompressed and compressed.

I got the idea of conditionally adding instructions in the `RevExt` derived class constructors, because `Zfa` works for both `F` and `D`, and rather than having two extensions `ZfaF` and `ZfaD`, you only need one extension `Zfa`, and its constructor can **_optionally_** add `D` instructions to its table based on `Feature->HasD()` before calling `RevExt::SetTable()` or `RevExt::SetCTable()`.

`RevCore::EnableExt()` was considerably simplified, by not passing a `bool Opt` argument to enable or disable "optional" (RV32C-only) instructions, and by making the uncompressed and compressed instruction table builders use the same code in the form of a lambda.

In `RevExt`, `GetInstTable()` and `GetCInstTable()` were renamed to `GetTable()` and `GetCTable()`, to match the already-existing `SetTable()` and `SetCTable()`, respectively.

Braces were elided in the `include/insns/*.h` instruction tables, because they were unnecessary:
```c++
  std::vector<RevInstEntry> RV32ITable = {
    { RevInstDefaults().SetMnemonic("lui %rd, $imm"        ).SetFunct3(   0b000).SetImplFunc(lui   ).Setrs1Class(RevRegClass::RegUNKNOWN).Setrs2Class(RevRegClass::RegUNKNOWN).SetFormat(RVTypeU).SetOpcode(0b0110111) },
    { RevInstDefaults().SetMnemonic("auipc %rd, $imm"      ).SetFunct3(   0b000).SetImplFunc(auipc ).Setrs1Class(RevRegClass::RegUNKNOWN).Setrs2Class(RevRegClass::RegUNKNOWN).SetFormat(RVTypeU).SetOpcode(0b0010111) },
…
```
Becomes:
```c++
  std::vector<RevInstEntry> RV32ITable = {
    RevInstDefaults().SetMnemonic("lui %rd, $imm"        ).SetFunct3(   0b000).SetImplFunc(lui   ).Setrs1Class(RevRegClass::RegUNKNOWN).Setrs2Class(RevRegClass::RegUNKNOWN).SetFormat(RVTypeU).SetOpcode(0b0110111),
    RevInstDefaults().SetMnemonic("auipc %rd, $imm"      ).SetFunct3(   0b000).SetImplFunc(auipc ).Setrs1Class(RevRegClass::RegUNKNOWN).Setrs2Class(RevRegClass::RegUNKNOWN).SetFormat(RVTypeU).SetOpcode(0b0010111),
…
```

**The toolchains do not currently support the `Zfa` extension.**


include/RevExt.h
```diff
@@ -51,33 +51,25 @@ struct RevExt {
...
-
   /// RevExt: retrieve the extension name
   std::string_view GetName() const { return name; }
 
   /// RevExt: baseline execution function
   bool Execute( unsigned Inst, const RevInst& Payload, uint16_t HartID, RevRegFile* regFile );
 
   /// RevExt: retrieves the extension's instruction table
-  const std::vector<RevInstEntry>& GetInstTable() { return table; }
+  const std::vector<RevInstEntry>& GetTable() { return table; }
```
rkabrick, line 61: Any particular reason for the change? Not opposed to it just curious

rkabrick (approved):
LGTM

include/RevExt.h
leekillough: Like I said, it's to match the names `SetTable` and `SetCTable`. I don't know why `Inst` should be in the middle of the names for `Get`ting, while not being in the middle of the names for `Set`ting.

This PR is functionally ready, but I need to find a toolchain which supports `Zfa`, and I need to add a few more tests, before this PR can be merged. We need to see if there is a newer toolchain which can be installed on `pennywise`. @jleidel

jleidel: what version would you like installed?

leekillough: I don't know -- just some GCC/Clang which supports `Zfa`. I don't know what version when that feature was added.

leekillough: It's been ratified -- it's in the lastest spec -- it's fairly simple to implement -- so I added it.

jleidel: I installed riscv-gcc 13.2.0 on `pennywise`

kpgriesser (approved):
I'm thinking of opening a PR for the tracer to format floating point data as floating point vs the internal representation. What do you think?

leekillough:
> I'm thinking of opening a PR for the tracer to format floating point data as floating point vs the internal representation. What do you think?

Hexadecimal floating-point (`%a` `printf` format) can print the exact value with as few digits as possible. I use hexadecimal floating-point literals in my tests (see `test/fenv` and this PR), to make the values exact. With hexadecimal literals, there is no need to list values an arbitrarily long number of decimal digits to approximate a value accurately.

src/RevCore.cc
```diff
@@ -281,19 +256,19 @@ bool RevCore::InitTableMapping() {
...
-        "Core %" PRIu32 " ; Table Entry %" PRIu32 " = %s\n",
+        "Core %" PRIu32 " ; Table Entry %" PRIu64 " = %s\n",
         id,
         CompressEncoding( InstTable[i] ),
         ExtractMnemonic( InstTable[i] ).data()
       );
     } else {
       // map compressed instruction
-      CEncToEntry.insert( std::pair<uint32_t, unsigned>( CompressCEncoding( InstTable[i] ), i ) );
+      CEncToEntry.insert( std::pair<uint64_t, unsigned>( CompressCEncoding( InstTable[i] ), i ) );
```
donofrio, line 271: CompressCEncoding returns a uint32_t, here we are upcasting to a uint64_t, which is probably fine, but should we just use uint64_t everywhere to be consistent?

leekillough:
I have fixed the problem which was keeping me up for half a day -- apparently someone cleared bits by doing this:
```
 Enc &= 0xfffff8ff;
```
I had to increase `Enc` from 32 to 64 bits, and didn't notice that code until I stepped through it with a debugger.

I have changed it to this much more "defensive programming" for clearing bits:
```
Enc = ~( ~Enc | 0x700 );
```
It may look more complicated, but compilers can optimize it as good as the original.

That way we don't need to know the size of `Enc`, as long as `0x700` fits inside it.

I have tests for the "floating point load immediate" instructions (64 of them -- 32 for `F` and 32 for `D`), the round-to-FP-integer floating-point instructions with and without "Inexact" exceptions similar to `rint()` vs `nearbyint()`, the "modular" conversion of floating-point to 32-bit sign-extended integers (motivated by JavaScript), the FP min/max instructions which return NaN if any of their operands are NaN instead of ignoring the NaN operands for the others, and the quiet FP _ordered_ comparison instructions which don't generate invalid operation exceptions if their operands are NaN -- they simply return `false` if any operand is NaN.

The toolchain not accepting `Zfa` may still be an issue. I can temporarily comment out the tests.

src/RevCore.cc
leekillough: We use `uint64_t` whenever it is necessary, but we don't make any unnecessary changes, such as in the compressed instruction code. `uint32_t` is automatically upconverted to `uint64_t` when adding to or searching the tables. I don't think that storing tables in `uint64_t` and storing or searching for `uint32_t` keys is bad.

PR #302, opened 2024-07-09
Misc fixes

leekillough:
This fixes a few issues noticed during merge conflict resolution:
- Avoids displaying long error messages as test warnings when autodetecting `Zfa` and `RV32` support -- just display a short warning that certain tests are being disabled because `Zfa` or `RV32` support is unavailable
- Make unsupported RISC-V extension versions take less priority than an unparseable architecture string -- if there is an error in the order of extensions, or an unrecognized extension, print that error instead of printing that we do not support a particular version of an extension during the middle of parsing and then aborting early
- Add back `-DBUILD_ASM_TESTING=ON` which was inadvertently removed
- Use `std::unique_ptr` for `RevCPU` members which are allocated dynamically, so that `~RevCPU() = default` now (no more `delete`s). Since some of them interact with SST, and since SST requires "dumb" pointers in places, we cannot use `std::shared_ptr` to automatically handle ownership as they are passed around. Instead, `RevCPU` will own a `std::unique_ptr`, and then `.get()` will be used to pass a dumb pointer. It is assumed that none of functions passed dumb pointers will use them after `RevCPU` gets destroyed (the old code made the same assumption, only with dumb pointers)
- Use C++11 "raw strings" to make `splash_msg` source code look neater (no need for `\n` or `\` continuations)
- Make `numHarts` `uint32_t` so that it can exceed `65536` and we print the error message -- previously, we were testing a `uint16_t` against `65536` which obviously can't be greater than or equal to `65536`
- Make the tests which are skipped have the `DISABLED` property rather than not being added at all, so that we always have a full list of tests in a test run, just that some may be `DISABLED` and cannot contribute to the failure count
- Reformat `test/CMakeLists.txt` using Emacs autoindentation in `cmake-mode` (an add-on [extension](https://github.com/Kitware/CMake/tree/master/Auxiliary))
- Remove `include/insns/CMakeLists.txt`, as `InstructionSrcs` variable was never being used


include/RevCPU.h
kpgriesser, line 1: line 269: int address{ -1 }
Could -1 ever be misinterpreted as a valid address? Should there be an associated bool indicated it is valid?

kpgriesser (approved):

src/RevCPU.cc
```diff
@@ -52,12 +48,12 @@ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
 
   // Derive the simulation parameters
   // We must always derive the number of cores before initializing the options
-  numCores = params.find<unsigned>( "numCores", "1" );
-  numHarts = params.find<uint16_t>( "numHarts", "1" );
+  numCores = params.find<uint32_t>( "numCores", "1" );
```
donofrio, line 51: these should probably stay as uint16s - no need to have more than 65K cores per CPU or more than 65K harts per core

include/RevCPU.h
leekillough: I merely moved it from the constructor to a default initializer. I did not introduce this variable or know its semantics.

RevCPU.cc:
```
-RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params )
-  : SST::Component( id ), PrivTag( 0 ), address( -1 ), EnableMemH( false ), DisableCoprocClock( false ), Nic( nullptr ),
-    Ctrl( nullptr ), ClockHandler( nullptr ) {
+ RevCPU::RevCPU( SST::ComponentId_t id, const SST::Params& params ) : SST::Component( id ) {
```
 `RevCPU.h`:
```
-  int address{};  ///< RevCPU: local network address
+  int address{ -1 };  ///< RevCPU: local network address
```
The `RevCPU.cc` changes are not in this PR; they seem to have been in a previous PR without the corresponding changes for `address` in `RevCPU.h`, so this PR tries to remedy that.

The rationale is that the default initializers (initializers placed at the point of the member's declaration) are always there by default, so if someone forgets to initialize them in a constructor, they will still be initialized. Also, it avoids clutter in the constructor's initialization list when the initializer is a constant, so the constructor can focus on only initializing those members which depend on the constructor's parameters. All other members which don't depend on constructor parameters can have a default initialization (either with `= init` or `{ init }`), and in the case of constructors with several overloads, all of these constant-initialized members will be initialized in one place, their point of declaration, rather than in each constructor overload.

@tdysart There seems to be something fishy about the state of the `rev` and `revforza` repos, because I can perform merges and push them, but then when I clone the `devel` branch in a new local repo, it acts like the merge never happened. That may explain things like the `revforza` merge conflicts, which were trivial like 1-line conflicts (one of them was `int addresss{};` versus `int address { -1 };`), and mostly needed resolving as choosing the `revforza` side. There shouldn't be a conflict. Even after I resolved the conflicts and it indicated that I had no net code changes to `revforza/devel` in my merge (only the merge parent commit ids), I pushed it to try to mark `revforza/devel` as being up to date with `rev/devel`, but if I clone or checkout the `revforza/devel` branch again, it forgets the merge and reports merge conflicts again. It should have recorded the `rev/devel` commit id as a parent commit of the merge, which should make it up-to-date. I might need to try rebasing `revforza/devel` onto `rev/devel`, which will probably create a lot of conflicts. I can do it in a test branch, and if it works, we can rename `forzarev/devel` to something else and rename the test branch for `forzarev/devel`.

I'm used to merge and repo corruption problems with CVS, SVN, and Clearcase, but never Git.

src/RevCPU.cc
leekillough: > these should probably stay as uint16s - no need to have more than 65K cores per CPU or more than 65K harts per core

Then we cannot produce an error when the user requests `>= 65536`, because the test will always be false. A `uint16_t` can never be `>= 65536` so it will end up being some value `mod 65536` which we will accept and which the user didn't ask for. By using `uint32_t` we can at least catch and enforce when the user exceeds `65535`. The original code compared a `uint16_t >= 65536` which is always `false`. We allow `uint32_t` input when getting the parameter, but then we enforce that it isn't too large:
```c++
  // Make sure someone isn't trying to have more than 65536 harts per core
  if( numHarts > _MAX_HARTS_ ) {
    output.fatal( CALL_INFO, -1, "Error: number of harts must be <= %" PRIu32 "\n", uint32_t{ _MAX_HARTS_ } );
  }
```
Actually right now, the `_MAX_HARTS_` is defined as `4096`:
```c++
common/include/RevCommon.h:#define _MAX_HARTS_    4096
```
We should parse as `uint32_t` or `uint64_t`, then enforce the limit. If we parse as `uint16_t` we will always be under `65536` and won't catch anything larger as an error.

leekillough: Also, there is no space / performance advantage to expressing `numHarts` and `numCores` as `uint16_t`; it might even be slower than using the more "native" word sizes of `uint32_t` and `uint64_t`. We should use the more native word sizes `uint32_t` or `uint64_t`, and enforce the limits as we already do above, and then keep them at a native word size. If we use `uint16_t` to read the value, we will get a mod 65536 value which the user didn't want and we won't know what they asked for. Even if SST takes the `params.find` template type into account and raises overflow errors if the value is too large, that's less helpful of an error message than our limit check above.

leekillough: Also, if you literally mean "Make sure someone isn't trying to have **more than** 65536 harts per core", then to allow `65536` requires `uint32_t` because `uint16_t` cannot represent `65536`.

The declaration type of a variable does not imply intention about its permitted range. I can have a `uint64_t` variable and still enforce that its value is `<= 65536`. But if I have a `uint16_t` variable, I cannot tell whether the user asked for `> 65536` and raise an error for it.

The type and the allowable range of a parameter are separate but partially related concerns.

PR #306, opened 2024-07-18
Fully templatize floating-point helper functions

leekillough:
This fully templatizes the floating-point helper functions. Previously, the short ones were written explicitly in `include/insns/*.h`. Now all of them are instantiations of floating-point helper template functions, in preparation for `float16` and `bfloat16`  support. This also improves test coverage by making `float` and `double` instructions use the same code.

`UnBoxNaN` performs the opposite of the existing `BoxNaN`, so that the `GetFP()` function is much simpler. It is also fully templatized so that it can support `float16` and `bfloat16` in the future.

`CvtFpToInt` was renamed to `fcvtif` to resemble the naming scheme for instructions, where the destination type and register (integer) is listed first, and the source type and register (float) is listed second. The template parameters indicating the source and destination types are listed in the same order.

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -34,6 +34,8 @@
 
 namespace SST::RevCPU {
 
+using float16 = _Float16;
+
 /// Zero-extend value of bits size
 template<typename T>
 constexpr auto ZeroExt( T val, size_t bits ) {
--- a/include/RevInstHelpers.h
+++ b/include/RevInstHelpers.h
@@ -65,8 +65,8 @@ inline constexpr double fpmin<double, uint64_t> = 0x0p+0;
 /// General template for converting between Floating Point and Integer.
 /// FP values outside the range of the target integer type are clipped
 /// at the integer type's numerical limits, whether signed or unsigned.
-template<typename FP, typename INT>
-bool CvtFpToInt( RevFeature* F, RevRegFile* R, RevMem* M, const RevInst& Inst ) {
+template<typename INT, typename FP>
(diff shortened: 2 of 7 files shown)


rkabrick (approved):
LGTM

kpgriesser (approved):
no issues. It inspires me to do more advanced template programming

leekillough:
> no issues. It inspires me to do more advanced template programming

I once did template metaprogramming for _expression templates_, where a complicated expression like `alpha * A * B + beta * C` is automatically converted to a `DGEMM` call. I could send you my presentation slides on it, which is also an introduction to C++11 because it was around the time of C++11's ratification.

Advanced template programming techniques include SFINAE (_Substitution Failure Is Not An Error_), which is used to make templates conditional, like:
```c++
template<typename T, typename = std::enable_if_t<std::is_same_v<T, double>>>
T func(T x) {
}
```
The function definition will only be considered if `T == double`. Unlike a simple overload of `f` with `double`, this definition _does not participate_ in the overload if the `enable_if_t< ... >` condition is false.

There is also CRTP (_Curiously Recurring Template Pattern_), where a derived class is passed as a template argument to its base class, so that the base class can perform polymorphism statically at compile time instead of using virtual functions:
```c++
template<typename T>
class base {
  public:
     T func() {  // not virtual
        return static_cast<T*>(this)->func();    // casts this to derived * and called derived::func() statically
   }
};

class derived : base<derived> {
    derived func() {      // not virtual
      return derived{};
   }
};
```

Issue #312, opened 2024-08-23
Store-conditional Interface (RevMem::SCBase) does not handle Len and Target correctly


Artlands:
**Describe the bug**
The Len in RevMem::SCBase should only be either 4 or 8 in bytes, not 32 or 64 in bits. 
While writing zeros to the Target, the implementation does not make sense.  

**To Reproduce**
A simple test does not catch this bug. I found this issue when implementing remote lr/sc instructions in the xBGAS project.

**Additional context**
Fix: convert Len to TotalBits, i.e. Len*8, and use the TotalBits for comparing A with B.

Set Target to 0 could be like the following:
```     // write zeros to target
uint8_t* Tmp = reinterpret_cast<uint8_t*>( Target );
for( unsigned i = 0; i < Len; i++ ) {
   Tmp[i] = 0x0;
}
```

leekillough:
Words are measured in `Len` bytes, and must be `1`, `2`, `4` or `8` bytes.

4 bytes = 32 bits
8 bytes = 64 bits
So there is no difference whether we count in bytes or bits.

But this code does not look safe:
```c++
      // write zeros to target
      for( unsigned i = 0; i < Len; i++ ) {
        uint64_t* Tmp = reinterpret_cast<uint64_t*>( Target );
        Tmp[i]        = 0x0;
      }
```
`Tmp[i]` is indexing `uint64_t` values by the subscript from `0` to `Len-1`. It was probably intended to be more like this:
```c++
      // write zeros to target
      char* Tmp = reinterpret_cast<char*>( Target );
      for( unsigned i = 0; i < Len; i++ ) {
        Tmp[i]        = 0x0;
      }
```
But a `memset( Target, 0, Len );` is much simpler, and does not need `reinterpret_cast`.

Except when changing the signedness of an integral type, or casting to different aggregate types which have common initial members, or casting to a `char*` or `unsigned char*`, it violates [strict aliasing](https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing) to dereference a pointer after casting to a different type.

I'll clean up the code.

PR #313, opened 2024-08-25
Fix LR/SC and split A into Zaamo and Zalrsc

leekillough:
This fixes fundamental problems in the current Atomics extension.

- The `A` extension was split into `Zaamo` and `Zalrsc`, per the current standard. `A` implies both `Zaamo` and `Zalrsc`.
- The model of `LR/SC` is currently broken:
  * A "reservation" is not a list or dictionary of addresses -- _it is a single reservation of an address range by a specific hart_.
  * It is not a failure if a reservation already exists during the execution of an `LR` instruction, whether at a specific address or not. If an `LR` instruction is executed while a reservation already exists for the current hart, it replaces the hart's current reservation.
  * It is not a comparison of previous and new values which is important, but whether the reservation address range _was written to by other harts_ between the `LR` and `SC`.
  * The actual value to be stored by the `SC` instruction should not be compared with the current value of the target in memory, nor with the value of the target loaded at the time of the `LR`
    - The value cannot be loaded in deterministic time inside of the `SC` instruction because of MemH
    - Even if the current value is the same as the original, the reservation is still invalidated if the address range was written to by another hart
    - It is difficult to implement storing the value loaded during the `LR` separately from the destination register
    - The value of the register loaded in the `LR` instruction may have changed, without invalidating the reservation
- The current `SC` implementation [wrote 64 bits when the target was only 32 bits](https://github.com/tactcomplabs/rev/issues/312).
- The `aq` and `rl` (acquire and release) flags were not properly passed to the `LR` and `SC` instructions' flags

Style issues:

- `RV32A` and `RV64A` do not need to be separated -- `Zaamo` and `Zalrsc` can represent each Atomics subset, and the `RV32`/`RV64` differences can be handled at construction runtime, such as whether to include `RV64` instructions in the `Zaamo` and `Zalrsc` instruction tables.
-   The 32-bit `w` and 64-bit `d` variants of `LR` and `SC` can be templatized to use the same source code. Same for the `Zaamo` atomic instructions.
- `IsRV32` is redundant with `IsRV64`. There was an `IsRV32` data member of `RevRegFile`, and an `IsRV32()` function member of `RevFeature`. `IsRV64` is a positive feature, while `IsRV32` is a negative feature. I prefer to test only for the positive feature. In `if-else` or conditional expression situations, testing for `if(IsRV64)` is preferable to testing for `if(!IsRV32)`.
- The `aq` and `rl` (acquire and release) properties of atomic instructions should be represented as `bool`.
- With [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding), the LRSC "tuple index macros" such as `#define LRSC_HART` are unnecessary.

Notes from the spec:

> An SC may succeed only if no store from another hart to the reservation set can be observed to have occurred between the LR and the SC, and if there is no other SC between the LR and itself in program order.

> Regardless of success or failure, executing an SC.W instruction invalidates any reservation held by this hart.

> An implementation can register an arbitrarily large reservation set on each LR, provided the reservation set includes all bytes of the addressed data word or doubleword. An SC can only pair with the most recent LR in program order. An SC may succeed only if no store from another hart to the reservation set can be observed to have occurred between the LR and the SC, and if there is no other SC between the LR and itself in program order. An SC may succeed only if no write from a device other than a hart to the bytes accessed by the LR instruction can be observed to have occurred between the LR and SC. Note this LR might have had a different effective address and data size, but reserved the SC’s address as part of the reservation set.


jleidel (commented):
This is a solid set of changes that 1) fixes the attending issue and 2) brings Rev up to the point where we're supporting the *new* (newer) A extensions.  As an aside, this also makes it easy to replace A with the Forza-style atomics.  I know this is currently marked WIP, but I'm good to approve when you think its stable.

leekillough:
Fixes https://github.com/tactcomplabs/rev/issues/312

jleidel (approved):
good to merge

include/RevFeature.h
```diff
@@ -77,11 +78,8 @@ class RevFeature {
   /// GetMaxCost: get the maximum cost
   auto GetMaxCost() const { return MaxCost; }
 
-  /// IsRV32: Is the device an RV32
-  bool IsRV32() const { return xlen == 32; }
-
   /// IsRV64: Is the device an RV64
-  bool IsRV64() const { return xlen == 64; }
+  bool IsRV64() const { return xlen >= 64; }
```
rkabrick, line 82: What's the purpose of this?

include/RevMem.h
```diff
@@ -203,17 +203,11 @@ class RevMem {
     return ReadMem( Hart, Addr, sizeof( T ), Target, req, flags );
   }
 
-  ///  RevMem: template LOAD RESERVE memory interface
-  template<typename T>
-  bool LR( unsigned Hart, uint64_t Addr, T* Target, uint8_t aq, uint8_t rl, const MemReq& req, RevFlag flags ) {
-    return LRBase( Hart, Addr, sizeof( T ), Target, aq, rl, req, flags );
-  }
+  ///  RevMem: LOAD RESERVE memory interface
+  void LR( unsigned hart, uint64_t addr, size_t len, void* target, const MemReq& req, RevFlag flags );
```
rkabrick, line 207: const?

include/RevMem.h
```diff
@@ -203,17 +203,11 @@ class RevMem {
...
 
-  ///  RevMem: template STORE CONDITIONAL memory interface
-  template<typename T>
-  bool SC( unsigned Hart, uint64_t Addr, T* Data, T* Target, uint8_t aq, uint8_t rl, RevFlag flags ) {
-    return SCBase( Hart, Addr, sizeof( T ), Data, Target, aq, rl, flags );
-  }
+  ///  RevMem: STORE CONDITIONAL memory interface
+  bool SC( unsigned Hart, uint64_t addr, size_t len, void* data, RevFlag flags );
```
rkabrick, line 210: const?

include/RevMem.h
```diff
@@ -246,15 +240,12 @@ class RevMem {
...
-  bool LRBase( unsigned Hart, uint64_t Addr, size_t Len, void* Data, uint8_t aq, uint8_t rl, const MemReq& req, RevFlag flags );
-
-  /// RevMem: Clear a memory reservation for the target address
-  bool SCBase( unsigned Hart, uint64_t Addr, size_t Len, void* Data, void* Target, uint8_t aq, uint8_t rl, RevFlag flags );
-
   /// RevMem: Initiated an AMO request
   bool AMOMem( unsigned Hart, uint64_t Addr, size_t Len, void* Data, void* Target, const MemReq& req, RevFlag flags );
 
+  /// RevMem: Invalidate Matching LR reservations
+  bool InvalidateLRReservations( unsigned hart, uint64_t addr, size_t len );
```
rkabrick, line 247: const?

rkabrick (approved):
If you wouldn't mind answering the couple comments I left that'd be great but nothing looks problematic here!

include/RevFeature.h
leekillough: `XLEN=128` implies `RV64`. But I'll doubt we'll ever support `XLEN=128`.

include/RevMem.h
leekillough: `LR()` can't be `const` because it modifies `LRSC`, a `RevMem` member.
```c++
void RevMem::LR( unsigned hart, uint64_t addr, size_t len, void* target, const MemReq& req, RevFlag flags ) {
  // Create a reservation for this hart, overwriting one if it already exists
  // A reservation maps a hart to an (addr, len) range and is invalidated if any other hart writes to this range
  LRSC.insert_or_assign( hart, std::pair( addr, len ) );
```
The `MemReq` can be passed by `const` reference.

include/RevMem.h
leekillough: `SC()` cannot be `const` since it modifies `LRSC` too.

rkabrick: I meant the `Hart` or `Addr` or `Len`

include/RevMem.h
leekillough: `InvalidateLRReservations()` cannot be `const` because it modifies `LRSC`.

`LRSC` is an `std::unordered_map` which maps between Hart ID and an (Addr, Len) pair, where the reservation is the half-open interval `[Addr, Addr+Len)`. Any other harts which write to this address range cause the reservation to be invalidated (erased). `SC` fails if no reservation from `LR` is found, or if not all of the `SC`'s address range lies in the reservation.

rkabrick: Sorry should've specified I meant the arguments not the function

include/RevMem.h
rkabrick: Same as the others I was referring to the hart addr and len

leekillough: > Same as the others I was referring to the hart addr and len

What about the hart addr and len? It's an integer address and the len is the size of the load. Since `addr` and `len` are scalar integers passed by value, there is no need to mark them `const` since it just adds to the verbosity of the code without clarifying it.

include/RevMem.h
leekillough: > I meant the `Hart` or `Addr` or `Len`

Those are just scalar integers passed by value; marking them `const` would prevent the function body from modifying them, but these functions are so short that not much is gained by such verbosity. They are copies of integers passed by value; they are not references to data which should be qualified by `const`.

include/RevMem.h
leekillough: https://stackoverflow.com/questions/117293/does-using-const-on-function-parameters-have-any-effect-why-does-it-not-affect

rkabrick: Ahhh damn I thought I almost had ya :-)

PR #316, opened 2024-09-04
Use consistent naming of options; fix flake8 warnings

leekillough:
This change uses the same names for variables, command-line options, and SST parameters whenever possible. The main changes are:

- `enable_memH` → `enableMemH` for Rev SST parameter so that it matches `--enableMemH` command-line option

- `mem_size` → `memSize` and `MEM_SIZE` → `memSize` in Python source files so that they match Rev SST parameter `memSize`
  - but SST MemH's `mem_size` parameter name is out of our control, so we still use it

Also remove [flake8](https://flake8.pycqa.org/en/latest/) errors in all of the Python source files. `flake8` is a Python Lint. It flags things like unused imported modules, tab characters in Python files, indentation not at 4-byte multiples, incorrect spaces around colons, parentheses and commas, too long lines, extraneous semicolons at ends of lines; incorrectly formatted comments; etc.


scripts/asmCheck.py
```diff
@@ -7,36 +6,36 @@
 args = parser.parse_args()
 
 try:
-  sst_out = open(args.sstOutfile, 'r')
-except:
-  print("Cannot open file " + args.sstOutfile)
-  exit(1)
+    sst_out = open(args.sstOutfile, 'r')
+except Exception:
```
leekillough, line 10: `except:` catches _all_ exceptions, even `CTRL-C` on the keyboard. By specifying the `Exception` base class, we ensure that this is a true Python exception and not an external signal. `flake8` reports an error if bare `except` is used.

scripts/asmCheck_tracer.py
```diff
@@ -1,25 +1,28 @@
-import sys
 import argparse
 
-parser = argparse.ArgumentParser(description="Match SST output with asm file and/or print call stack at a specified clock tick - use with TRACER_ON in target program.  Use 'riscv64-unknown-elf-objdump -dC -Mno-aliases <exe>' to create assembly file")
+parser = argparse.ArgumentParser(description="""\
+Match SST output with asm file and/or print call stack at a specified clock
+tick - use with TRACER_ON in target program.
+
+Use 'riscv64-unknown-elf-objdump -dC -Mno-aliases <exe>' to create assembly file""")
```
leekillough: `"""` starts and ends a multi-line string, which can use `\` to suppress newlines but otherwise includes them. This avoids creating an ultra-long source line.

test/argc_bug/rev-test-basim.py
```diff
@@ -22,17 +22,17 @@
...
-        "numCores" : 1,                               # Number of cores
-        "clock" : "1.0GHz",                           # Clock
-        "memSize" : 1024*1024*1024,                   # Memory size in bytes
-        "machine" : "[0:RV64IMFDC]",                  # Core:Config
-        "startAddr" : "[0:0x00000000]",               # Starting address for core 0
-        "memCost" : "[0:1:10]",                       # Memory loads required 1-10 cycles
-        "program" : os.getenv("REV_EXE", "dram_tests.exe"),  # Target executable
-        "args"    : "64 4 2 3",                       # Target command line arguments
-        "trcStartCycle" : "1",                        # Trace all instructions
-        "splash" : 1                                  # Display the splash message
```
leekillough, line 35: `flake8` considers it bad style to put spaces before `:` colons. The `key: value` should have no spaces between the key and the colon, and one or more spaces between the colon and the value.

test/backingstore/rev-backingstore.py
```diff
@@ -15,86 +15,86 @@
...
+        "verbose": 5,                                         # Verbosity
+        "numCores": 1,                                        # Number of cores
+        "clock": "2.0GHz",                                    # Clock
+        "memSize": memSize,                                   # Memory size in bytes
+        "machine": "[0:RV64GC]",                              # Core:Config; RV64G for core 0
+        "startAddr": "[0:0x00000000]",                        # Starting address for core 0
+        "memCost": "[0:1:10]",                                # Memory loads required 1-10 cycles
+        "program": os.getenv("REV_EXE", "backingstore.exe"),  # Target executable
+        "enableMemH": 1,                                      # Enable memHierarchy support
+        "splash": 1                                           # Display the splash message
```
leekillough, line 32: I have added spaces so that the `#` comments on the right all line up on the same column, which must be two spaces away from the longest line.

test/pan_test2/rev-pan-test.py
```diff
@@ -33,104 +33,104 @@
...
+iface0 = nic0.setSubComponent("iface", "merlin.linkcontrol")  # -- host
+iface1 = nic1.setSubComponent("iface", "merlin.linkcontrol")  # -- pan
 
 # setup the router
 router = sst.Component("router", "merlin.hr_router")
 router.setSubComponent("topology", "merlin.singlerouter")
 
 # set the network params
-#-- host
+# -- host
```
leekillough, line 88: `flake8` expects at least one space between the `#` and the start of the comment's text.

scripts/asmCheck_tracer.py
```diff
@@ -31,39 +34,39 @@
...
-      if(not args.callStackClk):
-        print(o)
+    if "Render:" in line and "Core" in line:
+        print(line)
+        clk = line.split("Render:")[1].split(']:')[0].strip()
+        PC = line.split("*I ")[1].split(':')[0].split('0x')[1]
+        so = line.split("*I ")[1].split('\t')[0].strip()
+        if PC in asmHash:
+            o = '{:<60} {:<3} {:<60}'.format(asmHash[PC].strip(), ":::", so)
+        if (not args.callStackClk):
```
donofrio: Indentation on this `if` changed, appears to be in a different scope now?

leekillough: > Indentation on this `if` changed, appears to be in a different scope now?

Good catch; I'll check to make sure the nesting is consistent.

leekillough:
I ran it through the Python `reindent` tool to avoid any surprises. `reindent` keeps the original syntax but reindents the code so that it is in 4-space indentation levels. After the reindentation, I eliminated all `flake8` errors. I renamed the variables as per my original comment.

jleidel (approved):
I'm good with all these changes.  Is it worth putting together an updated README file and/or a separate document that describes the various potential Python parameters and their values?  There are so many at this point that it might be valuable for the future stability of the simulator

leekillough:
> I'm good with all these changes. Is it worth putting together an updated README file and/or a separate document that describes the various potential Python parameters and their values? There are so many at this point that it might be valuable for the future stability of the simulator

I can add a `flake8` test to the `pre-commit` script if `flake8` command can be found, and it will reject commits which don't follow the `flake8` rules and point out exactly what needs to be changed.

PR #317, opened 2024-09-05
Make feature pointers pointers to const RevFeature

leekillough:
This makes the `RevFeature* Feature` pointers into `const RevFeature* Feature` since the feature data is read-only, and we don't want functions to modify `RevFeature`.

(The `const` does not apply to the function-local variable pointer `Feature`, but to _what it points to_.)


donofrio (approved):
looks good to me

PR #318, opened 2024-09-05
change spacing to be more consistent

leekillough:
This changes the spacing in some tables and declarations to look better.

Also, some `std::map` were changed to `std::unordered_map`. Unless the keys in a map must be kept in sorted order (such as to iterate across all keys in the map in sorted order), then `std::unordered_map` is generally faster.


kpgriesser (approved):
I don't foresee any problems with the map ordered/unordered changes

PR #320, opened 2024-09-06
Use enum instead of macros; fix bug where cbo.clean and cbo.flush were switched

leekillough:
Use `enum class` instead of macros in `Zicbom.h` for consistent type-safe naming.

Change `enum RevImmFunc` to `enum class RevImmFunc` to reduce the chances of naming conflicts.  Change all of its usages to have `RevImmFunc::` prefix (that is the bulk of the source changes in terms of lines of code).

Fix bug where `cbo.clean` and `cbo.flush` actions were reversed because their macro definitions were backwards.

Fix decoder to use `imm12` field when looking up CBO instructions -- previously it only matched `cbo.inval` because it did not consider the `imm12` field as part of the encoding (similar to `ecall` and `ebreak`).

Unfortunately, even though we can indicate `RevImmFunc::FEnc` in the instruction table, we cannot simply look this up during decoding, because decoding must look up the tables by encoding the `imm12` field in the key for some instructions before it can read the `RevImmFunc::FEnc` setting.

Add predicate that `Inst.rd == 0` for CBO instructions, and reserve `Inst.rd != 0` for other instructions.

Pass the CBO instruction inplementation function a non-type template argument, so that each CBO instruction can be its own separate template instantiation without runtime tests for the `imm12` value. The compiler will optimize out the `switch` statement to the only CBO instruction which applies.

Prepare for `cbo.zero` to be added in the future.


kpgriesser (approved):
LGTM

PR #321, opened 2024-09-10
Fast print utility

kpgriesser:
This custom ECALL is intended to help debug code by providing non-intrusive printf functionality to help speed up rev application and rev test code development. It is currently  limited to 1K output chars and takes a format string and up to 6 numeric parameters.

Advantages:
- allows formatting for numeric values
- thread safe with respect to other harts.
- single cycle execution
- does not clutter execution trace when someone just wants to print a few values.
- xml tagging allows for post-processing script to extract the printed strings.

Disadvantages:
- restricting usage to using only 1 numeric parameters and a 1K char buffer.
- cannot pass in strings as parameters.

Example:

```
#define printf rev_fast_printf
// ... bunch of code
      printf("Failed: check_data[%d]=0x%lx dram_src[%d]=0x%lx\n",
              i, check_data[i], i, dram_src[i]);
```

Output:
```
RevCPU[cpu0:operator():10080000]: <rev-print>Failed: check_data[0]=0xffffffffaced0000 dram_src[0]=0xffffffffaced0000
</rev-print>
```

### DIFF
--- a/common/syscalls/syscalls.h
+++ b/common/syscalls/syscalls.h
@@ -673,7 +673,7 @@ REV_SYSCALL( 9006, void dump_thread_mem( ) );
 REV_SYSCALL( 9007, void dump_thread_mem_to_file( const char* outputFile ) );
 
 // ==================== REV PRINT UTILITIES
-REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a4=0, uint64_t a6=0) );
+REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0) );
 
 // clang-format on
 
--- a/common/syscalls/syscalls.h
+++ b/common/syscalls/syscalls.h
@@ -671,6 +671,10 @@ REV_SYSCALL( 9004, void dump_valid_mem( ) );
 REV_SYSCALL( 9005, void dump_valid_mem_to_file( const char* outputFile ) );
 REV_SYSCALL( 9006, void dump_thread_mem( ) );
 REV_SYSCALL( 9007, void dump_thread_mem_to_file( const char* outputFile ) );
+
+// ==================== REV PRINT UTILITIES
+REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a4=0, uint64_t a6=0) );
+
 // clang-format on
 
 #endif  //SYSCALL_TYPES_ONLY
--- a/common/syscalls/syscalls.h
+++ b/common/syscalls/syscalls.h
@@ -673,7 +673,9 @@ REV_SYSCALL( 9006, void dump_thread_mem( ) );
 REV_SYSCALL( 9007, void dump_thread_mem_to_file( const char* outputFile ) );
 
 // ==================== REV PRINT UTILITIES
+#ifdef __cplusplus
 REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0) );
+#endif
 
 // clang-format on
 
--- a/include/RevCore.h
+++ b/include/RevCore.h
@@ -306,8 +306,8 @@ private:
   std::function<uint32_t()> const GetNewThreadID;
 
   // If a given assigned thread experiences a change of state, it sets the corresponding bit
-  std::vector<std::unique_ptr<RevThread>>
-    ThreadsThatChangedState{};  ///< RevCore: used to signal to RevCPU that the thread assigned to HART has changed state
+  std::vector<std::unique_ptr<RevThread>> ThreadsThatChangedState{
+  };  ///< RevCore: used to signal to RevCPU that the thread assigned to HART has changed state
 
   SST::Output* const                output;                       ///< RevCore: output handler
   std::unique_ptr<RevFeature>       CreateFeature();              ///< RevCore: Create a RevFeature object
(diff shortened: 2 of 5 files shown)


jleidel (approved):
would it be useful to store some sort of format enum as an integer in `a1`.  Something that corresponds to an enum table.  EG:

```
switch( RegFile->GetX<uint64_t>( RevReg::a0 ) ){
case 0 : str = "ERROR"; break;
case 1 : str = "NOTICE"; break;
default: ... 
}
```

otherwise, looks good as is.  I don't think this will collide with the Forza encodings for ECalls

leekillough:
Please wait for my review. I'm chasing a storm.

kpgriesser:
To John's point, I could add variants which support standard 'NOTICE, WARNING, FATAL' prefixes which would benefit. Since I'm using sst::Output we could use FATAL to exit the simulation ( an improvement over just crashing on an assert ). For these, I would suggest another PR that builds on this one.

common/syscalls/syscalls.h
```diff
@@ -671,6 +671,10 @@ REV_SYSCALL( 9004, void dump_valid_mem( ) );
 REV_SYSCALL( 9005, void dump_valid_mem_to_file( const char* outputFile ) );
 REV_SYSCALL( 9006, void dump_thread_mem( ) );
 REV_SYSCALL( 9007, void dump_thread_mem_to_file( const char* outputFile ) );
+
+// ==================== REV PRINT UTILITIES
+REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0) );
+
```
leekillough, line 679: Have you considered variadic templates (`uint64_t ...a`)?

They would allow an arbitrary number of arguments without having to use default values or placeholders.

include/RevCore.h
```diff
@@ -671,7 +671,10 @@ class RevCore {
 
   EcallStatus ECALL_dump_thread_mem();         // 9006, dump_thread_mem()
   EcallStatus ECALL_dump_thread_mem_to_file(); // 9007, dump_thread_mem_to_file(const char* outputFile)
-  // clang-format on
+
+  // =============== REV print utilities
+  EcallStatus ECALL_fast_printf();             // 9010, rev_fast_printf(const char *, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0)
+    // clang-format on
```
leekillough, line 677: Variadic templates would make it simple to pass an arbitrary number of arguments. `std::forward` is used to pass the rvalue-ness of the arguments.

leekillough:
`asprintf()` can be used to avoid the 1024-character limit. It automatically allocates memory, which should be passed to `free()`. It returns `-1` if there is an error.

include/RevCore.h
kpgriesser: I originally was using a variadic but they did not work on that version of REV ( much earlier this year ).  I will try it again on the latest version of REV. I would be good to confirm that the variadics now work.

leekillough: Well, since you're accessing each `ECALL` argument (`a0-a7`) separately, the variadics may be only useful at the C++ level but not lower-level.

You may also have to be concerned with `XLEN` being 32- or 64-bits, although 64-bits (`uint64_t`) should be safe at the C++ interface level.

leekillough: By the way, even if `XLEN==32`, if you use `GetX<uint64_t>(reg)`, the 32-bit register will be zero-extended to 64 bits. Similarly. `GetX<int64_t>(reg)` will sign-extend the 32-bit register to 64 bits.

Using size-specific types like `uint64_t` is strongly encouraged instead of using types like `unsigned` or `long`.

leekillough:
Go ahead if it passes all tests

include/RevCore.h
leekillough: > I originally was using a variadic but they did not work on that version of REV ( much earlier this year ). I will try it again on the latest version of REV. I would be good to confirm that the variadics now work.

It should not depend on the Rev version. Variadic templates are a C++11 feature and we've been compiling Rev with C++17 for over a year. If you use non-type template arguments like `<uint64_t... args>` then a variable number of integer arguments should be passed to the function.

kpgriesser:
Keeping this open after finding a fly in the ointment. When compiling with gcc instead of g++ the default function parameters will not compile.  For the time being, the function will be omitted from syscalls.h unless c++ is detected.
```
+#ifdef __cplusplus
 REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0) );
+#endif
```

leekillough:
> Keeping this open after finding a fly in the ointment. When compiling with gcc instead of g++ the default function parameters will not compile. For the time being, the function will be omitted from syscalls.h unless c++ is detected.
> 
> ```
> +#ifdef __cplusplus
>  REV_SYSCALL( 9110, int rev_fast_printf(const char* format, uint64_t a1=0, uint64_t a2=0, uint64_t a3=0, uint64_t a4=0, uint64_t a5=0, uint64_t a6=0) );
> +#endif
> ```

Up to now our `syscalls.h` API has been C-compatible. Default arguments is a C++ feature.

You can use `<stdarg.h>` varargs to make it C-compatible, but you'll need to use a sentinel value to indicate the end of list, or make the count depend on an earlier parameter.

leekillough:
> You can use `<stdarg.h>` varargs to make it C-compatible, but you'll need to use a sentinel value to indicate the end of list, or make the count depend on an earlier parameter.

The `vfprintf()` functions can be used to pass `varargs` to your version of `printf`, and the argument list length will be automatically deduced from the format string. You need to declare your function like this:
```
#include <stdarg.h>

int rev_fast_printf( const char* format, ... ) {
   va_list args;
   va_start( args, format );
   vfprintf( stdout, format, args );
   va_end( args );
   return 0;
}
```

This is C-compatible and will accept any number of arguments of any types. You can also use the GCC `__attribute__((format))` on the function to enforce matching of format to arguments (see [this](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) ).

You might need to use `vsnprintf()` instead of `vprintf()` if you need to buffer to a string instead of producing direct IO.

leekillough:
I am going to modify your changes to allow a general `printf` format string and work with C.

leekillough:
Please see this commit:

https://github.com/leekillough/rev/commit/12818964da2110efaa8ef84e6601043f09f9f9f0

It's sort of a hack, but it:

1. Uses `varargs` C syntax to pass arguments, exactly how `vprintf` and related functions work. There is no need to test for C++.
2. Uses `__attribute__((format ...))` to indicate that `rev_fast_printf()` is a `printf()`-style function, and produces warnings if the arguments don't match the format string.
3. Grabs 32- or 64-bit RISC-V `ECALL` arguments based on `XLEN`. Right now it really can only work for `XLEN`-sized integers. I have not tested it with `XLEN==32` because my toolchain doesn't support RV32.
4. Calls `asprintf()` on the host to automatically allocate memory for the string result, and calls `free()` if there are no errors.

leekillough:
On that last `asprintf` point: I realize it might not exist on MacOS. I'm willing to go back to the `1024` or whatever `snprintf` buffer limit.

The rest of my changes are much more important.

PR #322, opened 2024-09-17
Fast printf

leekillough:
This supersedes https://github.com/tactcomplabs/rev/pull/321


kpgriesser:
Tested on MacOS for gcc and g++ on Lee's branch. will verify one more time on this branch. Should be good.

PR #323, opened 2024-09-17
tests and post-processing script for rev-fast-print sys call

kpgriesser:
Update  rev/test/tracer to include various calls to rev-fast-print and also check for gcc and g++ compilation. When the tests complete a new post-processing script, scripts/rev-print.py, is run on each log file to extract the messages printed from user code run on REV.

The tests are not currently added to tests/CMakeLists.txt so this PR should not affect any users ( but these will be added in a future PR).


test/tracer/rev-test-tracer.py
```diff
@@ -33,22 +32,22 @@
...
+        "memSize" : 1024*1024*1024,                   # Memory size in bytes
+        "machine" : f"[CORES:{arch}]",                  # Core:Config; RV32I for all
+        "startAddr" : "[CORES:0x00000000]",           # Starting address for core 0
+        "memCost" : "[0:1:10]",                       # Memory loads required 1-10 cycles
+        "program" : rev_exe,                          # Target executable
+        "splash" : 1,                                 # Display the splash message
+        # "trcOp": "slli",                            # base command for tracing [default: slli]
+        # "trcLimit": 0,                              # Maximum number of trace lines [default: 0]
+        # "trcStartCycle" : 0                         # Starting trace cycle [default: 0]
+        })
```
leekillough: Use the `flake8` tool to verify all `*.py` files are good. Type:
```
flake8  --max-line-length 131 file.py
```

`brew install flake8` or `pip install flake8` installs it.

kpgriesser: will do

kpgriesser: @leekillough  There is some memory corruption that appears to be occurring with the new rev-fast-print.  I put a bug test and script in the tracer directory.

`cd test/tracer
./bug.sh
`
You should see this error
`FATAL: RevCPU[cpu0:CalcPhysAddr:6987000]: Segmentation Fault: Virtual address 0xf0c (PhysAddr = 0xffffffffffffffff) was not found in any mem segments`

The problem goes away if:

1. compile with -O0 ( change in bash.sh )
OR
2. remove the printf's.  ( in bug.cpp: uncomment #define printf(...) )

I'll see if I can narrow it down further by comparing the traces with the rev-fast-print branch. Maybe you'll see something obvious -

leekillough: It appears that the `naked` attribute cannot be used with `varargs` function parameter lists. The compiler emits code which corrupts the stack.
```
0000000000010210 <rev_fast_printf>:
   10210:       00b13423                sd      a1,8(sp)
   10214:       00c13823                sd      a2,16(sp)
   10218:       00d13c23                sd      a3,24(sp)
   1021c:       02e13023                sd      a4,32(sp)
   10220:       02f13423                sd      a5,40(sp)
   10224:       03013823                sd      a6,48(sp)
   10228:       03113c23                sd      a7,56(sp)
   1022c:       000028b7                lui     a7,0x2
   10230:       3968889b                addw    a7,a7,918 # 2396 <exit-0xdd52>
   10234:       00000073                ecall
   10238:       00008067                ret
```
It stores `a1-a7` on the stack, corrupting its caller.

I'm trying to find a workaround.

kpgriesser:
closing this PR. I had polluted the git history and was creating a clean branch and unintentionally pushed into devel. All the formatting changes are in and all the tests pass.  To see the results of the script go to 'test/tracer', invoke make, and look at the *.revlog files.

leekillough:
Tip: you can use `git reset tcl/devel` or `git reset --soft tcl/devel` to keep your changes and push them as a new commit once merges you need are merged in. Without `--soft` is makes all of your changes relative to 'tcl/devel` as new local files you must `git add`, interactively with `git add -p`. With `--soft`, it prepares the index with all of your changes to be combined into a new single commit. `git merge` can automatically merge changes while recognizing changes you already merged because of identical lines. If you have to save your changes to switch to another branch and work on something, you can use `git stash push` with optional `-k` to keep your index, and then `git stash pop` or `git stash apply` to later get your changes back.

PR #324, opened 2024-09-18
Fix rev_fast_printf

leekillough:
This fixes the problem mentioned in https://github.com/tactcomplabs/rev/pull/323 but removes the `format` attribute which does `printf` argument checking. The C and C++ solutions have to be slightly different.

### DIFF
--- a/common/syscalls/syscalls.h
+++ b/common/syscalls/syscalls.h
@@ -18,6 +18,10 @@
 #include <sys/types.h>
 #include <time.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 // The following is required to build on MacOS
 // because sigval & siginfo_t are already defined
 // in the two headers that get pulled in when building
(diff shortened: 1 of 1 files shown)


common/syscalls/syscalls.h
```diff
@@ -672,10 +676,32 @@ REV_SYSCALL( 9005, void dump_valid_mem_to_file( const char* outputFile ) );
...
+}  // extern "C"
+
+template<typename... Ts>
+__attribute__( ( naked ) ) static int rev_fast_printf( const char* format, Ts... args ) {
+  asm( " li a7, 9110; ecall; ret" );
+}
+
+#else  // __cplusplus
+
+__attribute__( ( naked ) ) static int rev_fast_printf() {
```
kpgriesser, line 694: Interesting,  how does the compiled code handle the variable number of arguments here? It's definitely working correctly in my tests for both C and C++

kpgriesser (approved):
Tested with gcc, g++, rv64g, and rv32i. Looks ready for prime-time.

common/syscalls/syscalls.h
leekillough: That is old K&R syntax, where a blank parameter list on a function declaration had an unspecified number of parameters. You could legally declare a function with something like `int func();` and then in a `.c` file actually implement it with specified parameters (either with an ANSI C prototype function definition or the old K&R syntax).

Since we're using `naked` mode and passing the arguments as-is to `ECALL`, we don't have to declare parameters or pass them. `ECALL` receives the same arguments as `rev_fast_printf()`.

Previously I had declared `const char* format` and `...` as the parameters and expected them to be passed to `ECALL` as-is, because on RISC-V even a `varargs` function parameter will fill register slots `a1-a7` (at least one argument has to come before a `varargs` parameter,  which is almost always going to `a0`).

The GCC compiler apparently spills `a1-a7` as caller-saved registers when a `varargs` parameter is specified, even when the `naked` attribute is used. It corrupts the `rev_fast_printf()` caller's stack, because it doesn't adjust the stack before spilling the caller-saved registers. Apparently that is a bug in the `naked` attribute when `varargs` are used, since it should not be spilling registers.

C23 obsoletes the old K&R syntax, and in C23 and later, `int func();` is the same as `int func(void);`.

C++ does not support K&R syntax of unspecified arguments, and has always treated `int func();` as a function taking no parameters, so I had to use a variadic template for its argument list. Templates in C++ cannot be mixed with `extern "C"`, so the function name must be mangled in C++ if it uses templates.

The `format` attribute cannot be used with variadic templates until GCC 7.x, and it cannot be used on K&R style functions with unspecified parameters. You must tell it which slots the format and the `printf` arguments begin at, and they have to be declared in the function definition.

PR #325, opened 2024-09-18
Print warnings about C++ filename extensions

leekillough:
This enforces the `.cc` and `.h` filename extensions, printing an error if `.cxx`, `.cpp`, `.C` or `.hpp` are used.


kpgriesser (approved):
Verified in fast-print-test branch.

PR #327, opened 2024-09-20
Overhaul of CSR functionality

leekillough:
This makes changes to the CSR implementation to make it more modular and more easily expanded for F***a:
- It separates out CSR functionality from `RevRegFile`, moving it to a `RevCSR` base class of `RevRegFile`.
- Instead of `RevCSR` being a _scoped_ `enum class` with names of CSR registers, it is now a full `class` with an _unscoped_ `enum` listing the names of all CSR registers. The syntax to access a CSR register is still `RevCSR::name`.
- Instruction decoding macros have been replaced with `constexpr` functions for better safety. The return type of a decoding is the smallest unsigned integer type capable of representing all values of the decoded field.
- In `ZeroExt()` if the type was smaller than `int`, it would get promoted to `int` and that could cause undefined behavior because left-shifting a negative signed value is undefined. I changed the syntax slightly so that it only left-shifts unsigned values.
- The `RevCore::output` member was made `public` so that error messages could be outputted just knowing the owning `RevCore`.
- Floating-point CSR type declarations were moved out to `RevFCSR.h`.
- A bug in `GetFRM()` was fixed -- it only returned the bottom 2 bits of the dynamic rounding mode, instead of the 3 bits it contains.
- Some header file dependencies were fixed -- previously some files depended on one header file they `#include`d to `#include` another header file they depended on. Now every header file that a file needs is explicitly `#include`d by it instead of depending on another header file to `#include` it.
- `Zicntr` (`rdcycle`, `rdtime`, `rdinstret`, etc.) was added as another extension (`Zicntr` implies `Zicsr`).
- `Zicntr` was moved into its own header file `RevZicntr.h`
- `RevCSR` was structured in a way so that handling additional CSR registers is made easy with base classes of `RevCSR`. `RevZicntr` is the first example.
- To access functions defined in derived classes, such as accessing `GetPC()` in `RevRegFile` from `RevZicntr`, the functions are declared pure virtual (`= 0`) in the base classes, not as an abstract interface declaration, but to indicate that some derived class must implement it for the base class. Pure virtual function calls do not use dynamic dispatch, so there is less of a performance hit than with regular virtual function calls.
- `RevRegFile::IsRV64` and `RevRegFile::HasD` were changed from `const` data members to member functions, so that they could be accessed from base classes which declare them pure virtual. Since they are just `RevFeature` queries, `R->IsRV64()` was changed to `F->IsRV64()` wherever possible, even though making them pure virtual in base classes should not require dynamic dispatch.
- Similarly, `RevRegFile::GetCore()` was added, so that `RevCSR` and its base classes could access the `RevCore` owning the hart by declaring `GetCore()` as a pure virtual function.
- For the floating-point environment handled in `RevFenv.h`, instead of storing a reference to the `FCSR` floating-point status and modifying it at `RevFenv` destruction time, only a pointer to the `RevRegFile` is stored, and a new function `RevCSR::SetFFlags()` is used to set floating-point exception flags.
- If `Zicntr` is not enabled, then accessing the `TIME`, `CYCLE`, and `INSTRET` CSR registers will print a fatal error message about `Zicntr` not being enabled.
- If `RV64` is in effect, the `RDCYCLEH`, `RDTIMEH` and `RDINSTRETH` instructions will produce a fatal error message about them being illegal instructions on `RV64`.
- `SetCSR()` was changed to return `bool`, currently always `true`. If `false`, it will make CSR instructions which modify CSR registers produce a fatal non-specific error message.
- A `zicntr` label was added to the tests, so that `RDCYCLE`, `RDTIME` and `RDINSTRET` tests pass the `Zicntr` architecture string to Rev. Currently `Zicntr` is not supported by my toolchain, but it is enabled by default with `Zicsr`. If we move up in toolchains to where `Zicntr` is not enabled by default, the test `Makefile`s have commented out `-march=` flags which will enable `Zicntr`.
- A detailed comment about the design and modularity of Rev's CSR code was added, to prevent confusion:
```
////////////////////////////////////////////////////////////////////////////////
//
// A note about the separation of scopes in include/insns/Zicsr.h and
// include/RevCSR.h:
//
// RevCSR.h: GetCSR() and SetCSR() are used to get and set specific CSR
// registers in the register file, regardless of how we arrive here. If a
// particular CSR register is disabled because of CPU extensions present, or if
// a particular CSR register does not apply to it (such as RDTIMEH on RV64),
// then raise an invalid instruction or other exception here.
//
// Zicsr.h: Decode and execute one of only 6 CSR instructions (csrrw, csrrs,
// csrrc, csrrwi, csrrsi, csrrci). Do not enable or disable certain CSR
// registers, or implement the semantics of particular CSR registers here.
// All CSR instructions with a valid encoding are valid as far as Zicsr.h is
// concerned. The particular CSR register accessed in a CSR instruction is
// secondary to the scope of Zicsr.h. Certain pseudoinstructions like RDTIME or
// FRFLAGS are listed separately in Zicsr.h only for user-friendly disassembly,
// not for enabling, disabling or implementing them.
//
// To ease maintainability and prevent large code size, it is recommended that
// functions related to specific CSR registers be made base classes of RevCSR
// in separate header files (e.g. RevZicntr). RevCSR can then dispatch the
// GetCSR()/SetCSR() functions of these CSR registers to the base class.
//
// To access a RevCSR or RevRegFile member function in one of its base classes,
// it is recommended that the function be made a pure virtual function in the
// base class so that RevCSR or RevRegFile must override it, similar to how
// GetCore() is a pure virtual function in RevZicntr which RevRegFile
// overrides. RevZicntr needs GetCore(), but rather than have to store a
// pointer in RevZicntr's constructor, it is much simpler to simply declare
// GetCore() as a pure virtual function in RevZicntr which must be overriden,
// which makes RevZicntr and RevCSR abstract classes which cannot be
// instantiated except as a base class of RevRegFile, which defines GetCore().
////////////////////////////////////////////////////////////////////////////////
```

Cc: @tdysart


kpgriesser:
I test using rv32g with and without '_zicntr' and confirmed that without the zicntr extension I received a very helpful message that will save people a lot of time scratching their heads ( thanks)

FATAL: RevCPU[cpu0:fatal:34790000]: Illegal instruction at PC = 0x11e44: Zicntr extension not available

I updated my tests in tests/tracer to use the zicntr extension and everything is passing. I'll push these changes after the PR is merged

include/RevInstTable.h
```diff
@@ -11,6 +11,8 @@
 #ifndef _SST_REVCPU_REVINSTTABLE_H_
 #define _SST_REVCPU_REVINSTTABLE_H_
 
+#include "../common/include/RevCommon.h"
```
kpgriesser: Just curious, does this have to be a relative path? It's like this for most of the files except for include/RevTracer.h and src/RevCommon.h

kpgriesser (approved):
This looks great. Will really make managing and debugging CSR's much easier

include/RevInstTable.h
leekillough: I just removed the `../common/include/` from 6 files. Apparently there is a `-I` include path option added somewhere in the build.

tdysart: The include directive is here: https://github.com/tactcomplabs/rev/blob/eaf1fe467f73b0d6df61a8137557a7c9eb3ca806/CMakeLists.txt#L131

PR #328, opened 2024-09-20
Cleanup shellcheck errors

leekillough:
This fixes warnings from the [shellcheck](https://www.shellcheck.net/) linter.

All default and optional warnings except a few have been enabled.

In all of the CI Slurm scripts, multiple redirections in multiple commands were replaced with a single `exec` redirection, which redirects the current shell and its children.

`.githooks/pre-commit` warns if you don't have `shellcheck` installed -- it can be installed with `brew install shellcheck` or `apt-get install shellcheck`.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -181,8 +181,6 @@ reformat_files() {
     get_files
 
     errors=false
-    flake8_warning=false
-    shellcheck_warning=false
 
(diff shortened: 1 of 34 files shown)


kpgriesser (approved):
confirmed that this detects issues in my scripts and also confirmed that there are no spurious messages on empty commit ( all on MacOS ).

Issue #331, opened 2024-10-04
SST Segmentation Fault When Accessing Memory Across Pages with memHierarchy Backend


Artlands:
**Describe the bug**
A segmentation fault occurs in SST when reading memory across pages using the memHierarchy backend due to incorrect logic in _RevMem::ReadMem_.

**To Reproduce**
The issue is observed in the devel branch when the page size is changed to a smaller value, such as 4KB, during the execution of the _test/syscalls/munmap_ test.

**Additional context**
In the _RevMem::ReadMem_ function, the code responsible for handling flag responses and clearing hazards (lines 770-783 in _RevMem.cc_) should be placed inside the else block when using the internal RevMem model.

**Other comments/questions**
When using the memHierarchy backend, the cross-paging access is not relevant not only for memory read (as commented in line 771) but also for memory write. There is redundancy in calling `ctrl->sendWRITERequest` in _RevMem::WriteMem_, and the logic in these functions can be simplified. Additionally, the `BaseMem` address passed to `ctrl->sendWRITERequest` and `ctrl->sendREADRequest` is never used; only the virtual address is needed when sending memory requests to memHierarchy. Therefore, the memory controller interface could also be simplified.

leekillough:
Can you please attach a test code?

Artlands:
> Can you please attach a test code?

The test code is `/test/syscalls/munmap/munmap.c`

After changing the pageSize to 4096 in RevMem.cc and reinstalling REV, run `sst rev-model-options-config.py --model-options="--program=./syscalls/munmap/munmap.exe --enableMemH=1"` in the test folder.

leekillough:
If I run as you say, I get another error, an assertion failure:
```
RevCPU[cpu:ECALL_mmap:386000]: ECALL: mmap called
FATAL: RevCPU[cpu:FetchAndDecodeInst:169784000]: Error: Core 0 failed to decode instruction at PC=0x11278; Inst=0
```

leekillough:
The assertion failure might be based on a page size assumption of the test. If I use the default page size, the test passes. There is no segfault anymore.

leekillough:
Please close this issue if you can verify that it's fixed.

Artlands:
Thank you for fixing the previous bug!

While reviewing the RevMem.cc code in the devel branch, I came across another issue that I wanted to bring to your attention. I believe it’s better to mention it here rather than opening a new issue.

[...]

PR #332, opened 2024-10-04
fix brace nesting in if statement; simplify code to use memcpy()

leekillough:
Fixes https://github.com/tactcomplabs/rev/issues/331

@Artlands

### DIFF
--- a/src/RevMem.cc
+++ b/src/RevMem.cc
@@ -642,6 +642,8 @@ bool RevMem::WriteMem( unsigned Hart, uint64_t Addr, size_t Len, const void* Dat
   uint64_t endOfPage   = ( pageMap[pageNum].first << addrShift ) + pageSize;
   char*    BaseMem     = &physMem[physAddr];
   char*    DataMem     = (char*) ( Data );
+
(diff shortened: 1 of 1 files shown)


src/RevMem.cc
```diff
@@ -755,31 +754,24 @@ bool RevMem::ReadMem( unsigned Hart, uint64_t Addr, size_t Len, void* Target, co
   char*    BaseMem     = &physMem[physAddr];
   char*    DataMem     = static_cast<char*>( Target );
 
+  // TODO: simplify to not consider crossing pages when using MemH
```
rkabrick, line 757: Is there a reason you don't want to do that in this PR?

rkabrick (approved):
If the TODO can wait then go ahead and merge

src/RevMem.cc
leekillough: > Is there a reason you don't want to do that in this PR?

This is a bugfix PR and I don't want to risk a non-bugfix change; I also wanted to consult with you or others before changing how the code handles crossing page boundaries.

PR #333, opened 2024-10-04
Refactor read/write functions to not use paging during MemH and to do paging calculations in a single function

leekillough:
In this PR:
- Remove checking page boundaries when MemH is used as described [here](https://github.com/tactcomplabs/rev/issues/331#issue-2566572802)
- Pass `0` as the third argument to `sendREADRequest()` and `sendWRITERequest()` since it isn't used
- Remove the deprecated `ReadMem()` and `WriteMem()` functions -- they have been deprecated for long enough to remove them as clutter
- Instead of having two `WriteMem()` functions, one with and one without a `RevFLag flag` last argument, make the flag argument have a default value of `RevFlag::F_NONE` which is the only difference between the two `WriteMem()` functions
- Make `ReadMem()` similarly have an optional flag at the end -- previously it was mandatory for `ReadMem()`
- Add a `AdjPageAddr()` function which consolidates page-crossing handling in one function:
  - It returns a `{ remainder, physAddr, adjPhysAddr }` tuple
  - The `remainder` is the remainder in the first page; the `physAddr` is the address in the first page; the `adjPhysAddr` is the address in the second page
  - It ensures that the `Len` parameter is `<= pageSize` because otherwise three or more pages would be needed
- Moves all of the `RevMem` address variables into the `if( ctrl ) { ... } else {` clause since they aren't needed by MemH

Cc: @Artlands

### DIFF
--- a/include/RevMem.h
+++ b/include/RevMem.h
@@ -166,14 +166,11 @@ public:
   // ----------------------------------------------------
   // ---- Base Memory Interfaces
   // ----------------------------------------------------
-  /// RevMem: write to the target memory location
-  bool WriteMem( unsigned Hart, uint64_t Addr, size_t Len, const void* Data );
-
   /// RevMem: write to the target memory location with the target flags
-  bool WriteMem( unsigned Hart, uint64_t Addr, size_t Len, const void* Data, RevFlag flags );
+  bool WriteMem( unsigned Hart, uint64_t Addr, size_t Len, const void* Data, RevFlag flags = RevFlag::F_NONE );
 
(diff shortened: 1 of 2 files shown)


leekillough:
Please review this so I can merge it ASAP.

jleidel (approved):
good to merge, note that we'll need to make similar changes in ForzaRev

leekillough:
> good to merge, note that we'll need to make similar changes in ForzaRev

They were already merged into ForzaRev; I may have to manually use `git merge -s ours` to mark it up to date with upstream.

PR #335, opened 2024-10-25
Add function to do safe casts from enum->int and int->int

leekillough:
Normally, `static_cast` must be used to convert `enum class` to integer types. This can be unsafe, because it may convert it to a smaller size than the underlying type of the `enum class`.

This adds a `safe_static_cast` function which acts just like `static_cast`, except it does not allow narrowing conversions (conversions which go from a larger integer/enumeration type to a smaller one).

All places where `static_cast` is used to convert `enum class` to integer have been replaced with `safe_static_cast`.

More changes like this will be needed on Forzarev, since it has many more `enum class` -> integer conversions.

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -36,9 +36,22 @@ namespace SST::RevCPU {
 
 // using float16 = _Float16;
 
+/// Safe non-narrowing cast of enum to integer type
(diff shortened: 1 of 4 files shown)


jleidel (approved):
good to merge

PR #336, opened 2024-11-04
testing improvements for tracer

kpgriesser:
This PR cleans up the test/tracer area by using the common sst configuration and removes the stale local ones.  The sst python configuration is enhanced to provided options for tracer controls. It also now allow specification of an output directory for csv files which is useful for running multiple tests in the some working directory.

The tracer test now includes code to check sign extension of the LW instruction which exposes some minor issues that will be the subject of a different PR.


test/tracer/bug-fastprint.cc
```diff
@@ -1,101 +0,0 @@
-/*
- * bug-fatprint.cpp: Testing various issues for fast printf development
```
leekillough, line 2: `.cc`

test/tracer/tracer.c
```diff
@@ -77,6 +89,49 @@ void* thread2() {
...
+  volatile uint64_t  checkneg = 0xffffffffbadbeef0;
+  TRACE_PUSH_ON
+lw_neq:
+  asm volatile( "ld   t3, 0(%1)   \n\t"     // t3 <- badneg
+                "ld   t4, 0(%2)   \n\t"     // t4 <- checkneg
+                "sd   t3, 4(%0)   \n\t"     // store badneg to upper half of mem[0]
+                "lw   t3, 0(%0)   \n\t"     // load mem[0]
+                "beq  t3, t4, _okneg \n\t"  // check
+                ".word 0x0 \n\t"            // fail
+                "_okneg: \n\t"
```
leekillough: This code is fine as-is; but when writing inline assembly with internal labels, it's good to use local labels, such as:
```
"beq  t3, t4, 0f    \n\t"
".word 0x0          \n\t"
"0:                 \n\t"
```
`0f` means "branch to the first label 0 going forwards". `beq t3, t4, 1b` would mean "branch to the first label 1 going backwards". The same numbers can be used multiple times because they bind forwards or backwards to the first matching label/branch and won't conflict if repeated later. This is good when the inline assembly is part of a macro, because it can be called several times and its local labels will be independent of local labels in other macro calls.

test/tracer/tracer.c
```diff
@@ -169,15 +226,17 @@ int main( int argc, char** argv ) {
...
+  printf( "Thread test starting\n" );
   rev_pthread_t tid1, tid2;
-  rev_pthread_create(&tid1, NULL, (void *)thread1, NULL);
-  rev_pthread_create(&tid2, NULL, (void *)thread2, NULL);
-  rev_pthread_join(tid1);
-  rev_pthread_join(tid2);
+  rev_pthread_create( &tid1, NULL, (void*) thread1, NULL );
+  rev_pthread_create( &tid2, NULL, (void*) thread2, NULL );
+  rev_pthread_join( tid1 );
+  rev_pthread_join( tid2 );
```
leekillough, line 237: I just ran `.githooks/pre-commit --reformat` on `rev/devel`, and saw no changes.

I think the MacOS version of `clang-format` you have, and the Ubuntu version of `clang-format` I have, have slight differences in how they handle spaces. So I see space formatting alternating back and forth sometimes.

Fortunately `pre-commit` only runs `clang-format` on files that a commit will change, unless you give it the `--reformat` flag.

Unfortunately `clang-format` does not allow the `.clang-format` file to specify a specific or even a minimum version of `clang-format` that its options should be interpreted as following, so it's hard to write `.clang-format` files which will work the same across different versions. There's nothing in `clang-format` like Perl's `require 5.6.1;`.

test/tracer/tracer.c
kpgriesser: I've changed the assembly code to use this local branching. I've run into the duplicate label problem many times so thanks for the tip!

test/tracer/bug-fastprint.cc
kpgriesser: This file should be deleted

test/tracer/tracer.c
```diff
@@ -77,6 +89,49 @@ void* thread2() {
...
+                "sd   t3, 4(%0)   \n\t"     // store badneg to upper half of mem[0]
+                "lw   t3, 0(%0)   \n\t"     // load mem[0]
+                "beq  t3, t4, _okneg \n\t"  // check
+                ".word 0x0 \n\t"            // fail
+                "_okneg: \n\t"
+                "sd   t3, 0(%0)  \n\t"  // store back into mem
+                : "=r"( p )
+                : "r"( &badneg ), "r"( &checkneg )
+                : "t3" );
+  printf( "neg result is 0x%llx\n", *p );
```
leekillough: `printf("neg result is 0x%" PRIx64 "\n", *p);` should be used instead.

`#include <inttypes.h>` to get the `PRI` and `SCN` type-specific format macros.

test/tracer/tracer.c
```diff
@@ -77,6 +89,49 @@ void* thread2() {
...
+                "sd   t3, 4(%0)   \n\t"     // store badpos to upper half of mem[0]
+                "lw   t3, 0(%0)   \n\t"     // load mem[0]
+                "beq  t3, t4, _okpos \n\t"  // check
+                ".word 0x0 \n\t"            // fail
+                "_okpos: \n\t"
+                "sd   t3, 0(%0)  \n\t"  // store back into mem
+                : "=r"( p )
+                : "r"( &badpos ), "r"( &checkpos )
+                : "t3" );
+  printf( "pos result is 0x%llx\n", *p );
```
leekillough: See above

test/tracer/tracer.c
```diff
@@ -77,6 +89,49 @@ void* thread2() {
...
+                "ld   t4, 0(%2)   \n\t"     // t4 <- checkneg
+                "sd   t3, 4(%0)   \n\t"     // store badneg to upper half of mem[0]
+                "lw   t3, 0(%0)   \n\t"     // load mem[0]
+                "beq  t3, t4, _okneg \n\t"  // check
+                ".word 0x0 \n\t"            // fail
+                "_okneg: \n\t"
+                "sd   t3, 0(%0)  \n\t"  // store back into mem
+                : "=r"( p )
+                : "r"( &badneg ), "r"( &checkneg )
+                : "t3" );
```
leekillough: `t4` needs to be in the list of clobbered registers, right?

test/tracer/tracer.c
```diff
@@ -77,6 +89,49 @@ void* thread2() {
...
+                "ld   t4, 0(%2)   \n\t"     // t4 <- checkpos
+                "sd   t3, 4(%0)   \n\t"     // store badpos to upper half of mem[0]
+                "lw   t3, 0(%0)   \n\t"     // load mem[0]
+                "beq  t3, t4, _okpos \n\t"  // check
+                ".word 0x0 \n\t"            // fail
+                "_okpos: \n\t"
+                "sd   t3, 0(%0)  \n\t"  // store back into mem
+                : "=r"( p )
+                : "r"( &badpos ), "r"( &checkpos )
+                : "t3" );
```
leekillough: See above about `t4`

kpgriesser:
updated tracer.c based on the PR feedback and will now merge

PR #340, opened 2024-11-10
Add warnings suggesting format and noreturn attributes, and override and final keywords

leekillough:
Add warnings suggesting the `__attribute__((format(...)))` and `[[noreturn]]` attributes when they are missing on a function which is passed `printf`/`scanf` format strings, or a function which never returns.

Add warnings suggesting the `override` and `final` inheritance-controlling keywords when they can enforce overriding or devirtualize method calls, respectively.

Fix code to fix warnings, including a couple from `clang-tidy`, such as `RevCore` functions which can be `static`, a variable assignment which isn't used, and a `[[noreturn]]` function which returned a value.

The `override` keyword implies `virtual` but also ensures that the function overrides a virtual function in a base class. Specifying `virtual` is redundant with `override`.

The `final` keyword prevents derived classes from overriding it. It requires `virtual`, either by overriding a virtual function in a base class, or by explicitly declaring it as the first and final instance of a virtual function with both `virtual` and `final`. If a `final` method is supposed to override a base class virtual function, then the `virtual` keyword should be left out and `final` is sufficient to ensure it overrides a base class virtual function -- `override` is never needed with `final`.

If a function is declared `final` and is invoked through a pointer / reference to a class type which is equal to or derived from the class type with the `final` method, then VTable lookups can be optimized away by the compiler, and a direct call to the derived class function can be made.

The [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override) recommend that exactly one of `virtual`, `override` or `final` be used for each virtual function declaration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ else()
   set(FP_MODE_FLAG "-frounding-math -ftrapping-math")
 endif()
 
-set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra ${WERROR_FLAG} -Wsuggest-override -Wsuggest-attribute=format -Wsuggest-attribute=noreturn -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Wvla -Wuninitialized -Wdouble-promotion -Wno-unused-parameter -Wno-Wmismatched-tags -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra ${WERROR_FLAG} -Wsuggest-override -Wsuggest-attribute=format -Wsuggest-attribute=noreturn -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Wvla -Wuninitialized -Wdouble-promotion -Wno-unused-parameter -Wno-mismatched-tags -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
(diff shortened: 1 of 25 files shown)


kpgriesser:
I compared the clang compile warnings on MacOS for the devel and suggest_override_final branch.  I was hoping there would be fewer warnings but it looks like some new ones cropped up (below). Probably not significant since these are all compiler 'suggestions' but nevertheless thought worth noting.

```
warning: unknown warning option '-Wsuggest-attribute=format'; did you mean '-Wproperty-attribute-mismatch'? [-Wunknown-warning-option]
warning: unknown warning option '-Wsuggest-attribute=noreturn'; did you mean '-Wbounds-attributes-redundant'? [-Wunknown-warning-option]
warning: unknown warning option '-Wsuggest-final-methods'; did you mean '-Wsuggest-override'? [-Wunknown-warning-option]
warning: unknown warning option '-Wsuggest-final-types' [-Wunknown-warning-option]
warning: unknown warning option '-Wvolatile' [-Wunknown-warning-option]
In file included from /Users/kgriesser/work/rev/src/RevExt.cc:11:
In file included from /Users/kgriesser/work/rev/include/RevExt.h:15:
/Users/kgriesser/work/rev/include/SST.h:21:32: warning: unknown warning group '-Wsuggest-final-methods', ignored [-Wunknown-warning-option]
   21 | #pragma GCC diagnostic ignored "-Wsuggest-final-methods"
      |                                ^
/Users/kgriesser/work/rev/include/SST.h:22:32: warning: unknown warning group '-Wsuggest-final-types', ignored [-Wunknown-warning-option]
   22 | #pragma GCC diagnostic ignored "-Wsuggest-final-types"
```

leekillough:
I can fix it to work with multiple compilers; sometimes the fix involves using different options or `#pragma`s; other times the warning can be fixed by changing the source (which is the intention of the warning).

leekillough:
@kpgriesser Try it now. It should have removed all Clang warnings.

kpgriesser:
Compared the devel branch and confirmed that that all the prior warnings are cleaned up and that there are no new ones using MacOS clang.

kpgriesser (approved):
see previous comment

PR #341, opened 2024-11-26
Add optional CSR handlers

leekillough:
This adds optional CSR handlers for getting/setting CSR registers. 

Some CSR registers, like the floating-point and performance counter registers, are already handled specially in `RevCSR.h` with `switch` statements, but they could be registered as handlers dynamically at runtime instead of being implemented statically in `RevCSR` or in one of its base classes like `RevZicntr`. After this change, there are three ways to implement CSR registers:
1. Implement them in the `RevCSR` class, like the floating-point CSR registers are handled today. A `switch` statement handles the `fcsr` register and the `frm` and `fflags` pseudo-registers.
2. Implement them in a base class of `RevCSR`, like the `Zicntr` performance counters are handled in `RevZicntr.h`, and dispatch to the base class functions which get/set them in a `switch` statement. The code is kept mostly separate from the `RevCSR` class, except that the CSR registers it needs to handle need to be dispatched in a `switch` statement.
3. Register Getter/Setter functions for specific CSR registers at runtime, so that `GetCSR()` and `SetCSR()` call these callback functions for specific CSR registers.

This PR allows any code, including coprocessor code, to register handler functions for specific CSR registers, without changing the decoding and execution of `Ziscr` extension CSR instructions.

The Getter function takes a `uint16_t` CSR register number and returns a `uint64_t` value (it does not matter whether XLEN==32 or XLEN==64 because `GetCSR()` explicitly casts it to the `XLEN` type, and `Zicsr.h` calls `GetCSR()` with either `XLEN=uint32_t` or `XLEN=uint64_t` depending on the mode in operation).

The Setter function takes a `uint16_t` CSR register number and a `uint64_t` value to set it to, and returns a `bool` indicating whether it was successful. If it returns `false`, a fatal error in the simulation occurs.

The Getter and Setter functions take a `uint16_t` argument so that the same function can be registered for multiple CSR registers and differentiate them when it is called.

The `SetCSRGetter()` and `SetCSRSetter()` functions in `RevCSR` allow a Getter and Setter function to be set. The `GetCSRGetter()` and `GetCSRGetter()` return the Getter and Setter functions set in `RevCSR`, or fall back to the Getter and Setter functions set in the parent `RevCore` if they were not set in `RevCSR`.

The default behavior of `GetCSR()` and `SetCSR()` is to read or write CSR registers as if they were ordinary registers, without any special semantics. If the CSR registers are simply load/store registers like GP registers, then no special handler needs to be registered if the CSR registers are hart-local.

All 4096 CSR registers **reside architecturally on the main core side**, but if they are core-local CSR registers instead of hart-local CSR registers, then the getters/setters need to store their data per-core and Getters/Setters be set in `RevCore`. The `RevCore` class has similar `SetCSRGetter()`, `SetCSRSetter()`, `GetCSRGetter()` and `GetCSRSetter()` functions.

All 6 CSR operational instructions (`csrrw`, `csrrs`, `csrrc`, `csrrwi`, `csrrsi`, `csrrci`) are handled exactly the same way for all 4096 CSR registers in `Zicsr.h` as a combination of optional read, optional modify, and optional write which is precisely specified in the `Zicsr` specification. There should be no handling of specific CSR registers in `Zicsr.h`, except that pseudoinstructions can be differentiated from their generic equivalents for friendly disassembly.

This adds the vector CSR registers by name and number in the `RevCSR` `enum`, but it removes any mention of whether they are implemented in a coprocessor. This is just a table of CSR register names and numbers made easy to access such as `RevCSR::vl`, but their semantics are handled elsewhere, and entries in this table do not imply the presence of any related extension.

### DIFF
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -39,19 +39,27 @@ namespace SST::RevCPU {
 /// Safe non-narrowing cast of enum to integer type
 /// C++17 allows non-narrowing cast of integer to scoped enum, but not the reverse
 template<typename INT, typename ENUM, typename = decltype( INT{ std::declval<std::underlying_type_t<ENUM>>() } )>
-inline constexpr std::enable_if_t<std::is_integral_v<INT> && std::is_enum_v<ENUM>, INT> safe_static_cast( ENUM e ) {
+constexpr std::enable_if_t<std::is_integral_v<INT> && std::is_enum_v<ENUM>, INT> safe_static_cast( ENUM e ) {
   return static_cast<INT>( e );
 }
 
 /// Allow non-narrowing int->int cast with enum_int_cast
 template<typename INT, typename ENUM, typename = decltype( INT{ std::declval<ENUM>() } )>
-inline constexpr std::enable_if_t<std::is_integral_v<INT> && std::is_integral_v<ENUM>, INT> safe_static_cast( ENUM e ) {
+constexpr std::enable_if_t<std::is_integral_v<INT> && std::is_integral_v<ENUM>, INT> safe_static_cast( ENUM e ) {
   return static_cast<INT>( e );
+}
+
+/// Make an expression dependent on arbitrary template type parameters.
+/// This has the effect of lazy evaluation of an expression until a template containing it is instantiated.
+/// It can be used to wrap an incomplete class type which will be completed by template instantiation time.
+template<typename..., typename T>
+constexpr T&& make_dependent( T&& x ) {
+  return std::forward<T>( x );
 }
 
 /// Zero-extend value of bits size
 template<typename T>
-inline constexpr auto ZeroExt( T val, size_t bits ) {
+constexpr auto ZeroExt( T val, size_t bits ) {
   return static_cast<std::make_unsigned_t<T>>( val ) & ( ( std::make_unsigned_t<T>( 1 ) << bits ) - 1 );
 }
 
@@ -64,7 +72,7 @@ constexpr auto SignExt( T val, size_t bits ) {
 
 /// Base-2 logarithm of integers
 template<typename T>
-inline constexpr int lg( T x ) {
+constexpr int lg( T x ) {
   static_assert( std::is_integral_v<T> );
 
   // We select the __builtin_clz which takes integers no smaller than x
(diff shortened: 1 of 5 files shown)


leekillough:
If we anticipate the need to access the calling hart's `RevRegFile` or parent core in one of these callback functions, then parameters like those can be added as additional parameters besides the CSR register number, so that, for example, a core's coprocessor's single vector CSR state can be accessed and manipulated by any of that core's harts by pointing to its parent core. (CSR access instructions are atomic, so a core should not switch harts in the middle of a CSR instruction.)

Also, if there is a CSR state which is not per-hart, but is per-core or per-CPU, we can add CSR handler mappings in `RevCore` and `RevCPU` and use the first one found up the tree in `RevCSR`, `RevCore`, and `RevCPU`, so that harts do not necessarily have to concern themselves with setting up handlers for CSR registers which are per-core or per-CPU.

kpgriesser:
Concept and code look great. Once I finish testing it should be good to go from my side

kpgriesser:
I believe you're working on this problem already but just wanted to mention it in this PR in case someone trips on it. The CSR access will not be issued into the coprocessor instruction queue so it will occur out of order with respect to the coprocessor instructions completing. Architecturally, I'm not sure if in order execution is a requirement but it's important for software developers to understand this and create any necessary barriers. At some point, a general synchronization method that allows the coprocessor to decide when the CSR access occurs would be extremely helpful in supporting a 'precise' mode of execution. Additionally, if we are more closely modeling hardware with an independent host controller than can modify CSR's we'll need to guarantee atomicity.
Again, just want to capture this here. What you have currently will work great for what we need for the near-term.

leekillough:
CSR scoreboard is one of the requirements on my list.

leekillough:
Also note that the CSR access instructions are atomic, so a core cannot switch harts in the middle of a CSR instruction. Also regardless of when CSR registers are updated, their reads are atomic. Scoreboarding may be used to stall a hart until a CSR is written by an outside processor. But that is more of a synchronization issue between the hart and a coprocessor than a CSR issue.

CSR registers should be per-hart, and the architectural state includes all 4096 CSR registers, which are like a RAM and are accessed orthogonally with 6 CSR access instructions. All 4096 CSR registers should be handled uniformly.

This PR allows callbacks to get/set particular CSR registers, but any synchronization should be handled uniformly for all CSR registers with scoreboarding or another method, rather than developing a complicated synchronization method for specific CSR registers like in the vector unit.

kpgriesser (approved):
Confirmed working with coprocessor

PR #342, opened 2024-12-02
Macroize fli instruction table generation

leekillough:
This uses a new macro `FLI( index, mnemonic, value)` to generate `fli` instructions in the `Zfa` extension. Since there are 32 constants for each floating-point type, this makes the code more concise, and will allow F16 and BF16 to be more easily added later.

### DIFF
--- a/include/insns/Zfa.h
+++ b/include/insns/Zfa.h
@@ -162,46 +162,63 @@ class Zfa : public RevExt {
     }
   };
 
+  // Macro to generate FLI entries
+#define FLI( index, mnemonic, value )                                                  \
+  ( RevZfaInstDefaults()                                                               \
+      .SetMnemonic( #mnemonic " %rd, " #value )                                        \
+      .SetPredicate( []( uint32_t Inst ) { return DECODE_RS1( Inst ) == ( index ); } ) \
+      .SetImplFunc( []( auto*, auto* R, auto*, auto& Inst ) {                          \
+        R->SetFP( Inst.rd, ( value ) );                                                \
+        R->AdvancePC( Inst );                                                          \
+        return true;                                                                   \
+      } ) )
+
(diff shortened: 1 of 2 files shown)


include/insns/Zfa.h
```diff
@@ -162,46 +162,63 @@ class Zfa : public RevExt {
...
+        R->SetFP( Inst.rd, ( value ) );                                                \
+        R->AdvancePC( Inst );                                                          \
+        return true;                                                                   \
+      } ) )
+
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
   // Single-Precision Instructions
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#undef min
```
kpgriesser, line 180: Does having to 'undef' the macro imply there is a risk that this macro could conflict with one in another header file?

kpgriesser (approved):

include/insns/Zfa.h
leekillough: > Does having to 'undef' the macro imply there is a risk that this macro could conflict with one in another header file?

No, it's just that redefining an already-defined macro is an error, and so when I write files with changing `#define` definitions with the same macro names (here there's one for `fli.s` and one for `fli.d`), I `#undef` each one before (re)defining it, so that the sections with different `#define` definitions can be changed around in any order and not cause a redefinition error.

`std::min` and `std::max` are defined, but they don't have anything to do with macros such as these.

However, there is a chance that if this `Zfa.h` header file is `#include`d in another one which uses `std::max` or `std::min` later on, the `min` and `max` macros will be expanded there and cause errors. I should `#undef` these macros at the end of `Zfa.h`.

The reason the macro names are as they are (`min`, `inf`, `nan`) and not longer, less common names, is because those are the RISC-V assembly mnemonics for those floating-point values, and I am stringizing the values to be used in the disassembly mnemonics as well as the FP values.

PR #343, opened 2024-12-03
Handle iconv errors more gracefully

leekillough:
`iconv` is used in `pre-commit` to convert non-ASCII symbols like smart quotes or continuous vertical or horizontal lines with no gaps into ASCII equivalents.

On MacOS this sometimes fails because it cannot find a conversion. In the example of horizontal and vertical lines @kpgriesser  put in `README.md`, the `iconv` on MacOS failed and the `pre-commit` script failed silently, preventing a commit unless `--no-verify` is used. On Linux, `iconv` converts these characters just fine, replacing them with `|` `+` and `-` characters.

This PR makes `pre-commit` display an informative error message if `iconv` fails, rather than failing silently.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -195,7 +195,18 @@ reformat_files() {
     for file in "${FILES[@]}"; do
         if [[ -f $file ]] && is_text_file "$file"; then
             # Replace non-ASCII characters with ASCII equivalents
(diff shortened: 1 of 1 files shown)


kpgriesser (approved):
My test case on MacOS now prints out an informational message. Good to merge.

PR #345, opened 2024-12-05
Add bit manipulation functions and prevent implicit changes in integer signedness

leekillough:
Three new bit manipulation functions are added:

- `BitExtract<pos, width>( T x )` returns bits `[ pos + width - 1 : pos ]` of `x` and is zero-extended or sign-extended depending on the signedness of `x`. You can also use `BitExtract<pos, width, type>( x )` to coerce the return type and signedness to `type` in case `X` has a different size or signedness as the desired result.

- `BitDeposit<pos, width>( T &x, U y )` deposits the lower `width` bits of `y` into bits `[ pos + width - 1 : pos ]` of `x`. It returns a reference to `x` so that multiple calls can be chained, e.g. `BitDeposit<7, 1>( BitDeposit<0, 3>( Vtype, vlmul ), vma ) )`.

Those instructions are analogous to the PA-RISC extract and deposit instructions.

- `BitShift( T x, int shift )` returns `x` shifted left `shift` bits, or shifted right `-shift` bits if `shift` is negative. This effectively returns `x * 2 ^ shift`.

Add `-Wsign-conversion` to the compiler warning flags, which warns whenever the signedness of an integer is implicitly changed. This can occur, for example, when `uint8_t` (unsigned) or `uint16_t` (unsigned) is implicitly promoted to `int` (signed).  In C/C++ arithmetic expressions, any type smaller than `int` is implicitly promoted to `int`, which can change an 8- or 16-bit unsigned value into a 32-bit signed `int` value. This warning prevents this silent conversion, requiring explicit casts to suppress it.

`unsigned char*` is used instead of `char*` when using pointers to bytes in the Rev functions such as memory instructions. `char*` is still used when it involves calling standard library routines such as `c_str()` which returns `const char*`, or when the pointer is initialized from a `"string literal"`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile")
 endif()
 
-set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations -Werror ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations -Wno-macro-redefined -Werror ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -Wall ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall ${REVCPU_COMPILER_MACROS}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Werror")
 endif()
 
-set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -Wall ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall ${REVCPU_COMPILER_MACROS}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,10 +80,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
 else()
   set(FP_MODE_FLAG "-frounding-math -ftrapping-math")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Werror")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile")
 endif()
 
-set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations -Werror ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -Wall ${REVCPU_COMPILER_MACROS}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall ${REVCPU_COMPILER_MACROS}")
 
--- a/common/include/RevCommon.h
+++ b/common/include/RevCommon.h
@@ -11,6 +11,7 @@
 #ifndef __REV_COMMON__
 #define __REV_COMMON__
 
+#include <climits>
 #include <cstddef>
 #include <cstdint>
 #include <functional>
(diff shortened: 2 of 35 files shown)


kpgriesser:
Changes to the CMakeLists.txt is causing this to fail compilation on MacOs. Were those changes intended?

common/include/RevCommon.h
```diff
@@ -59,32 +60,54 @@ constexpr T&& make_dependent( T&& x ) {
...
 
 /// Base-2 logarithm of integers
 template<typename T>
 constexpr int lg( T x ) {
   static_assert( std::is_integral_v<T> );
-
   // We select the __builtin_clz which takes integers no smaller than x
   if constexpr( sizeof( x ) <= sizeof( int ) ) {
-    return x ? 8 * sizeof( int ) - 1 - __builtin_clz( x ) : -1;
+    return x ? int{ 8 * sizeof( int ) - 1 } - __builtin_clz( x ) : INT_MIN;
```
kpgriesser, line 81: A pitfall for cls, ctz, ....: if x==0 then  __builtin_clz is undefined. I always check for 0 before using these. I guess since the function explicitly is called 'logarithm' the user should know this but maybe worth mentioning in the comments that x cannot be 0.

CMakeLists.txt
```diff
@@ -80,10 +80,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
 else()
   set(FP_MODE_FLAG "-frounding-math -ftrapping-math")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Werror")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile")
 endif()
 
-set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wno-unused-parameter -Wno-deprecated-declarations ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${FP_MODE_FLAG} -O2 -Wall -Wextra -Wsuggest-override -Wmissing-noreturn -Wvla -Wuninitialized -Wdouble-promotion -Wsign-conversion -Wno-unused-parameter -Wno-deprecated-declarations -Werror ${CMAKE_CXX_FLAGS} -I./ ${LDFLAGS} ${REVCPU_COMPILER_MACROS}")
```
kpgriesser: -Werror causes MacOS build to fail

include/RevCSR.h
```diff
@@ -541,9 +541,9 @@ struct RevCSR : RevZicntr {
     // clang-format off
     switch( csr ) {
       // Floating Point flags
-      case fflags: CSR[fcsr] = ( CSR[fcsr] & ~uint64_t{ 0b00011111 } ) | ( val & 0b00011111 ) << 0; break;
-      case frm:    CSR[fcsr] = ( CSR[fcsr] & ~uint64_t{ 0b11100000 } ) | ( val & 0b00000111 ) << 5; break;
-      case fcsr:   CSR[fcsr] = ( CSR[fcsr] & ~uint64_t{ 0b11111111 } ) | ( val & 0b11111111 ) << 0; break;
+      case fflags: BitDeposit<0, 5>(CSR[fcsr], val); break;
+      case frm:    BitDeposit<5, 3>(CSR[fcsr], val); break;
```
kpgriesser, line 545: Looks like the original behavior was different. The bit masks would clear bits 3 and 4 which I suspect was a bug or you wanted to make those hard coded to zero.  If it's the former then it shows the new BitDeposit code is less error prone. If the latter then adding a masked write function would be handy ( a & ~mask) | ( b & mask ).

src/RevMem.cc
```diff
@@ -13,45 +13,31 @@
...
-  stacktop  = ( _REVMEM_BASE_ + memSize ) - 1024;
+// We initialize StackTop to the size of memory minus 1024 bytes
+// This allocates 1024 bytes for program header information to contain
+// the ARGC and ARGV information
 
-  // Add the 1024 bytes for the program header information
-  AddMemSegAt( stacktop, 1024 );
+// Note: this constructor assumes the use of the memHierarchy backend
+RevMem::RevMem( uint64_t memSize, RevOpts* opts, RevMemCtrl* ctrl, SST::Output* output )
+  : memSize( memSize ), opts( opts ), ctrl( ctrl ), output( output ), pageSize( 262144 ), addrShift( uint32_t( lg( pageSize ) ) ),
```
kpgriesser, line 29: These changes to RevMem look like more than bit manipulation. @rkabrick better to review

src/RevSysCalls.cc
```diff
@@ -1217,8 +1217,8 @@ EcallStatus RevCore::ECALL_clock_gettime() {
   }
   memset( &src, 0, sizeof( *tp ) );
   SimTime_t x = timeConverter->convertToCoreTime( Stats.totalCycles );
-  src.tv_sec  = x / 1000000000000ull;
-  src.tv_nsec = ( x / 1000 ) % 1000000000ull;
+  src.tv_sec  = time_t( x / 1000000000000 );
```
kpgriesser, line 1220: just curious, does 'ull' cause a problem here? I tend to use it to be very explicit about the data type.

common/include/RevCommon.h
leekillough: I don't understand. When `x==0` `lg() ` used to return `-1`, but now it returns `INT_MIN` (the closest integer to -Infinity).

Nothing has changed except the return value for `lg(0)`. `lg()` has existed for over a year.

leekillough:
> Changes to the CMakeLists.txt is causing this to fail compilation on MacOs. Were those changes intended?

All changes were intended. I am looking for errors/warnings and will fix them, just like I fixed the previous MacOS/Clang build errors. Every compiler is slightly different. I don't have a MacOS test machine.

leekillough:
I don't understand the build errors, which occur on my machine now but didn't occur yesterday.

This is a slow experimental process, and needs to be tested on your machine, my machine, and the CI system.

All changes are, in principle, intentional.

CMakeLists.txt
```diff
@@ -80,10 +80,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
 else()
   set(FP_MODE_FLAG "-frounding-math -ftrapping-math")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmissing-format-attribute -Wsuggest-final-methods -Wsuggest-final-types -Wvolatile -Werror")
```
rkabrick, line 83: Why remove -Werror?

CMakeLists.txt
rkabrick: Disregard my comment I'm dumb

rkabrick:
Looks good to me

CMakeLists.txt
leekillough: I made it global. Previously it was limited to GCC and not Clang. I'm currently removing build errors.

It doesn't build in my environment right now, probably because Git silently broke it during an otherwise conflict-free merge of `devel` into this PR's branch.

src/RevSysCalls.cc
leekillough: Suffixes are not needed for constants except in specific situations. A constant ([integer literal](https://en.cppreference.com/w/cpp/language/integer_literal)) will be of the smallest type capable of holding its value. See the linked page for the priority of types based on the base and suffix.

Also, we don't know that `uint64_t` is necessarily the same type as `ull`. Best to avoid suffixes, and then the type of the constant will be the smallest type which can hold its value, and if it's stored in a variable, the conversions are well-defined.

When constants of a very specific type are needed, I use uniform initialization syntax such as `uint64_t{1}`. The braces `{}` ensure that the value is not narrowed (truncated) by the type conversion, while parentheses `uint64_t(1)` would allow narrowing and are a function-style cast with the same meaning as `((uint64_t) 1)`.

src/RevMem.cc
leekillough: There were problems with pointer types like `char*` and I fixed them and condensed the code by putting it in the initialization list rather than assigning to members inside the constructor body. The compiler can usually optimize away double-initializations, but technically, all members of a class are "initialized" in a constructor initialization list rather than assignments made in the body of the constructor.

include/RevCSR.h
leekillough: I don't understand. Bits 3 and 4 are part of the `fflags` portion of `fcsr`, and are not cleared.

`<<` has higher precedence than `&` and `|`. Precedence (highest to lowest):

`type{}` (uniform initialization cast)
`~` (complement)
`<<` or `>>` (shift)
`&` (and)
`|` (or)

`case frm:    CSR[fcsr] = ( CSR[fcsr] & ~uint64_t{ 0b11100000 } ) | ( val & 0b00000111 ) << 5`

If the CSR pseudo-register is `frm`, the lower 3 bits of `val` are deposited in bits `[7:5]` of `fcsr`, and the rest of the bits in `fcsr` are unchanged. New equivalent:

`  case frm:    BitDeposit<5, 3>(CSR[fcsr], val); break;`

Deposits lowest 3 bits of `val` into bits 5, 6, 7 of `CSR[fcsr]`, leaving the rest of `CSR[fcsr]` unchanged.

`case fcsr:   CSR[fcsr] = ( CSR[fcsr] & ~uint64_t{ 0b11111111 } ) | ( val & 0b11111111 ) << 0; break;`

If the `fcsr` CSR register is written, the lower 8 bits of `val` are deposited in bits `[7:0]` of `fcsr`, and the rest of the bits in `fcsr` are unchanged.

-----

FP CSR registers:

`fcsr`: `reserved` `[31:8]` ` frm` `[7:5]`  `fflags` `[4:0]`

`frm` and `fflags` are _pseudo_-CSR registers which refer to partial contents as `fcsr` but are all anchored with LSB = bit 0.

So if you store a 3-bit value to `frm`, you are really depositing it into bits `[7:5]` of `fcsr`, and if you store a 5-bit value in `fflags`, you are really depositing it into bits `[4:0]` of `fcsr`. You can also write 8 bits into `[7:0]` of `fcsr`.

CMakeLists.txt
leekillough: Like the previous PR, the warnings causing that failure need to be fixed. We shouldn't just sweep warnings under the rug.

common/include/RevCommon.h
leekillough: The `lg()`  function has always handled `0` as a special case. Look at the previous code. I just changed the code to explicitly cast the `size_t` value of `8 * sizeof (int ) -1` to `int{}` so that we don't get signedness-changing error, and I changed `-1` to `INT_MIN` for the return value of `lg(0)`.

I don't think we need to remind people that `lg(0) = -Infinity` or raise exceptions. This `lg()` function is `constexpr` and is intended for compile-time evaluation. It's not intended to be passed arguments from the end-user, but it passed arguments that we control in Rev. We are not going to call `lg()` with arguments dependent on end-user arguments that we need to flag as user errors.

kpgriesser:
glad to run build test cases on my Mac. Ping me on slack anytime to pull / test.
As long as it builds cleanly, I'm fine to merge

leekillough:
Changes have been pushed:
- `-Wsign-conversion` is disabled during `#include "disasm.h"` which is the Spike disassembler
- `-Wno-macro-redefined` option has been added to the compiler options. GCC does not support a `#pragma` version of it.

include/RevCSR.h
kpgriesser: I overlooked the <<5 of the second term. So this is ok

PR #346, opened 2024-12-05
Make sure file extensions and shebangs are as expected; check for common virtual/final/override mistakes

leekillough:
This makes sure that the `mime-type` reported by `file` and the filename extension match each other for `.sh`, `.py` and `.pl` files. If one of those extensions is used but `file` does not report the matching `mime-type`, an error is printed that a shebang `#!` needs to be added to the top of the file. If `file` reports the language's `mime-type` but the filename does not match the extension, such as using `.bash` instead of `.sh` extensions for shell scripts, it reports an filename extension error.

Common mistake combinations of `virtual`, `override` and `final` are found by using trivial regular expressions. It won't catch all mistakes, but it will catch most of them. At most one of `virtual`, `override` or `final` should be used on a member function declaration. `virtual` indicates the most-base class's virtual function declaration. `override` indicates a mid-hierarchy class override. `final` indicates the final override. Already, GCC is passed `-Wsuggest-final-methods -Wsuggest-final-types` and both GCC and Clang are passed `-Wsuggest-override` to print some of the warnings, but they don't usually enforce using only one of `virtual`, `override` or `final`, as specified in the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override).

(A corner case is when `virtual` and `final` are both used, which indicates that a class's virtual function is the first and final definition of  it. It only serves to prevent it from being overriden by functions of the same signature in derived classes. This is only useful if the class it belongs to is not itself marked `final`, and the class has _other_ virtual functions which _can_ be overriden. This is a rare use case that we don't take it into account, and so we reject `virtual` and `final` both being used on the same function.)


dawsfox (approved):
Looks good to me as far as I can tell

kpgriesser:
Test driving;
```
./scripts/rev-print.py --help
env: python: No such file or directory
```

The change made to the script was
```
-#!/usr/bin/python3
+#!/usr/bin/env python
```

It works with
#!/usr/bin/env python3

Can the checker respect the python version?

kpgriesser:
will create a symlink to resolve this - go ahead and push

kpgriesser (approved):

leekillough:
Python 2 is dead; I would have thought that `#!/usr/bin/env python` would find the "default" Python interpreter, which should be `python3` now. But I can change it to suggest/use `#!/usr/bin/env python3` instead.

PR #349, opened 2024-12-14
Fix shebang error message to print python3 instead of python

leekillough:


kpgriesser (approved):
tested ok on Mac and ubuntu

PR #350, opened 2024-12-15
change pre-commit handling of itself

leekillough:
Change pattern matching so that `pre-commit` script recognizes itself under any directory levels.


kpgriesser (approved):
tested on MacOS and Ubuntu. Modified several files as well as .githooks/pre-commit. No issues.

PR #352, opened 2024-12-23
Simplify rev opts

leekillough:
This change removes duplicate code in `RevOpts.h` and `RevOpts.cc` by templatizing the functions.

- The mappings are simplified to be `std::vector` instead of `std::unordered_map`, since the indices are integers and we know `numCores` in advance
- However, `startSym` must be a map rather than a vector of initialized elements, because of how it is constructed piecemeal
- Therefore the code has been made general, with templates, `if constexpr` and generic lambdas, to handle any `std::vector`, `std::unordered_map`, or `std::map`.
- The `std::vector` mapping members have default initializers which are based on the value of `numCores`, which is already initialized before them because it's declared before them. This initializes them the way they were initialized in a loop in the `RevOpts` constructor before.
- `numCores`, `numHarts` and `verbosity` have been marked as `const` members, so that failure to initialize them in the constructor is an error, and because they never change after construction
- A new `GetProperty()` template function returns the `Core -> Val` mappings. For `std::vector` it returns `true` iff `Core < numCores`, and for `std::unordered_map` it returns `true` iff a mapping entry for `Core` exists.
- It also is able to accept `std::tie()` as an assignable rvalue-reference argument for its destination, in order to unpack a pair or tuple into a set of destinations. This is how `GetMemCost()` extracts both the `Min` and `Max` values from a vector of pairs, by simply passing `std::tie( Min, Max )` to ` GetProperty()`.
- A new `InitPropertyMap()` template function parses the options, setting one for each `Core:Value` pair. It will work for both vector and map containers, using `templates`, `if constexpr`, and `lambdas` to make the code compact . 
- A new `InitPropertyMapCores()` template function allows an optional `CORES:value` and broadcasts the value to all cores. If it does not detect `CORES:value`, it falls back on `InitPropertyMap()` which requires per-element values.
- Instead of calling `map.insert(std::pair<type,type>(i, value))` in a loop to initialize the mappings, default member initialization is used to initialize the vectors with `numCores` copies of a value.
- Some mappings go from `Core` to an integer value, while others go from `Core` to a string. With `if constexpr()` both cases can be handled with an `if` which is resolved in the template at compile-time and whose `false` case is ignored so it cannot produce errors
- Some mappings are handed with `std::vector` while others are handled with `std::unordered_map`. With `if constexpr()` both cases can be similarly handled. To prevent source code explosion, generic lambdas are used to handle multiple cases with small code.
- The old code had an off-by-one error. It was only comparing `Core > numCores`, potentially allowing `Core == numCores` which is outside of the range of `[0, numCores-1]`. 
- The old code used the unsafe `map.find( Core )->second = value` which could crash if the key was not already in the map. Because all values were initialized with defaults, it `map.find( Core )` never returned `map.end()` which is not dereferenceable. I've replaced it with the simpler syntax `map[ Core] = value` which works for both `std::vector` and `std::unordered_map` and in the latter case, creates a new entry if one doesn't already exist.
- Because map keys and values might have different types, I've used explicit casting based on the underlying `decltype`s when calling `std::stoull` to avoid potential warnings. This way the declaration of the map variables is the only place where the underlying types need to be declared.

### DIFF
--- a/include/RevOpts.h
+++ b/include/RevOpts.h
@@ -52,7 +52,7 @@ class RevOpts {
   bool InitPropertyMap( const std::vector<std::string>& Opts, VEC& map );
 
   template<typename VEC>
-  std::pair<bool, bool> InitPropertyMapCores( const std::vector<std::string>& Opts, VEC& map );
+  bool InitPropertyMapCores( const std::vector<std::string>& Opts, VEC& map );
 
 public:
   /// RevOpts: options constructor
--- a/include/RevOpts.h
+++ b/include/RevOpts.h
@@ -60,17 +60,23 @@ public:
   RevOpts( uint32_t NumCores, uint32_t NumHarts, int Verbosity )
     : numCores( NumCores ), numHarts( NumHarts ), verbosity( Verbosity ) {}
 
+  /// RevOpts: Disallow copying and assignment
+  RevOpts( const RevOpts& )            = delete;
+  RevOpts( RevOpts&& )                 = delete;
+  RevOpts& operator=( const RevOpts& ) = delete;
+  RevOpts& operator=( RevOpts&& )      = delete;
+
   /// RevOpts: options destructor
-  ~RevOpts() = default;
+  ~RevOpts()                           = default;
 
   /// RevOpts: retrieve the number of configured cores
-  uint32_t GetNumCores() { return numCores; }
+  uint32_t GetNumCores() const { return numCores; }
 
   /// RevOpts: retrieve the number of configured harts per core
-  uint32_t GetNumHarts() { return numHarts; }
+  uint32_t GetNumHarts() const { return numHarts; }
 
   /// RevOpts: retrieve the verbosity level
-  int GetVerbosity() { return verbosity; }
+  int GetVerbosity() const { return verbosity; }
 
   /// RevOpts: initialize the set of starting addresses
   bool InitStartAddrs( const std::vector<std::string>& StartAddrs );
@@ -108,7 +114,7 @@ public:
   }
 
   /// RevOpts: retrieve the prefetch depth for the target core
-  bool GetPrefetchDepth( uint32_t Core, uint32_t& Depth ) { return GetProperty( Core, prefetchDepth, Depth ); }
+  bool GetPrefetchDepth( uint32_t Core, uint32_t& Depth ) const { return GetProperty( Core, prefetchDepth, Depth ); }
 
   /// RevOpts: set the argv array
   void SetArgs( const SST::Params& params );
@@ -120,7 +126,7 @@ public:
   static void splitStr( std::string s, const char* delim, std::vector<std::string>& v ) {
     char* ptr     = s.data();
     char* saveptr = nullptr;
-    for( v.clear(); auto token = strtok_r( ptr, delim, &saveptr ); ptr = nullptr )
+    for( v.clear(); char* token = strtok_r( ptr, delim, &saveptr ); ptr = nullptr )
       v.push_back( token );
   }
 
--- a/include/RevOpts.h
+++ b/include/RevOpts.h
@@ -20,6 +20,7 @@
 #include <string>
 #include <tuple>
 #include <type_traits>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
--- a/include/RevOpts.h
+++ b/include/RevOpts.h
@@ -18,15 +18,31 @@
 #include <cinttypes>
 #include <map>
 #include <string>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
 namespace SST::RevCPU {
 
 class RevOpts {
+  template<typename VAL>
+  bool GetProperty( uint32_t Core, const std::vector<VAL>& Vec, VAL& Val ) const {
+    if( Core >= Vec.size() )
+      return false;
+
+    Val = Vec[Core];
+    return true;
+  }
+
+  template<typename MAP>
+  bool InitPropertyMap( const std::vector<std::string>& Opts, MAP& map );
+
+  template<typename MAP>
+  std::pair<bool, bool> InitPropertyMapCores( const std::vector<std::string>& Opts, MAP& map );
+
 public:
   /// RevOpts: options constructor
-  RevOpts( uint32_t NumCores, uint32_t NumHarts, const int Verbosity );
+  RevOpts( uint32_t NumCores, uint32_t NumHarts, int Verbosity );
 
   /// RevOpts: options destructor
   ~RevOpts() = default;
@@ -59,22 +75,27 @@ public:
   bool InitPrefetchDepth( const std::vector<std::string>& Depths );
 
   /// RevOpts: retrieve the start address for the target core
-  bool GetStartAddr( uint32_t Core, uint64_t& StartAddr );
+  bool GetStartAddr( uint32_t Core, uint64_t& StartAddr ) const { return GetProperty( Core, startAddr, StartAddr ); }
 
   /// RevOpts: retrieve the start symbol for the target core
-  bool GetStartSymbol( uint32_t Core, std::string& Symbol );
+  bool GetStartSymbol( uint32_t Core, std::string& Symbol ) const { return GetProperty( Core, startSym, Symbol ); }
 
   /// RevOpts: retrieve the machine model string for the target core
-  bool GetMachineModel( uint32_t Core, std::string& MachModel );
+  bool GetMachineModel( uint32_t Core, std::string& MachModel ) const { return GetProperty( Core, machine, MachModel ); }
 
   /// RevOpts: retrieve instruction table for the target core
-  bool GetInstTable( uint32_t Core, std::string& Table );
+  bool GetInstTable( uint32_t Core, std::string& Table ) const { return GetProperty( Core, table, Table ); }
 
   /// RevOpts: retrieve the memory cost range for the target core
-  bool GetMemCost( uint32_t Core, uint32_t& Min, uint32_t& Max );
+  bool GetMemCost( uint32_t Core, uint32_t& Min, uint32_t& Max ) const {
+    if( Core >= memCosts.size() )
+      return false;
+    std::tie( Min, Max ) = memCosts[Core];
+    return true;
+  }
 
   /// RevOpts: retrieve the prefetch depth for the target core
-  bool GetPrefetchDepth( uint32_t Core, uint32_t& Depth );
+  bool GetPrefetchDepth( uint32_t Core, uint32_t& Depth ) { return GetProperty( Core, prefetchDepth, Depth ); }
 
   /// RevOpts: set the argv array
   void SetArgs( const SST::Params& params );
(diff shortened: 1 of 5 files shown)


include/RevOpts.h
```diff
@@ -18,15 +18,46 @@
...
 namespace SST::RevCPU {
 
 class RevOpts {
+
+  // Queries whether type is a std::vector type
+  template<typename T>
+  struct is_vector : std::false_type {};
+
+  template<typename T>
+  struct is_vector<std::vector<T>> : std::true_type {};
```
leekillough, line 37: This is standard traits programming, where a trait query class is defined as false in general, but then a _partial specialization_ is defined for which it's true for a proper subset of the original template.

include/RevOpts.h
```diff
@@ -18,15 +18,46 @@
...
+  bool GetProperty( uint32_t Core, const MAP& Map, VAL&& Val ) const {
+    if constexpr( is_vector<MAP>::value ) {
+      // If MAP is a vector
+      return Core < numCores ? Val = Map[Core], true : false;
+    } else {
+      // If MAP is a map/unordered_map
+      auto it                      = Map.find( Core );
+      return it != Map.end() ? Val = it->second, true : false;
+    }
+  }
```
leekillough, line 51: Only one of the `if constexpr()` branches will be taken, and the other will be discarded at compile-time, and will produce no errors.

So we can write one function which operates differently depending on whether `MAP` is a vector or not.

`VAL&& Val` is a _universal reference_, meaning `VAL` will be of lvalue reference type for lvalue arguments (arguments with a name whose address can be taken) and `VAL` will be of rvalue reference type for rvalue arguments (arguments with no name or address, which represent temporary expressions which are fleeting and whose resources can be moved instead of deep-copied). Inside `GetProperty`, `Val` is an lvalue even if it bound to an rvalue. `std::tie()` creates an rvalue tuple of references, and can be assigned to even though it is an rvalue, in which case it assigns to its lvalue reference elements.

include/RevOpts.h
```diff
@@ -59,22 +90,24 @@ class RevOpts {
   bool InitPrefetchDepth( const std::vector<std::string>& Depths );
 
   /// RevOpts: retrieve the start address for the target core
-  bool GetStartAddr( uint32_t Core, uint64_t& StartAddr );
+  bool GetStartAddr( uint32_t Core, uint64_t& StartAddr ) const { return GetProperty( Core, startAddr, StartAddr ); }
```
leekillough, line 101: All of the `Get*` functions are now implemented in terms of `GetProperty()`.

include/RevOpts.h
```diff
@@ -59,22 +90,24 @@ class RevOpts {
...
+  bool GetMachineModel( uint32_t Core, std::string& MachModel ) const { return GetProperty( Core, machine, MachModel ); }
 
   /// RevOpts: retrieve instruction table for the target core
-  bool GetInstTable( uint32_t Core, std::string& Table );
+  bool GetInstTable( uint32_t Core, std::string& Table ) const { return GetProperty( Core, table, Table ); }
 
   /// RevOpts: retrieve the memory cost range for the target core
-  bool GetMemCost( uint32_t Core, uint32_t& Min, uint32_t& Max );
+  bool GetMemCost( uint32_t Core, uint32_t& Min, uint32_t& Max ) const {
+    return GetProperty( Core, memCosts, std::tie( Min, Max ) );
```
leekillough, line 114: `std::tie( Min, Max )` is passed to `GetProperty( )`, whose universal reference `VAL` will be deduced to be an rvalue reference. But it can still be assigned to, and the assignment to `Val` will extract elements of `std::pair` or `std::tuple` into `Min` and `Max`.

include/RevOpts.h
```diff
@@ -91,18 +124,29 @@ class RevOpts {
...
+  // -- prefetch depth = 16
+  // -- pipeLine = 5 ???
+
+  // clang-format off
+  std::vector<uint64_t>                     startAddr{decltype( startAddr     )( numCores, 0 )};                 ///< RevOpts: starting address
+  std::vector<std::string>                    machine{decltype( machine       )( numCores, "G" )};               ///< RevOpts: machine model
+  std::vector<std::string>                      table{decltype( table         )( numCores, "_REV_INTERNAL_" )};  ///< RevOpts: inst table
+  std::vector<std::pair<uint32_t, uint32_t>> memCosts{decltype( memCosts      )( numCores, {0, 10} )};           ///< RevOpts: memory cost range
+  std::vector<uint32_t>                 prefetchDepth{decltype( prefetchDepth )( numCores, 16 )};                ///< RevOpts: prefretch depth
+  // clang-format on
```
leekillough, line 153: All of the vectors which are initialized to default values are initialized at constructor time, after `numCores`, declared above them, is initialized. The `decltype()` syntax is syntactic sugar for not having to repeat the type in the far-left column a second time, which we need in order to make direct constructor calls to `std::vector` rather than initialization-list calls to `std::vector`, which have a different meaning.

include/RevOpts.h
```diff
@@ -91,18 +124,29 @@ class RevOpts {
...
+
+  std::unordered_map<uint32_t, std::string> startSym;       ///< RevOpts: starting symbol
```
leekillough, line 155: Because of how `RevLoader` works, `startSym` has to remain a `std::unordered_map` instead of a `std::vector`. Here it is default-initialized as being empty.

src/RevOpts.cc
```diff
@@ -12,29 +12,6 @@
...
-  // -- prefetch depth = 16
-  for( uint32_t i = 0; i < numCores; i++ ) {
-    startAddr.insert( std::pair<uint32_t, uint64_t>( i, 0 ) );
-    machine.insert( std::pair<uint32_t, std::string>( i, "G" ) );
-    table.insert( std::pair<uint32_t, std::string>( i, "_REV_INTERNAL_" ) );
-    memCosts.push_back( InitialPair );
-    prefetchDepth.insert( std::pair<uint32_t, uint32_t>( i, 16 ) );
-  }
-}
-
```
leekillough, line 37: All of that loop has been moved to the constructor initialization syntax above, where `std::vector(numProcs, value)` initializes the vector to length `numProcs` copies of `value`.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
   }
 }
 
-bool RevOpts::InitPrefetchDepth( const std::vector<std::string>& Depths ) {
-  std::vector<std::string> vstr;
-  for( uint32_t i = 0; i < Depths.size(); i++ ) {
-    std::string s = Depths[i];
+template<typename MAP>
+bool RevOpts::InitPropertyMap( const std::vector<std::string>& Opts, MAP& map ) {
+  for( auto& s : Opts ) {
```
leekillough: Range-based-for is generally preferred to old C-style for loops when it works.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
-    std::string::size_type sz          = 0;
-    uint32_t               Depth       = std::stoul( vstr[1], &sz, 0 );
+    // Store as cast integer if target is integer; otherwise store as string
+    auto parse = [&]( auto val ) {
+      if constexpr( std::is_integral_v<decltype( val )> ) {
+        map[Core] = decltype( val )( std::stoull( vstr[1], nullptr, 0 ) );
+      } else {
+        map[Core] = vstr[1];
+      }
+    };
```
leekillough, line 51: This lambda represents an anonymous function. The `[&]` makes it capture any expressions it uses in its parent by reference. The `auto val` is an argument which is only used in `decltype( val )` but which is a way of making this lambda a "generic lambda" which is like a template under-the-hood. The `if constexpr` will only execute one of its branches, for integral and non-integral vector/map elements.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
 
-    prefetchDepth.find( Core )->second = Depth;
-    vstr.clear();
+    if constexpr( is_vector<MAP>::value ) {
+      parse( typename MAP::value_type{} );
+    } else {
+      parse( typename MAP::mapped_type{} );
+    }
```
leekillough, line 57: Here the `parse` lambda defined above is called one of two ways based on whether `MAP` is a vector type or not. If it's a vector, then its `value_type` member defines the vector element type, and we pass a default `{}` (or zero) value of it. If it's not a vector, we assume it's a `map` / `unordered_map` and get the value type from its `mapped_type` member and pass a similar value to the lambda.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
-  if( StartAddrs.size() == 1 ) {
-    std::string s = StartAddrs[0];
+  if( Opts.size() == 1 ) {
+    auto&                    s = Opts[0];
+    std::vector<std::string> vstr;
+
     splitStr( s, ":", vstr );
     if( vstr.size() != 2 )
-      return false;
+      return { true, false };
```
leekillough, line 70: Here we indicate that the caller should return immediately (`true`) with a value of `false` because an error has occurred.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
+          for( uint32_t i = 0; i < numCores; i++ )
+            map[i] = vstr[1];
+        }
+      };
+
+      if constexpr( is_vector<MAP>::value ) {
+        parse( typename MAP::value_type{} );
+      } else {
+        parse( typename MAP::mapped_type{} );
       }
```
leekillough, line 91: Here we have the similar lambda / template / `if constexpr` trickery, generating the cross product of `{map, vector}` x `{ integral element type, non-integral element type}` with as little code as possible.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
-      return true;
+
+      return { true, true };
```
leekillough, line 92: We succeeded with `CORES:` so we ask the caller to immediately (`true`) return `true`.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
     }
   }
+  return { false, false };
```
leekillough, line 95: We did not succeed with `CORES:` but did not encounter any errors, so we ask the caller not to return immediately (`false`), and to fall back on `InitPropertyMap()`.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
-    uint64_t               Addr    = std::stoull( vstr[1], &sz, 0 );
-
-    startAddr.find( Core )->second = Addr;
-    vstr.clear();
-  }
-  return true;
+/// RevOpts: initialize the set of starting addresses
+bool RevOpts::InitStartAddrs( const std::vector<std::string>& StartAddrs ) {
+  auto [ret, retval] = InitPropertyMapCores( StartAddrs, startAddr );
+  return ret ? retval : InitPropertyMap( StartAddrs, startAddr );
```
leekillough, line 101: Here the `{bool, bool}` values returned from `InitPropertyMapCores()` are extracted into `ret` and `retval`. If `ret` is `true`, we immediately return `retval`; otherwise we call `InitPropertyMap()` to handle the non-`CORES:` syntax.

src/RevOpts.cc
```diff
@@ -49,226 +26,117 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
-    uint32_t Max          = std::stoul( vstr[2], nullptr, 0 );
-    memCosts[Core].first  = Min;
-    memCosts[Core].second = Max;
-    if( ( Min == 0 ) || ( Max == 0 ) ) {
+    auto Core = std::stoull( vstr[0], nullptr, 0 );
+    auto Min  = decltype( memCosts[Core].first )( std::stoull( vstr[1], nullptr, 0 ) );
+    auto Max  = decltype( memCosts[Core].second )( std::stoull( vstr[2], nullptr, 0 ) );
+    if( Core >= numCores || !Min || !Max )
       return false;
-    }
```
leekillough, line 201: `InitMemCosts()` is different from the others because it has `{Min, Max}` pairs. But it does not need to be as general as the others, so it is fairly short code. The `decltype( memCosts[Core].first )` and `decltype( memCosts[Core].second )` are used to cast `stoull()` to the target `{Min, Max}` types, to prevent warnings in case they are smaller than `stoull()`.

src/RevOpts.cc
```diff
@@ -49,226 +26,115 @@ void RevOpts::SetArgs( const SST::Params& params ) {
...
+      if constexpr( is_vector<MAP>::value ) {
+        parse( typename MAP::value_type{} );
+      } else {
+        parse( typename MAP::mapped_type{} );
       }
+
       return true;
     }
   }
+  return InitPropertyMap( Opts, map );
```
leekillough, line 96: Here is where `InitPropertyMapCores()` falls back on `InitPropertyMap()` if no `CORES:*` is found.

kpgriesser (approved):
Thanks for the extremely helpful detailed code descriptions.  I have no specific feedback on the code itself.

PR #353, opened 2025-01-06
Compress RevFlag so that atomic operations take fewer bits

leekillough:
Since atomic instruction operations are mutually exclusive, change their `RevFlag` flags to use consecutive integers instead of consecutive powers of 2, so that fewer bits are used in `RevFlag`. They are still ORed with other `RevFlag` values, but now take up only 4 bits instead of 9.


jleidel (approved):
Good to merge.  This leaves just enough space for us

PR #354, opened 2025-01-07
Use regular expressions to match MIME types

leekillough:
Python MIME types vary across platforms, and this change lets the expected MIME type be an extended regular expression instead of just a plain string.


jleidel (approved):
I believe that should work for what we need

kpgriesser (approved):
performance quick sanity check for changing python3 script followed by commit. No issues.

PR #356, opened 2025-01-09
Add sanity checks for assumptions about instruction tables

leekillough:
This makes it a fatal error if a hash table collision occurs because two added entries have the same key, or if it's ambiguous which instruction matches an encoding because there is more than one match. This should help avoid surprises when two instructions map to the same decoding.


jleidel (approved):
LGTM

PR #357, opened 2025-01-09
RVV compile option to enable decoding vector instructions

kpgriesser:
This PR adds a compile option, RVV, to allow REV to be built to support the vector coprocessor. This PR is required for a forthcoming PR in the R2V2 repository that decouples that design from the REV source tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # RevCPU Top-Level CMake
-# Copyright (C) 2017-2024 Tactical Computing Laboratories, LLC
+# Copyright (C) 2017-2025 Tactical Computing Laboratories, LLC
 # All Rights Reserved
 # contact@tactcomplabs.com
 # See LICENSE in the top level directory for licensing details
@@ -72,6 +72,13 @@ if(REV_TRACER)
 else()
   # requires cmake 3.12
   add_compile_definitions(NO_REV_TRACER)
+endif()
+
+# Vector decoding
+option(RVV "Enable vector decoding (requires r2v2 coprocessor)" OFF)
+if (RVV)
+  message(RVV=ON)
+  add_compile_definitions("RVV")
 endif()
 
 # Compiler Options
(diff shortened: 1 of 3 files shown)


leekillough:
We should compare / consolidate this with the `REVVEC` macro definition.

```
#ifdef REVVEC
#include "insns/RVVec.h"
#endif
```

kpgriesser:
RVVec.h,. the vector instruction implementations, no longer exists in the rev source tree. The now independent coprocessor now includes this header so we can remove the #ifdef REVVEC code block. ( Tested with 'solo' branch in r2v2 repo )

leekillough:
I was mainly speaking of the `REVVEC` vs `RVV` naming discrepancy -- consolidate one and retire the other, as well as integrate them into the build. `grep REVVEC src/* include/* include/insns/*` to make sure there aren't others.

I don't think we need the `RevFeature` `V` extension parsing to be enabled/disabled by an `#ifdef`. Like `Zicntr`, a runtime error can be printed if a `V` coprocessor is not available.

kpgriesser:
There are no other REVVEC defines. If nobody has an issue with have the V extension parsing on by default ( I don't ) then we can just remove the RVV define as well and not require any special compile.

leekillough:
I would prefer if we didn't have to build Rev differently. I think the parsing of the `V` extension can always be supported. If the `V` instructions are handled by the coprocessor during decode, and the coprocessor code has access to `RevFeature` (directly or indirectly -- we can always add more accessors if the current passkey mechanism is insufficient), then it can tell whether to decode `V` instructions or just ignore them and cause a normal `unable to decode instruction` error in Rev Core. Or if it wants to be more fancy, the coprocessor can decode `V` instructions as if the `V` extension is enabled, and then abort with a runtime error if the `V` feature is not really enabled, similar to how `Zicntr` works today.

kpgriesser:
I agree.  My last push removes the RVV define. So this PR is essentially just removing REVVEC. From every other users' standpoint it is identical to the devel branch.

leekillough (approved):

PR #360, opened 2025-01-13
Adds new CI scripts for updated test system

jleidel:
Note: squashed 8 commits into a single commit + PR in order to avoid poisoning the git history.  This only **ADDS** new scripts, it does not remove the current set of scripts.  New scripts verified to be functional on `nosferatu`

adding llvm 18.1.3 ci scripts
increasing node-level test resource limits
forcing ntasks to 56
fixing test harness script line
reducing parallelism
back to sequential execution
reverting to sequential testing


leekillough (approved):
LGTM

PR #362, opened 2025-01-13
fix conversions

leekillough:
This is a one-liner fix for a conversion error which did not show up in earlier testing.

If you see errors like this again, it is probably due to compiler differences (different compiler, different compiler version, different host OS, different optimization level, etc.). My goal is to get zero errors like this.

These errors are trivial and are not showstoppers. They just indicate truncating conversions which need to be made explicit rather than implicit. Compilers raise these warnings differently depending on version and sometimes even optimization. There is no standard on which warnings are required, so every compiler behaves slightly differently.


jleidel:
do not merge this until after Feb 4.

kpgriesser (approved):

PR #363, opened 2025-01-16
remove stray characters generated by emacs xterm-extra-capabilities

leekillough:
Emacs `xterm-extra-capabilities` sometimes causes stray characters to creep into editor buffers on startup if keys are typed before `emacs` starts.

This can be avoided by adding `(setq xterm-extra-capabilities nil)` in `~/.emacs`.


jleidel (approved):
LGTM

PR #365, opened 2025-01-18
Move amo code

leekillough:
Post-CDR; This moves the atomic template code around so that it is out of header files and so that its template functions can be instantiated as many times as necessary from the same `.cc` file. It makes it so that RevMem and MemH can both call `RevBasicMemCtrl::performAMO()` to perform atomic operations. The changes needed to remove Forza atomics are minimized. This includes all of the changes in https://github.com/tactcomplabs/rev/pull/364 plus these moves of the AMO code.


leekillough:
This rewrites the MemH memory hierarchy implementation (this PR is mirrored in F****Rev).

- There is no duplicated code anywhere now. If one thing needs to be done in two or more ways, then a function, template, macro, parameterization, table, or state-changing loop is used so that code remains in one place only.
  - For example, previously there were 5 separate functions for `buildStandardMemRqst()`: 
    - `buildStandardMemRqst()`
    - `buildRawMemRqst()`
    - `buildCacheMemRqst()`
    - `getNumCacheLines()`
    - `getBaseCacheLineSize()`
  - `buildRawMemRqst()` and `buildCacheMemRqst()` duplicated the same `switch` statements, and `buildCacheMemRqst()` repeated the same `switch` statement twice for operating on the beginning and end of a memory block.
  - Now there is **only** `buildStandardMemRqst()`, which performs the operations of both cached and uncached memory with a **single** `switch` statement in a **single** loop, and whose complicated functions `getNumCacheLines()`, and `getBaseCacheLineSize()` have been replaced with **single lines of code** using the correct mathematical formulas.
- The sending of memory requests during a `ClockTick()` has been made correct, by using `RevBasicMemCtrl` member variables to track outstanding memory requests as they are started and completed, rather than local variables in `RevBasicMemCtrl:::ClockTick()` which did not correctly reflect the state of outstanding requests, but only tracked counts while traversing a queue of pending requests. (These memory request counts may need to be made Hart-specific or Core-specific if that's the intention; they are only used to enforce outstanding request limits, and do not affect functionality.)
- A `sendMemRqst()` function has been added as a single point where memory requests can be tracked and sent to SST. What used to take 4-6 lines of code to send a `StandardMem` request, **now only takes one line**.
- The **separate** outstanding memory request counter variables have been replaced with a **single** `class MemOpParams` which is indexable by `MemOp`.
- Linear search algorithms have been replaced with `O(1)` or `O(lg n)` algorithms.
  - The `AMOTable` has been eliminated (it was linearly searched), as has the `requests` vector (which was largely unused). `AMOTable` was redundant because there are faster, simpler ways to detect outstanding atomic operations.
  - The `outstanding` hash table maps from a  `StandardMem::Request` id to a `pair`: a smart pointer to the `RevMemOp` and an iterator to a `std::multimap` node mapping Hart ids to `RevMemOp`. In constant time, a completed request can be looked up and then its `RevMemOp` handled. In logarithmic time, a completed request can be removed from the mapping from Harts to `RevMemOp`. In the future, this functionality can perhaps be combined with or replace the `LSQueue` logic.
- `RevFlag` and related functions has been moved out into its own `RevFlag.h` header file.
- `RevFlag` enum values have been defined in Rev for multiple architectures, to reduce merge conflicts and to reduce the chances of running out of flag values on other architectures.
- There is no need to concern ourselves with "split requests" or "split request counts" and provide separate functions to get/set these counts, or use them for our own memory management (such as calling `delete` after a specified number of "split requests" has been seen). Only a reference counter in the `RevMemOp` is needed, which counts the number of outstanding memory requests referring to it. Each instruction's memory operation will have a unique `RevMemOp` even if multiple requests are used for it. If a memory request is split across cache lines, then multiple requests will be sent, and each one will increment the reference count of the `RevMemOp`. When the requests finish, they will decrease the reference count and only when it reaches `0` will the whole memory operation need to be finished and marked completed.
- `std::shared_ptr<RevMemOp>` handles automatic memory management of `RevMemOp`, which can have many shared pointers to it in many places. This `std::shared_ptr<RevMemOp>` reference count is independent of the count of outstanding memory requests to a `RevMemOp`. There are places where `std::shared_ptr<RevMemOp>` might have more references to a `RevMemOp` than actual memory requests.
- Memory hazards is handled based on actual outstanding requests which have been sent, not the `rqstQ` queue of pending memory requests. The original code only took into account pending memory requests in the `rqstQ` when deciding whether a memory request was ready to be sent or not. Now it uses true memory ordering semantics based on the actual state of memory requests in progress:
   - If a memory request has Release Semantics (`RevFlag::F_RL`), then it won't be sent if there are any actual outstanding memory operations on the same hart in effect.
   - If any outstanding memory request has Acquire Semantics (`RevFlag::F_AQ`), another memory request won't be sent from the same hart.
   - If any outstanding memory request has Atomic flags set and is a Read, another memory request won't be sent from the same hart (but if the outstanding request is the Write of an atomic operation, another memory request may be sent).
- `handleResp()` handles all memory responses now regardless of type. It is templated and has replaced 5 functions. Compile-time conditional code is enabled with `if constexpr(...)` on template parameters.
- `MarkLoadComplete()` is called after a memory Read completes, even if the Read is part of an atomic operation. The corresponding Write does not need to complete before marking the destination register of the atomic operation as complete. The destination register, and the new value to be written to memory, will have completed by the end of the Read, and so it can be marked complete and register hazards cleared.
- When an atomic Read completes, an atomic Write is immediately sent -- it does not go through the rqstQ. This allows the Read/Load to be marked Complete and the register hazard released immediately, before the Write completes, and it allows other memory requests in the same hart to be immediately sent after the Read completes and before the Write completes, as long as the atomic operation does not have Acquire semantics and the following memory operation does not have Release semantics.
- The memory FENCE is handled more logically now. Instead of being tested at the beginning of `ClockTick()`, it is tested at the beginning of `processNextRqst()`, and the actual state of `outstanding` memory requests is tested. If there are no outstanding memory requests, the FENCE flag is cleared, and `processNextRqst()` can immediately consider memory operations on the same clock. If a FENCE operation is encountered in the `rqstQ`, the FENCE flag is set and `ClockTick()` immediately cycles back during the same clock and tests outstanding requests, and only stalls if there are any.
- `buildStandardMemRqst()` returns `true` now if it has enough slots for a memory operation, and `false` if it doesn't (previously this was handled by a `Success` variable passed in by reference). Previously the success of a memory operation was tested incorrectly based on badly-kept counters instead of a coherent view of `outstanding` requests. If a request requires so many cache lines that it always exceeds the maximum number of requests allowed at a time, a fatal error occurs.
- Virtual methods which were not used in `RevMemCtrl` but only used in `RevBasicMemCtrl`, and weren't expected to be used in the future, were removed from the `RevMemCtrl` abstract base class.
- Loops to construct `std::vector<uint8_t>` temporaries from other vectors or strings for memory data were replaced with iterator ranges, such as `{ curByte, curByte + size }` to create a `StandardMem::Write`'s input data from a pair of iterators which construct a vector argument.  Previously the code created a temporary `std::vector<uint8_t>` and performed a loop to `push_back` each character into the temporary vector.
- Atomic code has been broken up into separate functions so as to maximize reuse of it across RevMem, MemH, and F****Rev. For example, `performAMO()` works on both `RevMem` and `MemH`, and `ApplyAMO()` can be specialized for different architectures.
- The code in `RevMemCtrl.cc` is 60% (918 lines) smaller.
- The code in `RevMemCtrl.h` is 26% (191 lines) smaller.

This is the biggest rewrite since I templatized instruction execution code, but fortunately, nearly all of the changes are just in `RevMemCtrl.h` and `RevMemCtrl.cc`.

The new code should make adding memory features much easier.

kpgriesser:
Confirmed compile and ctest clean for MacOS.
However, I'm seeing the following issue on ubuntu on the install step

```
$ make install
[100%] Built target revcpu
Install the project...
-- Install configuration: ""
-- Up-to-date: /home/ken/work/rev/src/librevcpu.so
'/home/ken/work/rev/src/../scripts/test_undefined_symbols.sh' '/home/ken/work/rev/src/librevcpu.so'

undefined symbol: bool SST::RevCPU::RevBasicMemCtrl::QRequest<bool&>(SST::RevCPU::RevBasicMemCtrl::MemCtrlStats, SST::RevCPU::MemOp, unsigned int, unsigned long, unsigned long, unsigned int, SST::RevCPU::RevFlag, bool&)	(/home/ken/work/rev/src/librevcpu.so)

CMake Error at src/cmake_install.cmake:82 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 1"

Call Stack (most recent call first):
  cmake_install.cmake:47 (include)


make: *** [Makefile:110: install] Error 1
```

leekillough:
That means that a template instantiation is not defined.

The common idea that you can only define C++ templates in header files is not 100% true, because they can be explicitly instantiated and defined in compilation units (`.cc` files) which do not otherwise use them.

This error means that a template instantiation is referenced but not defined. There are many ways to fix it, such as making an explicit instantiation of it in the `RevMemCtri.h` or `RevMemCtrl.cc` file, or by moving the definition of the template from `RevMemCtrl.cc` to `RevMemCtrl.h`.

The problem is that the template instantiation is made in a file which includes `RevMemCtrl.h` but not `RevMemCtrl.cc`. If a template has been instantiated with certain template arguments when the template definition occurs, such as in `RevMemCtrl.cc`, then the template definition with those arguments will be defined. If only the declaration of the template exists at template instantiation time, another instantiation of the template with those arguments will be necessary when the template definition is reached, in order for the template with those arguments to be defined.

leekillough:
@kpgriesser : The problem is that `RevBasicMemCtrl::sendFLUSHRequest()` is only called from `RevMem.h`.

`RevBasicMemCtrl::sendFLUSHRequest()` calls `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )` but since `RevBasicMemCtrl::sendFLUSHRequest()` is implicitly `inline`, and since it is not called from `RevMemCtrl.h` or `RevMemCtrl.cc`, it can be optimized away from `RevMemCtrl.cc`'s compilation, and thus the implicit instantiation of `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )` does not occur, and `RevMemCtrl.cc` does not define that instantiation of `RevBasicMemCtrl::QRequest()`, leading to the link-time error. When `RevMem.h` calls `RevBasicMemCtrl::sendFLUSHRequest()`, that calls `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )`, but since `RevMem.h` and its clients do not define the `RevBasicMemCtrl::QRequest()` template. that instantiation's definition does not occur even though a reference to it occurs.

Solutions:
1. Define `RevBasicMemCtrl::QRequest()` in `RevMemCtrl.h`, so that when `RevBasicMemCtrl::sendFLUSHRequest()`  is called as an `inline` method,  `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )`  gets called and implicitly instantiated, and the definition in `RevMemCtrl.h` is able to define that instantiation of `RevBasicMemCtrl::QRequest( )`. (This is the solution people mean when they say "You can only define templates in headers", but that's not true.)
2. Explicitly instantiate `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )`  (and all other variants) in either `RevMemCtrl.h` or `RevMemCtrl.cc`. This explicit instantiation cannot be ignored like the implicit instantiation caused by an `inline` function which never gets called can be ignored, and so the explicit instantiation will cause `RevMemCtrl.cc` to define `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )`.
3. Make `RevBasicMemCtrl::sendFLUSHRequest()` non-`inline` by defining it in `RevMemCtrl.cc` outside of the `RevBasicMemCtrl` class. Its non-inline definition will call `RevBasicMemCtrl::QRequest( MemCtrlStats::FlushPending, MemOp::MemOpFLUSH, Hart, Addr, PAddr, Size, flags, Inv )`  and will cause that instantiation, which `RevMemCtrl.cc` will define.
4. Some combination of the above. I'm considering different combinations and their trade-offs.

kpgriesser:
The latest version now compiles clean and passes testing on macOS and Ubuntu. As always, thanks for the detailed explanations as it really helps me appreciate and improve template programming.

src/RevSysCalls.cc
```diff
@@ -3435,7 +3435,7 @@ EcallStatus RevCore::ECALL_fast_printf() {
 /* System Call (ecall) Implementations Below */
 /* ========================================= */
 // clang-format off
-const std::unordered_map<uint32_t, EcallStatus(RevCore::*)()> RevCore::Ecalls = {
+decltype(RevCore::Ecalls) RevCore::Ecalls = {
```
kpgriesser, line 3438: Could the return type be just 'auto' here?

kpgriesser (approved):
I've tested this for Mac and ubuntu using it as REVHOME for r2v2 and everything appears to be passing. But it's really not testing the memh or amo code other than what's in the rev repo already.  Just one comment/question on the code.

jleidel (approved):
LGTM

src/RevSysCalls.cc
leekillough: It is a class member _variable_ definition rather than a class member _function_ definition.  I cannot use `auto` here. I could only use `auto` on the member variable if it was initialized _inside of the class_ (and was `static`).

I discovered that `decltype(classType::classMember) classType::classMember = {...};` was a shorthand way of declaring a member's type outside of a class without having to repeat its long type declaration (such as `const std::unordered_map<uint32_t, EcallStatus(RevCore::*)()>` in this example). 

(`auto` can be used with member functions defined _outside of the class_, but then it cannot be called except in the same `.cc` file as the function's out-of-class definition, which allows the `auto` to be deduced. Other `.cc` files which `#include` the `.h` file and get the function's `auto` declaration inside the class but no function definition inside the class cannot call the function, since its return type is unknown.)

kpgriesser: I see. So ecalls is declared in RevCore.h as
`static const std::unordered_map<uint32_t, EcallStatus ( RevCore::* )()> Ecalls;`
I was thinking it was somehow inferring the type from the body.
It all makes sense now - thank you

rkabrick (approved):
No issues on my end 👍

PR #367, opened 2025-01-27
Fix lambda captures could have had dangling references

leekillough:
This fixes the atomics code which captures-by-reference. 

The current code captures variables by reference  `[&]` and stores the resulting functions in a `static` table. And although these references may continue to point to the same stack offsets where the captured variables were at, and may work on subsequent calls to the function which reuse the `static` variable, this is not guaranteed. To be correct, the `static` `table`  would need to be non-`static` and reinitialized every time the function is called, which is wasteful since we are only interested in one of the alternatives at each call.

The new code is also simpler. Historically, it had been much more complicated, which made lambdas at the time seem like a simplification, but  now a simple `switch` statement is simpler.

### DIFF
--- a/include/RevMemCtrl.h
+++ b/include/RevMemCtrl.h
@@ -713,24 +713,19 @@ void ApplyAMO( RevFlag flags, void* Target, T value ) {
 
   // Table mapping atomic operations to executable code
   // clang-format off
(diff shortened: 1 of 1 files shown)


kpgriesser (approved):
definitely simpler and more readable.

PR #369, opened 2025-04-24
Add ecall open support

ct-clmsn:
initial import of ecall support for code 1024 (system call `open`) from [here](https://man7.org/linux/man-pages/man2/open.2.html)


README.md
```diff
@@ -2,6 +2,8 @@
 
 ![rev](documentation/imgs/rev_logo.png)
 
+## NOTICE: REV HAS MIGRATED THE PRIMARY BRANCH FROM `MASTER` TO `MAIN` ON FEBRUARY 05, 2024
+
```
rkabrick: Not sure this should be included in this PR

include/RevCore.h
```diff
@@ -424,6 +424,7 @@ class RevCore {
   EcallStatus ECALL_fchownat();               // 54, rev_fchownat(int dfd, const char  *filename, uid_t user, gid_t group, int flag)
   EcallStatus ECALL_fchown();                 // 55, rev_fchown(unsigned int fd, uid_t user, gid_t group)
   EcallStatus ECALL_openat();                 // 56, rev_openat(int dfd, const char  *filename, int flags, umode_t mode)
+  EcallStatus ECALL_open();                   // 1024, rev_open(const char  *filename, int flags)
```
rkabrick: I think these are in order based on the ecall code so I would move this down to probably here? 

image

Maybe @leekillough can decide where it makes more sense. I arbitrarily added 1000 as the pthread functions way back so unfortunately this one having the code 1024 doesn't fit in nicely with the ordering. Really just a nit more than anything

src/RevSysCalls.cc
```diff
@@ -620,6 +620,43 @@ EcallStatus RevCore::ECALL_openat() {
...
+EcallStatus RevCore::ECALL_open() {
+  auto& EcallState = Harts.at( HartToExecID )->GetEcallState();
+  if( EcallState.bytesRead == 0 ) {
+    output->verbose(
+      CALL_INFO, 2, 0, "ECALL: openat called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
+    );
+  }
+  auto pathname = RegFile->GetX<uint64_t>( RevReg::a0 );
+
+  // commented out to remove warnings
```
rkabrick: Not sure if the comments in here are related to this function or an artifact of copy/paste. If they are related to `open` then keep em, otherwise probably best to ditch em.

rkabrick (changes requested):
Just take a look at my comments and then add a test or two in the `test/syscalls` directory

include/RevCore.h
ct-clmsn: The support was a modified copy+paste of the `openat` implementation. Wasn't sure if keeping similar implementations closer together makes sense or to retain the current numerical ordering. Either way, small enough fix.

src/RevSysCalls.cc
ct-clmsn: Relic taken from `openat`. Will remove.

include/RevCore.h
rkabrick: Yeah I could be convinced either way. Will leave it to @leekillough to make the call

leekillough: I would move it so that it is in numerical order with other system calls.

Is the value `1024` special? I thought most system calls were < `512` and that we added "custom" system calls >= `1024` whose number was not tied to any kernel system call but was our own "pseudo-system-call".

Also a number like `1024`, being right on a power-of-2 boundary, might conflict with other system call numbers, so if we are choosing the value to be something special, we should choose a larger value and put it in the next multiple-of-100 after what we have already, or something like that.

ct-clmsn: Cool, I'll move it soon!

Yep! `1024` is the risc-v kernel syscall table identifier for `open`. I've some kernels that break when the `open` system call is made and the assembly has the ecall code set to `1024`.

I'll try to dig up the value in the kernel; this link is what I could dig up quickly.

https://github.com/westerndigitalcorporation/RISC-V-Linux/blob/master/riscv-pk/pk/syscall.h

src/RevSysCalls.cc
ct-clmsn: comment removed!

README.md
ct-clmsn: comment removed!

ct-clmsn:
@rkabrick added a test case

include/RevCore.h
ct-clmsn: @leekillough looks like I'll have to do a cross compile to risc-v assembly and double check. the gcc risc-v backend doesn't define a system call number for `open`; getting the impression `open` is implemented _over_
`openat`.

https://github.com/riscv-collab/riscv-gnu-toolchain/blob/master/linux-headers/include/asm-generic/unistd.h

common/syscalls/syscalls.h
```diff
@@ -656,6 +656,7 @@ REV_SYSCALL(  440, int rev_process_madvise(int pidfd, const struct iovec *vec, s
 
 REV_SYSCALL(  500, int rev_cpuinfo(struct rev_cpuinfo *info) );
 REV_SYSCALL(  501, int rev_perf_stats(struct rev_stats *stats) );
+REV_SYSCALL(  1024, int rev_open(const char *filename, int flags ) );
```
leekillough, line 659: Might want to keep it in numerical order, but with a blank line and a comment line explaining what it is for, so that it doesn't appear to just be another entry in the long list above it.

ct-clmsn:
PR is functional and seems to be working well for test suite

src/RevSysCalls.cc
```diff
@@ -620,6 +620,36 @@ EcallStatus RevCore::ECALL_openat() {
...
+    );
+  }
+  auto pathname = RegFile->GetX<uint64_t>( RevReg::a0 );
+
+  auto flags = RegFile->GetX<int>(RevReg::a1);
+
+  /* Read the filename from memory one character at a time until we find '\0' */
+
+  auto action   = [&] {
+    // Do the openat on the host
```
leekillough: Might want to change `openat` to `open` in the comments

src/RevSysCalls.cc
```diff
@@ -620,6 +620,36 @@ EcallStatus RevCore::ECALL_openat() {
...
+    std::string const full_path =
+      std::filesystem::current_path().append(EcallState.string).string();
+    auto fd  = open( full_path.c_str(), flags);
+
+    // Add the file descriptor to this thread
+    Harts.at( HartToExecID )->Thread->AddFD( fd );
```
leekillough: Might want to guard against `fd == -1` which indicates error.

ct-clmsn:
@leekillough thanks for the review. Made those 2 corrections and pushed and update.

src/RevSysCalls.cc
```diff
@@ -620,6 +620,50 @@ EcallStatus RevCore::ECALL_openat() {
...
+       output->fatal(
+          CALL_INFO,
+          -1,
+          "Core %" PRIu32 "; Hart %" PRIu32 "; Thread %" PRIu32 " tried to close file descriptor %" PRIu32
+          " but did not have access to it\n",
+          id,
+          HartToExecID,
+          ActiveThreadID,
+          fd
+       );
```
leekillough: I would not make it a fatal error message, but simply not call `AddFD()` if `fd == -1`:
```c++
if (fd != -1) Harts.at( HartToExecID )->Thread->AddFD( fd );
```
It will return `-1` to the caller by setting `a0` to `fd`, error or not, before returning.

It is up to the caller what to do in case of error.

Also, your error message says "tried to **close** file descriptor 4294967295 but did not have access to it". The file descriptor should not be printed if it's `-1`, should not be printed as unsigned, and the reason for the failure varies (it is in `errno`).

src/RevSysCalls.cc
```diff
@@ -620,6 +620,43 @@ EcallStatus RevCore::ECALL_openat() {
...
+      std::filesystem::current_path().append(EcallState.string).string();
+    auto fd  = open( full_path.c_str(), flags);
+
+    if(fd != -1) {
+      // Add the file descriptor to this thread
+      Harts.at( HartToExecID )->Thread->AddFD( fd );
+    }
+    else {
+      output->verbose(
+        CALL_INFO, 2, 0, "ECALL: open called by thread %" PRIu32 " on hart %" PRIu32 "; file descriptor invalid (fd == -1)\n", ActiveThreadID, HartToExecID
```
leekillough: I would use:

`"ECALL: open called by thread %" PRIu32 " on hart %" PRIu32 " returned -1: %m\n"`

 Nothing about "file descriptor invalid". `%m` will print the cause of the error, such as `No such file or directory`.

kpgriesser:
I've got a more rigorous test based on the newlib documentation for fopen. It's kicking out some new issues that I'm going to work on resolving.

kpgriesser (approved):
Go ahead and rebase/merge. The functionality is at the same level as openat() so good for now. I'm working on changes related to issues I'm seeing with my testing in a separate branch.

kpgriesser:
covered by #373

PR #373, opened 2025-06-10
Enhance ECALLS for open and lseek, testing, and methodology improvements

kpgriesser:
Depends on #369 ( or we could use this one in it place as it includes the same functionality + more )

The primary goal of this PR was to see how far I could enhance the `open` system call and still work on MacOS and Ubuntu host platforms. My conclusion is that system calls that support flags (e.g. open and lseek) but do not requiring writing to complex structs (e.g. stat) are viable candidates for support.   

A secondary goal was to provide a consistent way of running system call tests on the host platform or on REV. This makes it much easier to develop more sophisticated tests quickly and debug them on REV using the host simulation as a reference. These tests also use the REV fast-print system call to efficiently provide data to the console which can be pretty-printed using <rev>/scripts/rev-print.py.  See test/syscalls/open/open.c

Specifically:
- Improves `open` system call to support O_CREAT, O_RDONLY, O_WRONLY, O_APPEND, O_TRUNC for multiple platforms. In order to do this a flag translation function to convert those passed by the RV application to the host version in the ECALL. There will likely be problems on other untested platforms.
- Changed rev_llseek, which was doing nothing, to rev_lseek with function prototype matching that in // https://man7.org/linux/man-pages/man2/lseek.2.html. This initial implementation supports SEEK_SET, SEEK_CUR, SEEK_END. These appear to be more consistent across platforms.
- Added a more rigourous test for `open` to test the varios flags and correct file i/o operations.
- Created makefiles for system calls to enable building and running the test on REV or the host system with provisions to add spike target. These do not affect ctest which only uses the REV compile recipe.  For an example running locally in `rev/test/syscalls/open`
``` make clean run TARG=host```
``` make clean run```
``` make clean run > rev.log; ../../../scripts/rev-print -l rev.log```
- Added OPEN test to CMakeLists.txt
- Changed the test for PR #369, file_open.c, to also specify the 'mode' parameter which is only used when O_CREAT is used. However, the current ECALL macro does not compile with default parameter values. Perhaps I could not use the macro but thought that better left to Lee. 
- Possible issue: open takes an optional parameter, mode, but was unable to specify this in the current ECALL macro.
- No changes were made to `open_at` as I want to see if everyone agrees with this approach. For cross platform testing we may need to add checks and exclude the system call tests that are incompatible.

### CODE REVIEW

**Comment by leekillough** [comment]
A comment about the "optional" argument. In C (not sure of _very_ latest C, but at least up to C11), functions cannot have "default" arguments, so in these system calls, the "mode" parameter is used in the `open` function signature all of the time, but it is ignored when file creation is not involved. So you should model the system calls with pure C APIs which do not have default arguments, and just pass `0` for the mode if it is not used. Some declarations use varargs (`...`) but that gets complicated and usually involves passing arguments on the stack.

ECALLS on RISC-V take the system call number in `a7` and arguments in `a0-a6`. I don't think any system calls we care about need more than 7 arguments, so we don't need to deal with arguments being passed on the stack.

There is a `naked` function which implements the ECALLS:
```c++
define REV_SYSCALL(NUM, PROTO) __attribute__((naked)) \
  static PROTO{ asm(" li a7," #NUM "; ecall; ret"); }
```
The `naked` attribute makes the compiler avoid messing with the stack, saving registers, etc., and just emits the assembly code as we've written it, in this case a `li a7` which loads `a7` with the system call number, an `ECALL` which performs the system call using exactly the same `a0-a6` which were passed to the C/C++ function and the `a7` we've set, and then a `ret` returns from the function. The `ECALL` returns its return value in `a0`, which is the return value for the function.

The `PROTO` is the prototype of the function, like `open( const char* path, int flags, mode_t mode)`. The macro is called with this prototype.
 
I'm curious what error you were getting when you tried to use default arguments, which is a C++ feature and why you need it. We use `extern "C" {}` around the code in `common/syscalls/syscalls.h` when C++ is detected.

src/RevSysCalls.cc
```diff
@@ -3764,4 +3841,43 @@ decltype(RevCore::Ecalls) RevCore::Ecalls = {
...
+  // O_APPEND        0x8    0x8    0x400*
+
+  // Broadly assuming compatability here
+  if( O_CREAT == 0x200 )
+    return flags;
+
+  // mask for source bits that will be moved
+  int rvmask       = O_CREAT | O_EXCL | O_TRUNC | O_APPEND;
+  // mask for all supported bits
+  int support_mask = rvmask | 0x3;
```
leekillough: I had to add `[[maybe_unused]]` to `support_mask`:

```Diff
diff --git a/src/RevSysCalls.cc b/src/RevSysCalls.cc
index 2d776a5af..88c09f860 100644
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -3860,7 +3860,7 @@ int RevCore::hostOFlags( int flags ) {
   // mask for source bits that will be moved
   int rvmask       = O_CREAT | O_EXCL | O_TRUNC | O_APPEND;
   // mask for all supported bits
-  int support_mask = rvmask | 0x3;
+  [[maybe_unused]] int support_mask = rvmask | 0x3;
   // Fail if any unsupported bits are set
   // TODO print warning and allow system call to complete?
   assert( ( flags & ~support_mask ) == 0 );
```
Otherwise release builds get the error:
```
In file included from /home/lkillough/rev/src/RevCore.cc:12:
/home/lkillough/rev/src/RevSysCalls.cc: In member function ‘int SST::RevCPU::RevCore::hostOFlags(int)’:
/home/lkillough/rev/src/RevSysCalls.cc:3863:7: error: unused variable ‘support_mask’ [-Werror=unused-variable]
 3863 |   int support_mask = rvmask | 0x3;
      |       ^~~~~~~~~~~~
```

kpgriesser: I removed the `support_mask` in the latest push. I had added it for debug and meant to remove it.

kpgriesser: Regarding the REV_SYSCALL prototype, if mode is always required then there is no problem. I'm using this as a reference: 
`https://www.man7.org/linux/man-pages/man2/open.2.html`
and could be misinterpreting the way the function is specified.
`int open(const char *pathname, int flags, ...
                  /* mode_t mode */ );`

The reason I mentioned this was that the test for PR #369 originally did not pass the mode in so I modified the test to compile with this PR. I assumed it was a default  parameter but if it's required then that is a better solution.

leekillough: >  I assumed it was a default parameter but if it's required then that is a better solution

In C it's "unspecified" sorta, which is why the manage lists it like that, because in most ABIs, you can leave arguments off at the end of a function call and it doesn't change how the preceding arguments get passed, so if `mode` is not needed unless `O_CREAT` is turned on, you can leave `mode` off and it won't mess up the call if it doesn't need `mode`, but it definitely goes against strict prototyping in ANSI C and C++.

I'd suggest leaving `mode` in the prototype to make the function signature match what it is when `mode` is actually needed, and just pass `0` if it's not going to be used. Since Syscalls are supposed to be callable from C or C++ and have C linkage (no overloading, name mangling or namespaces), they don't use C++ features like default arguments.

Summary: `open(const char* name, int flags, mode_t mode)` is the full signature of the `open()` function and should be used when declaring/defining it, but in most ABIs if you ignore prototypes and call `open()` with only 2 arguments when `mode` is not used, it will still work. Pre-ANSI C, this was how it worked:
```c
int open(name, flags, mode)
char *name;
int flags;
mode_t mode;
{
}

int main() {
  open("file", O_RDONLY);   // Perfectly fine!!!  No prototype and open() doesn't care about mode so I don't pass it!!!
}
```

kpgriesser: For the test, `open.c`, I removed the mode for calls that do not require it and the test compiles and runs cleanly.  So no issue with the function prototype after all. Thanks for the very helpful explanations.

common/syscalls/syscalls.h
```diff
@@ -666,6 +670,18 @@ typedef unsigned long int rev_pthread_t;
 REV_SYSCALL( 1000, int rev_pthread_create( rev_pthread_t* thread, void* attr, void* fn, void* arg ) );
 REV_SYSCALL( 1001, int rev_pthread_join( rev_pthread_t thread ) );
 
+// ===================== RV system calls beyond 1024
```
rkabrick: Do we want all the commented ones pushed in?

src/RevSysCalls.cc
```diff
@@ -620,6 +622,55 @@ EcallStatus RevCore::ECALL_openat() {
...
+      CALL_INFO, 2, 0, "ECALL: open called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
+    );
+  }
+  auto pathname = RegFile->GetX<uint64_t>( RevReg::a0 );
+  auto flags    = RegFile->GetX<int>( RevReg::a1 );
+  auto mode     = RegFile->GetX<int>( RevReg::a2 );  // ignore unless O_CREAT is set
+
+  /* Read the filename from memory one character at a time until we find '\0' */
+  auto action   = [&] {
+    flags                       = hostOFlags( flags );
```
rkabrick, line 639: Is this formatting from the clang-tidy script or something else? If it's clang-tidy leave it, if not it looks a bit strange

src/RevSysCalls.cc
```diff
@@ -686,13 +737,29 @@ EcallStatus RevCore::ECALL_getdents64() {
...
 }
 
+#if 1
+// https://man7.org/linux/man-pages/man2/lseek.2.html
+// 62, off_t rev_lseek( int fd, off_t offset, int whence)
+EcallStatus RevCore::ECALL_lseek() {
+  output->verbose(
+    CALL_INFO, 2, 0, "ECALL: lseek called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
+  );
+  auto fd     = RegFile->GetX<int>( RevReg::a0 );
```
rkabrick, line 746: Might be stupid but just want to check if @leekillough has any thoughts on the auto here? When you're explicitly getting an `<int>` does it make sense to do auto?

test/syscalls/file_open/file_open.c
```diff
@@ -0,0 +1,42 @@
...
+  // Null-terminate the buffer so we can use it as a string
+  buffer[bytesRead]  = '\0';
+
+  // Write to STDOUT
+  rev_write( STDOUT_FILENO, buffer, bytesRead );
+  // rev_write(STDOUT_FILENO, "\n", 1);
+  // rev_write(STDOUT_FILENO, buffer, bytesRead);
+
+  // Close the file and directory
+  // rev_write(STDOUT_FILENO, CloseMsg, sizeof(CloseMsg));
```
rkabrick: Lots of things commented in here

rkabrick (approved):
Good to go from a logic perspective. Cool changes! 

As long as Lee is cool with the `auto` usage and the comments are annotated or deleted I'm good to go

common/syscalls/syscalls.h
kpgriesser: removing commented functions and added this reference so we can add them when/if needed
```+// ===================== RV system calls beyond 1024 used in riscv-pk
+// --- ref: https://github.com/westerndigitalcorporation/RISC-V-Linux/blob/master/riscv-pk/pk/syscall.
```

src/RevSysCalls.cc
kpgriesser: I'm following the same pattern as other `ecall` routines. Easy to change if anyone has heartburn but then it follows all of them would change :)

test/syscalls/file_open/file_open.c
kpgriesser: Actually, this test is a carryover from #369 and is redundant now that we have the 'open' test. Removing this test.

src/RevSysCalls.cc
kpgriesser: probably a formatting thing but I removed the whitespace and will recommit

src/RevSysCalls.cc
leekillough: I am fine with or without `auto`. We are using `<int>` so we're explicit about the RHS.

`int fd` is just as fine as `auto fd` 

When I prefer `auto` is when the type on the RHS is very long to type out (which `int` isn't), or when the RHS return type is declared somewhere else (like in a function being called), we don't control it here, and we don't care about the type for our uses here, like if we call a function which returns an integer type we don't care about and we only care if the return value is `0`:

```c++
auto val = some_func();
if( val != 0) {
}
```

leekillough (approved):
As long as all tests pass for both release and debug builds

PR #374, opened 2025-06-14
detecting unsupported ECALLs and test checking improvements

kpgriesser:
The majority of ecall routines perform no operation but return 'SUCCESS'. This can lead to a false sense of security when writing RV code for REV that that does system calls that are essentially NOPs.  This PR provides a compile time overridable macro to print an error string for these ECALLs.  Originally I was going to make this a fatal error but it would have required adding a compile pragma to suppress the 'maybe no return' compiler warning. Also, I thought other projects might be running code with a rich mix of 'NOP' system calls that would not appreciate these kinds of failures.  In the worst case, we can revert to NOP functionality using  CMAKE_CXX_FLAGS = -D _REV_ECALL_OVERRIDE.

For regression testing, the test/CMakeList.txt was updated to provide a fail regular expression that detects these kinds of error strings. I ran the regressions on Ubuntu and MacOs as well as for both Debug and Release compilations and all current tests continue to pass.

I should also mention that the error string check is added to allow self checking tests to print error strings to the log file (standing for console output for now) and generate a regression test failure that is independent from the simulation return status code

### DIFF
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -7,6 +7,12 @@
 #include <sys/fcntl.h>
 #include <sys/xattr.h>
 
+#ifndef _REV_ECALL_OVERRIDE
+#define UNIMPLEMENTED() output->verbose( CALL_INFO, 0, 0, "Error: unimplemented ecall\n" )
+#else
+#define UNIMPLEMENTED()
+#endif
+
 namespace SST::RevCPU {
 
 /// Parse a string for an ECALL starting at address straddr, updating the state
@@ -62,6 +68,7 @@ EcallStatus RevCore::ECALL_io_setup() {
   output->verbose(
     CALL_INFO, 2, 0, "ECALL: io_setup called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
   );
+  UNIMPLEMENTED();
   return EcallStatus::SUCCESS;
 }
 
@@ -70,6 +77,7 @@ EcallStatus RevCore::ECALL_io_destroy() {
   output->verbose(
     CALL_INFO, 2, 0, "ECALL: io_destroy called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
   );
+  UNIMPLEMENTED();
   return EcallStatus::SUCCESS;
 }
 
@@ -78,6 +86,7 @@ EcallStatus RevCore::ECALL_io_submit() {
   output->verbose(
     CALL_INFO, 2, 0, "ECALL: io_submit called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
   );
+  UNIMPLEMENTED();
   return EcallStatus::SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


test/rev-model-options-config.py
```diff
@@ -29,7 +29,7 @@
 parser.add_argument("--machine", help="Machine type/configuration", default="[CORES:RV64GC]")
 parser.add_argument("--args", help="Command line arguments to pass to the target executable", default="")
 parser.add_argument("--startSymbol", help="ELF Symbol Rev should begin execution at", default="[0:main]")
-parser.add_argument("--trcStartCycle", help="Starting cycle for rev tracer [default: 0 (off)]")
+parser.add_argument("--trcStartCycle", help="Starting cycle for rev tracer [default: 0 (off)]", default=0)
```
kpgriesser, line 32: This is not related to the PR. Not having the default value causes SST  to error out if trcStartCycle is not set because it is set to Python 'None' type. The default=0 disables the tracer

src/RevSysCalls.cc
```diff
@@ -62,6 +68,7 @@ EcallStatus RevCore::ECALL_io_setup() {
   output->verbose(
     CALL_INFO, 2, 0, "ECALL: io_setup called by thread %" PRIu32 " on hart %" PRIu32 "\n", ActiveThreadID, HartToExecID
   );
+  UNIMPLEMENTED_BEHAVIOR;
```
leekillough:
```c++
UNIMPLEMENTED();
```
would seem better.

src/RevSysCalls.cc
```diff
@@ -7,6 +7,12 @@
 #include <sys/fcntl.h>
 #include <sys/xattr.h>
 
+#ifndef _REV_ECALL_OVERRIDE
+#define UNIMPLEMENTED_BEHAVIOR output->verbose( CALL_INFO, 0, 0, "error: unimplemented ecall" );
```
leekillough: I would leave off the semicolon at the end of the macro definition, so that:
```c++
if(test)
  UNIMPLEMENTED_BEHAVIOR;
else
  {
    /* something else */
  }
```
will work.

It might also be better to declare it as a function-like macro, `#define UNIMPLEMENTED_BEHAVIOR()`.

I also think `UNIMPLMENTED()` is a better shorter name than `UNIMPLEMENTED_BEHAVIOR()`.

test/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ else()
 endif()
 
 set (passRegex "Simulation is complete")
-
+set (failRegex "[^_a-z][Ee][Rr][Rr][Oo][Rr][: ]")
```
leekillough: Might want to write it as:
```
set(failRegex "\b[Ee][Rr][Rr][Oo][Rr]\b")

**Review by leekillough — CHANGES_REQUESTED** [review]

**Comment by leekillough** [comment]
Also note that I am getting a strange error in Release builds but not in Debug builds:
```
FATAL:  SST Core: can't find requested element library 'Statistic' with element type 'sst.AccumulatorStatistic'

```
It happens with `RevMem` tests (`--enableMemH=0`), which is strange, because `sst.AccumulatorStatistic` is not used unless `MemH` is enabled.

kpgriesser:
@leekillough  I've merged your requested changes and retested on Mac and ubuntu for Release and Debug builds. Everything appears to be passing. Are you still seeing failures for --enabledMemH=0?

leekillough (approved):
It passes with both Debug and Release builds with my latest tests

PR #376, opened 2025-06-18
Provide support and testing for printf and sprintf

kpgriesser:
After noticing several versions of 'printf.h' floating around I thought it best to provides a centralized and maintained copy along with more complete testing and complete the sprintf support. This PR provides additional verified functionality as well as testing improvements to allow developing tests outside of the rev repository. 

- Replaced 'printf.h' with 'rev-printf.h' and moved into test/include directory. 
- Introduced the REVHOME environment variable (currently used by other projects)
- CMakeLists.txt changed to set REVHOME to the project root directory for `ctest`.
- External test development can override REVHOME to point to independent builds.
- Continued to standardize the tests in the syscalls area to run common source code on REV or the host using the local makefiles.  For example, `test/syscalls/printf` can be run in that local test directory   as:
```make clean run TARG=host```
or 
``` make clean run```

The former  will compile and run the test on the host system while the latter will compile and run it on REV.

- Added additional testing for printf and added support for sprintf. 
- Completed functionality so the correct byte count is returned from printf and sprintf. 
- Employed rev_fast_printf ecall in the debug of the printf and sprintf code.

### CODE REVIEW

File: src/RevSysCalls.cc
Comment (kpgriesser) [inline src/RevSysCalls.cc:1]: This file was missing the copyright boilerplate for some reason

File: test/include/rev-printf.h
Comment (kpgriesser) [inline test/include/rev-printf.h:1]: The original printf.h file was developed by Dave. Most of my changes were related to sprintf and having the routines return a byte count.

File: test/syscalls/mkinc/rev.mk [hunk 1]
```diff
@@ -8,34 +8,42 @@
 # See LICENSE in the top level directory for licensing details
 #
 
-#CC=riscv64-unknown-elf-gcc
-CC="${RVCC}"
-#CCOPTS += -march=rv64g
-CCOPTS += -march=rv64imafdc
+ifndef REVHOME
```
kpgriesser, line 11: ctest based testing is only using the makefile for compiling. However, using the local makefile allows portability outside of the rev repo. Also, we can run and debug tests in the local directory using make.

test/syscalls/printf/Makefile
```diff
@@ -1,27 +1,29 @@
 #
 # Makefile
 #
-# makefile: ex2
+# makefile: printf
```
kpgriesser, line 4: the idea of the makefile in the test directory to include the recipes for the various targets in a common set of makefiles. This allows us to write code that can be run on rev, the host, and eventually, on spike.

test/syscalls/printf/printf.c
```diff
@@ -1,33 +1,90 @@
...
+    exit( 3 );
+
+  const char msg4[] = "Greetings once again - this is a much much longer message and some even nicer yet more ambitious text, in "
+                      "fact, it is bigger than 128 bytes\n";
+  printf( "[printf]%s", msg4 );
+
+  const char shortstring[] = "string";
+  printf( "[printf]Multiple strings and data: %s[%d] %s[%d] %s[%d]\n", shortstring, 0, shortstring, 1, shortstring, 2 );
+
+  printf( "[printf]hex(0x%x) float(%4.2f) size_t(%zu)\n", 0xace, 3.1415, sizeof( shortstring ) );
```
kpgriesser, line 55: This `printf( "[printf]hex(0x%x) float(%4.2f)",...)` is not quite working right but I wanted to check the code in to get general approval on the major changes.

src/RevSysCalls.cc
```diff
@@ -1,3 +1,14 @@
+//
+// _RevSysCalls_cc_
```
kpgriesser, line 2: for some reason the legal boilerplate was missing

test/include/rev-printf.h
```diff
@@ -1,17 +1,42 @@
-// See LICENSE for license details.
-
+//
+// _rev_printf_h_
```
kpgriesser, line 2: this code was originally developed by Dave. My changes are mostly related to bug fixes, adding sprintf, and having the correct number of bytes returned from the routines.

test/include/rev-printf.h
```diff
@@ -1,17 +1,42 @@
...
+// - If the version of printf in stdio is called then rev
+//   should print an error message for an unimplemented ecall.
+
+#ifndef __REV_PRINTF_H__
+
+// #define REV_DEBUG
+#ifdef REV_DEBUG
+#define dprintf rev_fast_printf
+#else
+#define dprintf
```
leekillough: A different name like `debug_printf()` should be used rather than `dprintf()`, since `dprintf()` is a POSIX function like `fprintf()` but to file descriptors like `STDOUT_FILENO=1` and `STDERR_FILENO=2` (see `man dprintf`).

test/include/rev-printf.h
```diff
@@ -210,153 +266,41 @@ static void vprintfmt( void ( *putch )( int, void** ), void** putdat, const char
...
+      bytes++;
       fmt = last_fmt;
       break;
     }
   }
+  return -1;
 }
 
-int printf( const char* fmt, ... ) {
+int rev_printf( const char* fmt, ... ) {
```
leekillough, line 304: You might want to add `__attribute((format(printf, 1, 2)))` or the like, so that warnings are produced when arguments do not match formats. I seem to recall adding `-Wsuggest-attribute=format` at one point.

test/include/rev-printf.h
```diff
@@ -210,153 +266,41 @@ static void vprintfmt( void ( *putch )( int, void** ), void** putdat, const char
...
-  vprintfmt( (void*) putchar, 0, fmt, ap );
-
+  int bytes = rev_vprintfmt( (void*) rev_putchar, 0, fmt, ap );
   va_end( ap );
-  return 0;  // incorrect return value, but who cares, anyway?
+  return bytes;
 }
 
-int sprintf( char* str, const char* fmt, ... ) {
+int rev_sprintf( char* str, const char* fmt, ... ) {
```
leekillough, line 312: See `__attribute__((format(...)))` above.

test/include/rev-printf.h
kpgriesser: done

test/include/rev-printf.h
kpgriesser: done

test/include/rev-printf.h
kpgriesser: done

test/syscalls/printf/printf.c
kpgriesser: Updated the test as this is expected behavior for as yet unsupported format strings

test/include/rev-printf.h
```diff
@@ -89,20 +137,30 @@ static long long getint( va_list* ap, int lflag ) {
...
+static int rev_vprintfmt( void ( *putch )( int, void** ), void** putdat, const char* fmt, va_list ap ) {
   register const char* p;
   const char*          last_fmt;
   register int         ch, err;
   unsigned long long   num;
   int                  base, lflag, width, precision, altflag;
   char                 padc;
 
+  int bytes = 0;
+  debug_printf( "Entered rev_vprintfmt. putdat is %x\n", putdat );
```
donofrio: Is this a spurious debug print statement?

test/include/rev-printf.h
```diff
@@ -89,20 +137,30 @@ static long long getint( va_list* ap, int lflag ) {
...
+  bytes = 0;
   while( 1 ) {
     while( ( ch = *(unsigned char*) fmt ) != '%' ) {
-      if( ch == '\0' )
-        return;
+      if( ch == '\0' ) {
+        debug_printf( "End of string. bytes=%d\n", bytes );
```
donofrio: another one (?)

test/include/rev-printf.h
kpgriesser: I  commented out the debug_printf statements

donofrio (approved):
Looks good! Always useful to have printf!

PR #377, opened 2025-06-18
Disable instruction prefetch when ecalls are executing

kpgriesser:
This PR is motivate by a bug where an illegal instruction is positioned after an 'exit' system call which is prefetched, decoded and fails the simulation. My claim is that ECALLS themselves should not be pipelined at all. Since, in reality, these calls are themselves comprised of code which may need to execute at a different protection level and could affect system behavior. 

The general fix implemented here is to simply not allow instruction prefetching when an ecall is executing.

One caveat noted in the comments is that I would like to change the way 'exit' is handled in REV in general. I will open a separate PR to define this and flush out the details.

### DIFF
--- a/include/RevRegFile.h
+++ b/include/RevRegFile.h
@@ -1,7 +1,7 @@
 //
 // _RevRegFile_h_
 //
-// Copyright (C) 2017-2024 Tactical Computing Laboratories, LLC
+// Copyright (C) 2017-2025 Tactical Computing Laboratories, LLC
 // All Rights Reserved
 // contact@tactcomplabs.com
 //
@@ -268,6 +268,9 @@ public:
 
   /// Get the exception cause
   RevExceptionCause GetSCAUSE() const { return SCAUSE; }
+
+  /// Check if exception is active. Currently only ECALL_USER_MODE supported
+  bool isExceptionActive() const { return SCAUSE == RevExceptionCause::ECALL_USER_MODE; }
 
   /// GetX: Get the specifed X register cast to a specific integral type
   template<typename T, typename U>
--- a/src/RevCore.cc
+++ b/src/RevCore.cc
@@ -1656,8 +1656,8 @@ bool RevCore::ClockTick( SST::Cycle_t currentCycle ) {
 
     feature->SetHartToExecID( HartToDecodeID );
 
-    // fetch the next instruction
-    if( !PrefetchInst() ) {
+    // fetch the next instruction unless exception is active
+    if( !PrefetchInst() || RegFile->isExceptionActive() ) {
       Stalled = true;
       Stats.cyclesStalled++;
     } else {
@@ -1950,7 +1950,7 @@ void RevCore::CreateThread( uint32_t NewTID, uint64_t firstPC, void* arg ) {
 //
 // Returns true if an ECALL is in progress
 bool RevCore::ExecEcall() {
-  if( RegFile->GetSCAUSE() != RevExceptionCause::ECALL_USER_MODE )
+  if( !RegFile->isExceptionActive() )
     return false;
 
   // ECALL in progress
--- a/src/RevSysCalls.cc
+++ b/src/RevSysCalls.cc
@@ -1202,12 +1202,19 @@ EcallStatus RevCore::ECALL_exit() {
     CALL_INFO,
     0,
     0,
-    "thread %" PRIu32 " on hart %" PRIu32 "exiting with"
+    "thread %" PRIu32 " on hart %" PRIu32 " exiting with"
     " status %" PRIu64 "\n",
     ActiveThreadID,
     HartToExecID,
     status
   );
+  // TODO exit shuts down the sst process which circumvents the
+  //      component lifecycle and overrides sst status code. We
+  //      should instead indicate the simulated component is finished
+  //      and let sst shutdown cleanly. The status code from REV is
+  //      printed to the log file for post-processing. We can support
+  //      testing by providing on option to 'fatal' out when status
+  //      is not 0 (or a user desired value if non-zero is expected)
   exit( int( status ) );
   // return EcallStatus::SUCCESS;
 }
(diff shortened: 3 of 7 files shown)


src/RevCore.cc
```diff
@@ -476,8 +476,7 @@ RevInst RevCore::DecodeCIInst( uint32_t Inst, uint32_t Entry ) const {
   //if c.addi, expands to addi %rd, %rd, $imm so set rs1 to rd -or-
   // c.slli, expands to slli %rd %rd $imm -or -
   // c.addiw. expands to addiw %rd %rd $imm
-  if( ( 0b01 == CompInst.opcode && 0b000 == CompInst.funct3 ) ||
```
kpgriesser: clang-format only

src/RevCore.cc
```diff
@@ -476,8 +476,7 @@ RevInst RevCore::DecodeCIInst( uint32_t Inst, uint32_t Entry ) const {
...
-      ( 0b10 == CompInst.opcode && 0b000 == CompInst.funct3 ) ||
```
kpgriesser: clang-format

src/RevCore.cc
```diff
@@ -1536,7 +1535,8 @@ bool RevCore::DependencyCheck( uint32_t HartID, const RevInst* I ) const {
   // For ECALL, check for any outstanding dependencies on a0-a7
   if( I->opcode == 0b1110011 && I->imm == 0 && I->funct3 == 0 && I->rd == 0 && I->rs1 == 0 ) {
     for( RevReg reg : { RevReg::a7, RevReg::a0, RevReg::a1, RevReg::a2, RevReg::a3, RevReg::a4, RevReg::a5, RevReg::a6 } ) {
-      if( LSQCheck( HartToDecodeID, RegFile, safe_static_cast<uint16_t>( reg ), RevRegClass::RegGPR ) || ScoreboardCheck( RegFile, uint16_t( reg ), RevRegClass::RegGPR ) ) {
```
kpgriesser: clang-format

src/RevCore.cc
```diff
@@ -1808,7 +1808,9 @@ bool RevCore::ClockTick( SST::Cycle_t currentCycle ) {
       RegFile->IncrementInstRet();
 
       // Only clear the dependency if there is no outstanding load
-      if( ( RegFile->GetLSQueue()->count( LSQHash( Pipeline.front().second.rd, InstTable[Pipeline.front().second.entry].rdClass, HartID ) ) ) == 0 ) {
```
kpgriesser: clang-format

src/RevCore.cc
```diff
@@ -1656,8 +1656,8 @@ bool RevCore::ClockTick( SST::Cycle_t currentCycle ) {
 
     feature->SetHartToExecID( HartToDecodeID );
 
-    // fetch the next instruction
-    if( !PrefetchInst() ) {
+    // fetch the next instruction unless in ECALL
+    if( !PrefetchInst() || ExecEcall() ) {
```
kpgriesser: This is the only functional change to fix the instruction after the ecall from entering the pipeline.

src/RevCore.cc
kpgriesser: clang-format

src/RevSysCalls.cc
```diff
@@ -1191,12 +1191,19 @@ EcallStatus RevCore::ECALL_exit() {
...
     0,
     0,
-    "thread %" PRIu32 " on hart %" PRIu32 "exiting with"
+    "thread %" PRIu32 " on hart %" PRIu32 " exiting with"
     " status %" PRIu64 "\n",
     ActiveThreadID,
     HartToExecID,
     status
   );
+  // TODO exit shuts down the sst process which circumvents the
```
kpgriesser, line 1211: topic of another PR

src/RevCore.cc
leekillough: `ExecEcall()` executes one clock of the `ECALL` if an `ECALL` is in effect. It does not simply return whether an `ECALL` is in effect, but also executes `ECALL` if one is in effect (hence the name _Exec_ Ecall).

With this change, you are moving the `ExecEcall()` call up next to `PefetchInst()`. I would make sure that it is safe to move this `ExecEcall()` call up:
```c++
    // fetch the next instruction
    if( !PrefetchInst() || ExecEcall() ) {
                           ^^^^^^^^^^^
      Stalled = true;
      Stats.cyclesStalled++;
    } else {
      Stalled = false;
    }

    if( !Stalled && !CoProcStallReq[HartToDecodeID] ) {
      Inst       = FetchAndDecodeInst();
      Inst.entry = RegFile->GetEntry();
    }

    // Now that we have decoded the instruction, check for pipeline hazards
    if( ExecEcall() || Stalled || DependencyCheck( HartToDecodeID, &Inst ) || CoProcStallReq[HartToDecodeID] ) {
        ^^^^^^^^^^^
```

You are also duplicating the call to `ExecEcall()`, which has the effect of executing two clocks of the `ECALL` but without a complete clock cycle having been fully simulated between them.

I would suggest moving the `ExecEcall()` up and removing the second `ExecEcall()` if it is safe to do so. If not, and you only need to test whether an `ECALL` is active, you can use the test at the beginning of `ExecEcall()`, like:

```c++
if( !PrefetchInst() || RegFile->GetSCAUSE() == RevExceptionCause::ECALL_USER_MODE ) {
    Stalled = true;
    Stats.cyclesStalled++;
} else {
    Stalled = false;
}
```
However, I don't know whether setting `Stalled` to true has the right side effects, or whether it might have other side effects we don't want.

If this `RegFile->GetSCAUSE() == RevExceptionCause::ECALL_USER_MODE` test is repeated, both in the prefetch test and in the `ExecEcall()` function, it might be good to write a separate function for it, for brevity, such as in `class RevRegFile`. I imagine that if protected mode or more exception states are handled, there might need to be a function which tests not just `SCAUSE` but possibly other "causes" of exceptions. A function like `bool isExceptionActive() const` might be good for testing whether such a state exists.

kpgriesser:
Added `bool isExceptionActive()` per Lee's suggestion. To look for unintended side effects I compared the traces for ecalls with and without these changes but saw no differences.
Also squashed the commits and reformatted on Ubuntu instead of Mac which eliminated the formatting-only differences.
Ready for review and approval.

donofrio (approved):
Looks good - this is a good step towards pipelining

include/RevRegFile.h
```diff
@@ -1,7 +1,7 @@
 //
 // _RevRegFile_h_
 //
-// Copyright (C) 2017-2024 Tactical Computing Laboratories, LLC
+// Copyright (C) 2017-2025 Tactical Computing Laboratories, LLC
```
rkabrick, line 4: https://tenor.com/view/spongebob-patrick-24-25-gif-13595902

rkabrick (approved):
this looks great!

PR #378, opened 2025-07-09
Provide build and run flow for spike, rev, and host

kpgriesser:
This is a follow-on PR to #376 and completes the support for running tests on spike ( in addition to rev and host). This is useful in the development of emulated system calls by providing the ability to compare results on different targets. Only two tests are currently using this in the rev repository but the flow is also being used in developing external test suites through the use of the REVHOME environment variable.

Examples:
```
cd test/syscalls/printf
make clean run                          # build and run for  rev
make clean run TARG=host    # build and run for host
make clean run TARG=spike   # build and run for spike
```
Tests may use compile-time flags to differentiate the different targets. 

Specifically:
 REV_TARGET
 HOST_TARGET
 SPIKE_TARGET

The make file is only used by `ctest` for compiling the test for REV so the tests are compatible with the current testing methodology.

### DIFF
--- a/test/syscalls/.gitignore
+++ b/test/syscalls/.gitignore
@@ -1,2 +1,4 @@
 *.tmp
 *.dSYM
+*.spike
+*.d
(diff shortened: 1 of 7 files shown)


leekillough (approved):
LGTM but you might consider supporting QEMU too.

PR #379, opened 2025-08-29
Clean up Rev

leekillough:
This cleans up a few Rev issues discovered recently:
* Re-enable forced Git hooks for code quality assurance which were removed in https://github.com/tactcomplabs/rev/commit/18aa6457f25e04d1542c59047a5d3b292336ca6a
  - The Git hooks check script excludes very specific users, so `devuser` was added for CI testing
  - Other exceptions should be added `scripts/test_git_hooks.sh` rather than turning it off completely in `CMakeLists.txt`
* Increase the timeout of the recently-added `printf` tests from 30 to 60 seconds -- I was getting timeout failures
* A `printf.d` file is created in the test source tree when running the `printf` tests -- I added it to `.gitignore`
* Fix the RISC-V FP->Integer conversion instructions, and their automatic test generator, so that Inexact exceptions are not generated nor expected when Invalid exceptions are generated. IEEE 754 does not require, but allows, Inexact exceptions to be raised when Invalid exceptions are raised. RISC-V does not raise Inexact exceptions on FP->Integer conversions when Invalid exceptions occur, while x86 does.
* Some minor whitespace changes caused by `clang-format` version differences


kpgriesser:
No issues from my side. I would let John approve this since he initiated https://github.com/tactcomplabs/rev/commit/18aa6457f25e04d1542c59047a5d3b292336ca6a

jleidel:
don't re-enable the git hook

kpgriesser:
FYI - I incorporated your change to increase the timeout for the printf test in #378

leekillough:
> don't re-enable the git hook

Why?

jleidel:
> > don't re-enable the git hook
> 
> Why?

the dev environment should not depend upon a git hook.

PR #380, opened 2025-09-10
Change DumpMem() to work on translated rather than physical addresses 

donofrio:
Updated the `DumpMem()` debug function to output based on the translated addresses - the same way `ReadMem()` works in RevMem only.  Previously `DumpMem()` operated on physical addresses. `mem_dump` Tests updated to dump valid address ranges. Output format remains the same:

Data segment dump:
```0x000000003ffffaba: 41 01 82 80 72 73 75 2f 6e 69 62 2f 79 65 78 2f  A...rsu/nib/yex/
0x000000003ffffaca: 00 00 73 65 be fa ff 3f 00 00 00 00 00 00 00 00  ..se���?........```

Test segment dump:
```0x000000000001020a: 33 0f 00 00 33 0f 00 00 01 00 e6 60 46 64 25 61  3...3.....�`Fd%a
0x000000000001021a: 82 80 01 45 82 80 49 66 c5 65 49 65 13 06 06 02  ...E..If�eIe....```

### DIFF
--- a/src/RevMem.cc
+++ b/src/RevMem.cc
@@ -782,6 +782,8 @@ void RevMem::DumpMem( const uint64_t startAddr, const uint64_t numBytes, const u
     }
     outputStream << std::endl;
   }
+  delete[] printMem;
+  printMem = NULL;
 }
 
 void RevMem::DumpMemSeg( const std::shared_ptr<MemSegment>& MemSeg, const uint64_t bytesPerRow, std::ostream& outputStream ) {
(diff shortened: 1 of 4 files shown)


kpgriesser (approved):
No specific comments on the code but this looks like it has only been supported when MemH is disabled. Is that correct? If so, I think it would be nice to have that mentioned in the help text. Even better, have the dump print a message as well so the user doesn't misinterpret the data. 
If it is supported for MemH then I apologize for my misunderstanding.

donofrio:
Added a check to print a warning message if you try to use this with Mem Hierarchy

